PR #164, opened 2014-07-02
Remove phstore flag from style and replace ist by a generic hstore flag

giggls:
Remove phstore flag from style and replace ist by a generic hstore flag
The generic flag is also usable for points and lines rather than only polygons.

This is useful in conjunction with the --hstore-match-only option


giggls:
Would you please tell me what I _exactly_ need to do to get this merged? I'm really fed up with this issue now because this is an almost trivial patch and it will not work, that you are trying to make me fix other stuff all along the way to get this merged.

Actually setting up the test you requested has been more work than doing the patch itself! This is already pending for half a year now. I already suggested back in summer to name the flag something like nocolumn. I don't care how it is called.

Just to point out this again: I have been the one who once introduced the PHSTORE flag into the code and it has just been a mistake to enable it for polygon geometries only. All I want to do now is to correct this mistake I once made.

pnorman:
EWG meeting log: http://osmfoundation.org/wiki/Working_Group_Minutes/EWG_2014-12-22

It needs to be clear what this new flag does - from a combination of the name of the flag and the docs. Because it has not been, there has been virtually no interest in this PR. Normally there would be others commenting on a PR, but the confusion seems to of kept them from doing that.

Needless to say, it also needs to pass existing tests and have tests added for new functionality - i.e. tests for both the existing phstore the new flag.

default.style
```diff
@@ -24,7 +24,9 @@
 #
 # linear - Create a column for this tag
 #
-# phstore - Don't create a column for this tag, but objects with the tag are areas
+# hstore - add tag to hstore only instead of its own column
```
lonvia: This is confusing, the flag rather does something along the lines of "don't create a separate column for this tag, but always include objects with this tag".

default.style
```diff
@@ -148,12 +150,12 @@ way        way_area     real                # This is calculated during import
 # Area tags
 # We don't make columns for these tags, but objects with them are areas.
 # Mainly for use with hstore
```
lonvia: Comments need to be adapted.

lonvia:
I'm with @giggls here. The pull request doesn't break anything and makes the meaning of the flag much more logical. I only find the documentation still a bit misleading, see inline comments.

pnorman:
If I get time next month, I'll rewrite the docs, rename hstore to nocolumn, and merge. Don't let this stop anyone else from doing it.

PR #173, opened 2014-07-25
Added the capability to specify a target schema for the tables.

shaycraft:
This adds the capability to use tables in a different schema than public. Pass -X or --schema to specify which schema to use.  Uses SET search_path TO schema now, as opposed to explicitly defining it per each table.


pnorman:
I've been rewriting this against C++ and I realized there's a non-obvious bug in the logic which needs fixing.

The default schema for a user can be `public`, but is not always. If it is not public, then the logic of `if schema != "public" then set search_path` does not allow you to set the schema to public.

boost::optional might be the way to do this.

pnorman:
Thanks for this PR - I've rewritten against the current code and expanded tests/documentation.

Issue #247, opened 2015-01-01
middle-pgsql fails to escape special characters


lonvia:
The function that escapes tag values, [escape_tag() in pgsql_middle](https://github.com/openstreetmap/osm2pgsql/blob/master/middle-pgsql.cpp#L144), escapes '\n', '\r' and '\t' to 'n', 'r' and 't'. That doesn't look right. It isn't clear from the code if the intention was to write the special character into the database or a slash and a letter.

To reproduce load http://www.openstreetmap.org/api/0.6/node/30029830 into the database. I'll provide a test shortly.

pnorman:
When I was looking at this part of the code I as wondering if we should really be writing our own escaping code or if we could use something from libpq or prepared statements

lonvia:
That escape code certainly isn't pretty, in particular, as it is repeated three times throughout the code. Prepared statements are not really an alternative as much of it is about escaping strings for copy mode. The libpq escaping seems to want to do its own memory allocation. That sounds expensive. Also, we might want to switch to C++ bindings first.

pnorman:
> Prepared statements are not really an alternative as much of it is about escaping strings for copy mode

Oh ya - that excludes prepared, and [`COPY` is far faster](http://www.postgresql.org/docs/9.3/static/populate.html#POPULATE-COPY-FROM).

> switch to C++ bindings first

Do you mean something like like libpqxx?

lonvia:
Yes, I had libpqxx in mind. Documentation states that it just wraps libpq, I wonder if that is true for the escape functions as well.

pnorman:
I used libpqxx with cgimap, I'm not sure it's a great option. Several PostgreSQL people recommended just using libpq. @zerebubuth, any thoughts?

zerebubuth:
I think libpqxx is a really good library. There are some issues around the escaping there as well (IIRC it doesn't natively support hstore, we'd have to add that), and I don't remember whether it allocates while escaping as well...

As for `escape_tag()` - I can't immediately see the problem with it. It seems to add one (or two, if `escape` is set - but I have no idea when that would be set) backslash in front of ", \, \n, \r, \t - which seems to be what the [documentation](http://www.postgresql.org/docs/9.2/static/sql-copy.html#AEN66683) is asking for...

lonvia:
[...]

Issue #228, opened 2014-12-06
Memory leaks of C++ version (Windows-only?)


alex85k:
Some out-of-memory problems of C++-switched Windows build were exposed https://github.com/openstreetmap/osm2pgsql/issues/17#issuecomment-62354744

After adding opensource Visual Leaks Detector (no Valgrind available) to the project (``#include <vld.h>``),  I was able to catch multiple leaks. Will post the messages from log here and try to provide possible fixes in separate PRs.

pnorman:
We were just down to non-critical leaks, and since then we're using std::string in the middle. Any new leaks can get their own issues.

PR #248, opened 2015-01-04
Replace manually allocated strings in middle-pgsql with std::string

lonvia:
This removes all the remaining char* that were manually allocated through either malloc or strdup and replaces them with std::string. Namely they are

 * sql commands in the table descriptions
 * buffers for building tag and node lists
 
It fixes another couple of smaller memory leaks (see #228). Note that it also contains a test for #247 which is currently disabled because it fails.

@pnorman would you mind running another round of performance tests with the full planet? It would also be good to have some numbers on osc updates, if your setup permits to do that easily.

### DIFF
--- a/middle-pgsql.cpp
+++ b/middle-pgsql.cpp
@@ -256,9 +256,7 @@ const char *pgsql_store_tags(std::string &buffer, const struct keyval *tags, int
     if (!first) buffer += ',';
     buffer += '"';
     escape_tag( &buffer, i->key, escape );
-    buffer += '"';
-    buffer += ',';
-    buffer += '"';
+    buffer += "\",\"";
     escape_tag( &buffer, i->value, escape );
     buffer += '"';
 
@@ -396,7 +394,7 @@ int middle_pgsql_t::local_nodes_set(osmid_t id, double lat, double lon, const st
       pgsql_store_tags(buffer, tags, 1);
       buffer += '\n';
 
-      pgsql_CopyData(__FUNCTION__, node_table->sql_conn, buffer.c_str());
+      pgsql_CopyData(__FUNCTION__, node_table->sql_conn, buffer);
     } else {
         std::string idstr = (single_fmt % id).str();
     #ifdef FIXED_POINT
@@ -581,7 +579,7 @@ int middle_pgsql_t::ways_set(osmid_t way_id, osmid_t *nds, int nd_count, struct 
       pgsql_store_tags(buffer, tags, 1);
       buffer += '\n';
 
-      pgsql_CopyData(__FUNCTION__, way_table->sql_conn, buffer.c_str());
+      pgsql_CopyData(__FUNCTION__, way_table->sql_conn, buffer);
     } else {
       // Three params: id, nodes, tags */
       std::string idstr = (single_fmt % way_id).str();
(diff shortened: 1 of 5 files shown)


pnorman:
> @pnorman would you mind running another round of performance tests with the full planet

pinging myself on this.

> It would also be good to have some numbers on osc updates, if your setup permits to do that easily

Yes  - I can get numbers. They're a bit harder to generate as each set of updates requires a fresh import

pnorman:
note: timing 66f645a24585d548b6ea81aaadc43b5919c77060 and 282c51bd81b8965aee39dd48bd3fa09b83d8c75b

pnorman:
@lonvia Am I looking for performance improvements in processing or pending, and which parts of those tasks?

lonvia:
The main effect should be on writing ways and relations to the middle database, so in the processing stage. I'd expect import to be slightly faster and append to be a bit slower.

pnorman:
On the existing branch I hit this error - makes it hard to compare
```
Using 8 helper-processes
*** glibc detected *** osm2pgsql/osm2pgsql: double free or corruption (out): 0x00007fed580032e0 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x7eb96)[0x7fee7ace2b96]
osm2pgsql/osm2pgsql[0x427cd2]
osm2pgsql/osm2pgsql[0x4272f2]
osm2pgsql/osm2pgsql[0x427414]
osm2pgsql/osm2pgsql[0x427636]
osm2pgsql/osm2pgsql[0x42778a]
osm2pgsql/osm2pgsql[0x421f94]
osm2pgsql/osm2pgsql[0x4200a6]
osm2pgsql/osm2pgsql[0x452f6b]
osm2pgsql/osm2pgsql[0x4534ba]
osm2pgsql/osm2pgsql[0x42f2ae]
/usr/lib/x86_64-linux-gnu/libboost_thread.so.1.55.0(+0xbdaa)[0x7fee7ba5edaa]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x7e9a)[0x7fee7b02be9a]
/lib/x86_64-linux-gnu/libc.so.6(clone+0x6d)[0x7fee7ad583fd]
======= Memory map: ========
```

I'm trying with 1 process for the update, but I wonder if the results will be valid

lonvia:
Fair enough, let's leave that open until after the threading bug has been fixed.

pnorman:
Hardware: i7-4700, 32GB RAM, 2x240GB SSD in RAID0
import command line: `osm2pgsql/osm2pgsql --create --style osm2pgsql/default.style --flat-nodes flat.nodes --cache 20000 --slim --number-processes 8 planet-130904.osm.pbf`
Update command line: ``sudo ./free_mem.sh && time (for n in `seq N1 N2`; do echo $n && osmosis/bin/osmosis --rxc  "diffs/008/$n.osc.gz" --simc --wxc - | osm2pgsql/osm2pgsql -d gis -a -S osm2pgsql/default.style --flat-nodes flat.nodes -C 2000 --slim --number-processes 1 - && sleep 10; done)``

(N1,N2) = (555,575), (576,595), (596,615), (616,635)

Time
```
version      master  std::string
proc n          895          896
proc w         4532         5226
proc r         7611         7566
pend w         3812         3812
index         21753        20213
total import  38753        37870
total update  14575        14252
```

Update time does not include `sleep 10` time and is a total of 81 hours of diffs.

lonvia:
Thanks, @pnorman. That's so not what I expected. I'll see if I can get the performance of processing ways to what it was before. The difference in indexing is curious. Hopefully, it's not due to missing data. I'll check that.

lonvia:
Closing because incompatible with #331.

PR #257, opened 2015-01-19
use one instance of node-persistant-cache per thread

### LINKED ISSUE #229: Crash when updating with more than one thread
Version current as of opening this issue.

Command line (updating a database imported with a very recent version of osm2pgsql)

/usr/local/bin/osm2pgsql -a --slim --number-process 4 -e15:15 --flat-nodes /fast1/osm2pgsql/node.cache --hstore -o /var/lib/mod_tile/dirty_tiles.24387 /var/lib/mod_tile/changes.osc.gz

Fails either with 

     dupplicate!

or with

    dupplicate!
    pure virtual method called
    terminate called without an active exception
    Aborted (core dumped)
 
Reducing --number-process to 1 allows the update to complete without errors.

lonvia:
Somehow the way the node-persistent cache was used, was really confused. New instances have been created but not in the new threads but in the old one and, what is worse, before the cache was actually commited to disk, so any reading would have been rather random. This commit now closes the cache used during the reading phase properly and then creates new instances for each thread. That's how the C code did it before.

There is also some strangeness in the way the cache is opened for reading only (using append mode instead of the cache_already_written variable) but that doesn't look fatal, so it's for another pull request.

Hopefully, fixes #229. @simonpoole would you mind trying it out?

### DIFF
--- a/middle-pgsql.cpp
+++ b/middle-pgsql.cpp
@@ -687,9 +687,6 @@ void middle_pgsql_t::iterate_ways(middle_t::pending_processor& pf)
     // Make sure we're out of copy mode */
     pgsql_endCopy( way_table );
 
-    // at this point we always want to be in append mode, to not delete and recreate the node cache file */
-    if (out_options->flat_node_cache_enabled) persistent_cache.reset(new node_persistent_cache(out_options,1,cache));
-
     // enqueue the jobs
     osmid_t id;
     while(id_tracker::is_valid(id = ways_pending_tracker->pop_mark()))
@@ -881,9 +878,6 @@ void middle_pgsql_t::iterate_relations(pending_processor& pf)
 {
     // Make sure we're out of copy mode */
     pgsql_endCopy( rel_table );
-
-    // at this point we always want to be in append mode, to not delete and recreate the node cache file */
-    if (out_options->flat_node_cache_enabled) persistent_cache.reset(new node_persistent_cache(out_options, 1, cache));
 
     // enqueue the jobs
     osmid_t id;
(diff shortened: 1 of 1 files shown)


pnorman:
I noticed that it moved node cache reporting information to the end of the import

```
Maximum node in persistent node cache: 2445279231
Maximum node in persistent node cache: 2445279231
...repeated...
node cache: stored: 2013054242(100.00%), storage efficiency: 86.67% (dense blocks: 2125712, sparse nodes: 72947236), hit rate: 99.76%
```


Other than that, I'm processing hourly diffs without errors
```
time (for n in `seq 555 654`; do 
echo $n
~/osmosis/bin/osmosis --rxc "$HOME/diffs/008/$n.osc.gz" --simc --wxc - | ~/src/osm2pgsql/osm2pgsql -d gis -a -S ~/src/osm2pgsql/default.style --flat-nodes ~/nodes.bin -C 2000 --slim --number-processes 8 -
sleep 10
done) 2>&1 | tee log.txt
```

lonvia:
The node cache reporting already moved with the switch to C++. It's actually pretty scary because it means that the node cache isn't cleared before indexes are created, so the whole process uses much more memory than before (osm2pgsql:node cache + postgres:maintenance_workmem, instead of max(osm2pgsql:node cache, postgres:maintenance_workmem)). But that's a different issue.

pnorman:
> The node cache reporting already moved with the switch to C++.

Hmm - right you are; hadn't noticed that.

pnorman:
Anyways, this looks to fix the problems in #229, so :+1:

PR #291, opened 2015-02-15
check for empty geometries before using geohash

lonvia:
The geohash function on postgis 1.5. cannot handle empty geometries.

Just fixes the symptoms. Still need to find out why osm2pgsql is creating empty geometries in the first place.

### DIFF
--- a/table.cpp
+++ b/table.cpp
@@ -221,7 +221,9 @@ void table_t::stop()
         pgsql_exec_simple(sql_conn, PGRES_COMMAND_OK, (fmt("ANALYZE %1%") % name).str());
         fprintf(stderr, "Analyzing %s finished\n", name.c_str());
 
-        pgsql_exec_simple(sql_conn, PGRES_COMMAND_OK, (fmt("CREATE TABLE %1%_tmp %2% AS SELECT * FROM %3% ORDER BY ST_GeoHash(ST_Transform(ST_Envelope(way),4326),10)") % name % (table_space ? "TABLESPACE " + table_space.get() : "") % name).str());
+        // Special handling for empty geometries because geohash chokes on
+        // empty geometries on postgis 1.5.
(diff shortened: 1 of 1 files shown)


table.cpp
```diff
@@ -221,7 +221,9 @@ void table_t::stop()
         pgsql_exec_simple(sql_conn, PGRES_COMMAND_OK, (fmt("ANALYZE %1%") % name).str());
         fprintf(stderr, "Analyzing %s finished\n", name.c_str());
 
-        pgsql_exec_simple(sql_conn, PGRES_COMMAND_OK, (fmt("CREATE TABLE %1%_tmp %2% AS SELECT * FROM %3% ORDER BY ST_GeoHash(ST_Transform(ST_Envelope(way),4326),10)") % name % (table_space ? "TABLESPACE " + table_space.get() : "") % name).str());
+        // Special handling for empty geometries because geohash chokes on
+        // empty geometries on postgis 1.5.
+        pgsql_exec_simple(sql_conn, PGRES_COMMAND_OK, (fmt("CREATE TABLE %1%_tmp %2% AS SELECT * FROM %3% ORDER BY CASE WHEN ST_IsEmpty(way) THEN NULL ELSE ST_GeoHash(ST_Transform(ST_Envelope(way),4326),10) END") % name % (table_space ? "TABLESPACE " + table_space.get() : "") % name).str());
```
pnorman: Might st_geohash(st...(nullif(way, 'empty'::geometry)...)) work? I'm on my phone so I can't test right now.

lonvia: `'empty'::geometry` results in a parse error. According to the examples in the postgis docs is has to be `'geometrycollection empty'::geometry` or `'polygon empty'::geometry`, which would be too specific.

pnorman:
:+1:, although I have not yet had a chance to test this, being on my phone.

PR #299, opened 2015-02-21
Cleanup based on the output of LLVM's static analyzer

daniel-j-h:
Mainly memory issues, heap corruptions.
But also non-standard VLA usages, ::qsort vs std::sort and minor issues.

### DIFF
--- a/geometry-builder.cpp
+++ b/geometry-builder.cpp
@@ -21,6 +21,7 @@
 */
 
 #include <iostream>
+#include <algorithm>
 #include <cstring>
 #include <cstdlib>
 #include <cmath>
@@ -81,15 +82,12 @@ struct polygondata
     unsigned        containedbyid;
 };
 
-int polygondata_comparearea(const void* vp1, const void* vp2)
-{
-    const polygondata* p1 = (const polygondata*)vp1;
-    const polygondata* p2 = (const polygondata*)vp2;
+struct polygondata_comparearea {
+  bool operator()(const polygondata& lhs, const polygondata& rhs) {
+    return lhs.area > rhs.area;
+  }
+};
 
-    if (p1->area == p2->area) return 0;
-    if (p1->area > p2->area) return -1;
-    return 1;
-}
 } // anonymous namespace
 
 geometry_builder::maybe_wkt_t geometry_builder::get_wkt_simple(const osmNode *nodes, int count, int polygon) const
@@ -130,12 +128,12 @@ geometry_builder::maybe_wkt_t geometry_builder::get_wkt_simple(const osmNode *no
         wkt->geom = WKTWriter().write(geom.get());
         return wkt;
     }
-    catch (std::bad_alloc&)
+    catch (const std::bad_alloc&)
     {
         std::cerr << std::endl << "Exception caught processing way. You are likelly running out of memory." << std::endl;
         std::cerr << "Try in slim mode, using -s parameter." << std::endl;
     }
-    catch (std::runtime_error& e)
+    catch (const std::runtime_error& e)
     {
         //std::cerr << std::endl << "Exception caught processing way: " << e.what() << std::endl;
     }
@@ -247,12 +245,12 @@ geometry_builder::maybe_wkts_t geometry_builder::get_wkt_split(const osmNode *no
             }
         }
     }
-    catch (std::bad_alloc&)
+    catch (const std::bad_alloc&)
     {
         std::cerr << std::endl << "Exception caught processing way. You are likely running out of memory." << std::endl;
         std::cerr << "Try in slim mode, using -s parameter." << std::endl;
     }
-    catch (std::runtime_error& e)
+    catch (const std::runtime_error& e)
     {
         //std::cerr << std::endl << "Exception caught processing way: " << e.what() << std::endl;
     }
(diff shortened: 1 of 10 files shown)


output-gazetteer.cpp
```diff
@@ -742,7 +740,7 @@ int output_gazetteer_t::process_relation(osmid_t id, struct member *members,
         return 0;
 
     /* get the boundary path (ways) */
-    osmid_t *xid2 = new osmid_t[member_count+1];
+    std::vector<osmid_t> xid2(member_count+1);
```
pnorman: I'm not sure we need the +1 here.

The largest access I see of xid2 is at member_count, when all members are ways.

pnorman: `xid2[count] = members[i].id;` below could become xid2.push_back(members[i].id), and the `count == 0` check could then use xid2.size()

tests/regression-test.py
```diff
@@ -336,7 +336,7 @@ def dbConnect(self):
             self.conn=psycopg2.connect("dbname='osm2pgsql-test'")
             self.conn.autocommit = True
             self.cur = self.conn.cursor()
-        except Exception, e:
+        except Exception as e:
```
pnorman: Fixes #300, but needs checking on Python 2

tests/test-pgsql-escape.cpp
```diff
@@ -1,3 +1,4 @@
+#include <iostream>
```
pnorman: One case of #301

geometry-builder.cpp
```diff
@@ -363,7 +361,7 @@ geometry_builder::maybe_wkts_t geometry_builder::build_polygons(const osmNode *
         WKTWriter writer;
 
         // Procces ways into lines or simple polygon list
-        polygondata* polys = new polygondata[merged->size()];
+        std::vector<polygondata> polys(merged->size());
```
lonvia: If you do that (and I'm all in favour), it would be better to go the whole nine yards and write a constructor for polygondata from gf and merged[i] and then use push_back(). totalpolys can then be replaced by a simple polys.size(). Also means that you can reduce the code duplication with the same stuff below.

middle-pgsql.cpp
```diff
@@ -1263,21 +1263,21 @@ struct pthread_thunk {
...
-};
+}
 } // anonymous namespace
 
 void middle_pgsql_t::stop(void)
 {
     int i;
 #ifdef HAVE_PTHREAD
-    pthread_t threads[num_tables];
+    std::vector<pthread_t> threads(num_tables);
```
lonvia: I don't think there is much to gain here from using a std::vector, as the array is completely static and just local anyways. Doesn't hurt much either, so feel free to ignore me. Dito below.

output-gazetteer.cpp
```diff
@@ -692,22 +692,20 @@ int output_gazetteer_t::process_way(osmid_t id, osmid_t *ndv, int ndc,
 
     /* Are we interested in this item? */
     if (places.has_data()) {
-        struct osmNode *nodev;
         int nodec;
 
+        assert(ndc > 0 && "no nodes to proccess");
```
lonvia: Really? Is this guaranteed anywhere further up in the processing chain? I would not assume that OSM data does not contain ways without nodes.

tests/regression-test.py
lonvia: Works fine in python 2.

middle-pgsql.cpp
daniel-j-h: This is not legal C++. VLA's are an extension and not included in (even the latest) C++ standard. The reason for using std::vector here is not failing to compile using a standard conforming compiler (e.g. with -pedantic or without extensions).

output-gazetteer.cpp
daniel-j-h: I added the assertion, because otherwise your old code would have done a malloc(0), which quoting the docs
> returns either NULL, or a unique pointer value that can later be successfully passed to free()

only passing it to free is guaranteed to work, but not e.g. dereferencing it, even in cases where it is _not_ a null pointer.

It's the same with &nodev[0]; if ndc == 0, &nodev[0] is illegal (this could even be simplified in C++11 with nodev.data() which returns nullptr in this case).

If ndc == 0 can indeed happen, the code has to be rewritten with this in mind.

lonvia: Then fix it correctly by discarding ways with 0 nodes. Before your change the malloc(0)'d pointer was never dereferenced, so things happened to work fine.

pnorman: Could we change the if check above to `if (places.has_data() && ndc > 0)`?

lonvia: The current code actually produces empty geometries for 0-node ways. Simply ignoring them might leave zombies when running in append mode. Better to dump empty ways at the very beginning of the function as follows:
~~~
  if (ndc <= 0) {
    delete_unused_full('W', id);
    return 0;
  }
~~~

pnorman:
I'm thinking I'll pick off up to and including 4a5fc188e70a24c9772765c5acbb324e18b07be0 and PR it after addressing https://github.com/openstreetmap/osm2pgsql/pull/299#discussion_r25123148

pnorman:
Looking at bringing in the first parts prior to tagging 0.87.2

pnorman:
f347bf88967ffc5d420bd284a3ded2466b7950da is the one commit remaining, as the VLAs were addressed elsewhere

pnorman:
What f347bf88967ffc5d420bd284a3ded2466b7950da applied to has been rewritten, and #424 fixes issues `scan-build` found in the current codebase.

PR #302, opened 2015-02-22
Catch exceptions in threads and print them upon join

### LINKED ISSUE #298: Core dump with test files
`./osm2pgsql -d gis -S default.style tests/test_output_pgsql_route_rel.osm` produces the output

```
All indexes on  planet_osm_polygon created  in 0s
All indexes on  planet_osm_point created  in 0s
Completed planet_osm_polygon
Completed planet_osm_point
Abort trap (core dumped)
```

pnorman:
This should stop thrown errors from aborting the thread without any message.

Fixes #298

### DIFF
--- a/output-pgsql.cpp
+++ b/output-pgsql.cpp
@@ -44,6 +44,15 @@
 #include <limits>
 #include <stdexcept>
 
+/* make the diagnostic information work with older versions of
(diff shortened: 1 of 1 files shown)


lonvia:
Looks good although I haven't tested the part where a thread throws an exception.

PR #306, opened 2015-03-02
Don't link with libstdc++

AMDmi3:
First, there's no need, as c++ compiler does this implicitly.
Next, this is erroneous as standard c++ library may be named differently (e.g. libc++ from llvm project).


pnorman:
Builds on FreeBSD with clang, `--without-lockfree`

AMDmi3:
> Builds on FreeBSD with clang, --without-lockfree

Yup, I'm going to update the port right away.

pnorman:
> Yup, I'm going to update the port right away

I should also note that the current version builds on freebsd with clang --without-lockfree, it's just a check that we didn't have a regression

AMDmi3:
Of course, I've added that flag to the port as it doesn't build otherwise.

PR #310, opened 2015-03-05
disable way-node index for gazetteer output

lonvia:
Gazetteer never supported marking of objects with changed nodes.
Now way lookup from nodes is no longer needed by the indexing
process either. Therefore disable the index and also marking of
pending ways and relations.


lonvia:
Merged with additional comment as d2afef62a4a0d72c547103392cffcdda25687c6e

PR #312, opened 2015-03-09
use atoll, not atoi, to parse node IDs in nodecachefilereader

woodpeck:
This fixes the problem where nodecachefilereader suffers an integer wraparound if node IDs larger than 2^31-1 are given on command line.


pnorman:
Can [`strtoosmid`](https://github.com/openstreetmap/osm2pgsql/blob/master/osmtypes.hpp#L13) be used?

woodpeck:
Hm, it probably should - or perhaps get rid of the indirection (strtoosmid, PIRdOSMID etc.) in the whole code base?

pnorman:
> Hm, it probably should - or perhaps get rid of the indirection (strtoosmid, PIRdOSMID etc.) in the whole code base?

I'd rather keep the indirection for now.

Most of the `PIRdOSMID` I see are in code that probably needs refactoring anyways. Is that needed with boost::format?

pnorman:
merged

Issue #196, opened 2014-10-25
Compilation error on FreeBSD and OS X


pnorman:
```
libtool: compile:  c++ -DHAVE_CONFIG_H -I. -D_THREAD_SAFE -pthread -I/usr/local/include -I/usr/local/include/libxml2 -I/usr/include -I/usr/include -I/usr/local/include -I/usr/local/include -DOSM2PGSQL_DATADIR=\"/usr/local/share/osm2pgsql\" -Igeos-fallback -pthread -I/usr/local/include -g -O2 -MT osmdata.lo -MD -MP -MF .deps/osmdata.Tpo -c osmdata.cpp  -fPIC -DPIC -o .libs/osmdata.o
In file included from osmdata.cpp:1:
In file included from ./osmdata.hpp:9:
In file included from ./output.hpp:25:
/usr/local/include/boost/lockfree/queue.hpp:91:5: error: implicit instantiation of undefined template 'boost::STATIC_ASSERTION_FAILURE<false>'
    BOOST_STATIC_ASSERT((boost::has_trivial_assign<T>::value));
    ^
/usr/local/include/boost/static_assert.hpp:169:13: note: expanded from macro 'BOOST_STATIC_ASSERT'
            sizeof(::boost::STATIC_ASSERTION_FAILURE< BOOST_STATIC_ASSERT_BOOL_CAST( __VA_ARGS__ ) >)>\
            ^
osmdata.cpp:353:21: note: in instantiation of template class 'boost::lockfree::queue<std::__1::pair<long, unsigned long>, boost::parameter::void_,
      boost::parameter::void_, boost::parameter::void_>' requested here
    pending_queue_t queue;
                    ^
/usr/local/include/boost/static_assert.hpp:87:26: note: template is declared here
template <bool x> struct STATIC_ASSERTION_FAILURE;
                         ^
osmdata.cpp:390:28: error: non-const lvalue reference to type 'middle_t::pending_processor' cannot bind to a value of unrelated type
      '<anonymous>::pending_threaded_processor'
        mid->iterate_ways( ptp );
                           ^~~
./middle.hpp:57:50: note: passing argument to parameter 'pf' here
    virtual void iterate_ways(pending_processor& pf) = 0;
                                                 ^
osmdata.cpp:395:33: error: non-const lvalue reference to type 'middle_t::pending_processor' cannot bind to a value of unrelated type
      '<anonymous>::pending_threaded_processor'
        mid->iterate_relations( ptp );
[...]

PR #317, opened 2015-03-21
convert pending_job_t to a simple struct

lonvia:
The lockfree queue requires a trivially assignable type which std::pair isn't, see #196.

I've confirmed that it compiles successfully against libc++ but could not actually run the tests because I don't have a clang-compiled boost available. Could somebody on OSX or FreeBSD check that this works?

cc @springmeyer


pnorman:
works. updated the readme with new instructions, but left the flag in since it allows testing of the old boost version codepath on systems with a newer boost.

Issue #336, opened 2015-04-20
outtags redeclearation in output_multi_t::process_relation


pnorman:
From https://github.com/openstreetmap/osm2pgsql/pull/273#issuecomment-93937271
> In output_multi::process_relation() `outtags` is redeclared [here](https://github.com/openstreetmap/osm2pgsql/blob/master/output-multi.cpp#L366), that can't be right. The tags are probably meant to be filtered again.

I can see two meaningful behaviors here

1. Reset outtags (if not overwritten by m_tagtransform->filter_rel_member_tags). 
2. Use the transformed tags from before, and toss the later generated outtags

I haven't written a lua transform that does fancy relations processing, multi-backend or otherwise, so the desired behavior is entirely unclear to me. Additionally, this function has six TODO comments

cc @kevinkreiser, who wrote the code.

Reading @systemed's [docs/lua.md](https://github.com/openstreetmap/osm2pgsql/blob/master/docs/lua.md#how-to) it seems that an object can end up in tables based on *either* `filter_basic_tags_rel` or `filter_tags_relation_member`. This would imply that we need to do something with outtags from *both* `m_tagtransform->filter_rel_tags` and `m_tagtransform->filter_rel_member_tags`.

pnorman:
fixed

Issue #158, opened 2014-06-16
Polygons not properly created when a node is at 0.0,0.0


pnorman:
This XML
```xml
<?xml version='1.0' encoding='UTF-8'?>
<osm version="0.6" generator="hand">
<node id="1" version="1" timestamp="2010-01-01T00:00:00Z" uid="1" user="user1" changeset="1" lat="0.0" lon="0.0"/>
<node id="2" version="1" timestamp="2010-01-01T00:00:00Z" uid="1" user="user1" changeset="1" lat="0.01" lon="0.0"/>
<node id="3" version="1" timestamp="2010-01-01T00:00:00Z" uid="1" user="user1" changeset="1" lat="0.01" lon="0.01"/>
<way id="1" version="1" timestamp="2010-01-01T00:00:00Z" uid="1" user="user1" changeset="1">
<nd ref="1"/>
<nd ref="2"/>
<nd ref="3"/>
<nd ref="1"/>
<tag k="building" v="yes"/>
</way>
<relation id="1" version="1" timestamp="2010-01-01T00:00:00Z" uid="1" user="user1" changeset="1">
<member type="way" ref="1" role="outer"/>
<tag k="type" v="multipolygon"/>
</relation>
</osm>
```
with the stylesheet ``node,way building text polygon`` does not result in any rows in planet_osm_polygon when imported with ``./osm2pgsql  -S buildings.style test.osm``

If node 1 is changed to ``<node id="1" version="1" timestamp="2010-01-01T00:00:00Z" uid="1" user="user1" changeset="1" lat="0.001" lon="0.001"/>`` (note lat/lon) it then creates a polygon.

The polygon in the former case is ``LINESTRING(0 1113.19,1113.19 1113.19)`` and in the latter case is ``POLYGON((0 1113.19,1113.19 1113.19,111.32 111.32,0 1113.19))``

lonvia:
Fixed in 27a75c7. Ram cache no longer uses 0,0 as invalid marker.

PR #337, opened 2015-04-20
Fix multi tags redeclaration

zerebubuth:
I think this is a fix for #336 - the re-use of `outtags` and passing the original relation tags to the member filter function rather than the previously filtered tags seems to have been causing missing tag values under some circumstances. Since this type of tag mangling is pretty common in multi-polygon processing, it's fairly likely that some missing tags would be fixed by this.

Also added a new test case which (hopefully) encapsulates the expected behaviour.

### DIFF
--- a/output-multi.cpp
+++ b/output-multi.cpp
@@ -331,9 +331,9 @@ int output_multi_t::process_relation(osmid_t id, const memberlist_t &members,
         relation_delete(id);
 
     //does this relation have anything interesting to us
-    taglist_t outtags;
+    taglist_t rel_outtags;
     unsigned filter = m_tagtransform->filter_rel_tags(tags, *m_export_list.get(),
-                                                      outtags, true);
+                                                      rel_outtags, true);
     if (!filter) {
         //TODO: move this into geometry processor, figure a way to come back for tag transform
         //grab ways/nodes of the members in the relation, bail if none were used
@@ -364,7 +364,7 @@ int output_multi_t::process_relation(osmid_t id, const memberlist_t &members,
         //all this trickery
         int make_boundary, make_polygon = 1;
         taglist_t outtags;
-        filter = m_tagtransform->filter_rel_member_tags(tags, filtered, m_relation_helper.roles,
+        filter = m_tagtransform->filter_rel_member_tags(rel_outtags, filtered, m_relation_helper.roles,
                                                         &m_relation_helper.superseeded.front(),
                                                         &make_boundary, &make_polygon, &roads,
                                                         *m_export_list.get(), outtags, true);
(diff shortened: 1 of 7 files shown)


output-multi.cpp
```diff
@@ -364,7 +364,7 @@ int output_multi_t::process_relation(osmid_t id, const memberlist_t &members,
         //all this trickery
         int make_boundary, make_polygon = 1;
         taglist_t outtags;
-        filter = m_tagtransform->filter_rel_member_tags(tags, filtered, m_relation_helper.roles,
```
pnorman: this should probably be called with the tags from the relation, not the transformed tags from the relation

zerebubuth: The way it currently works in `output-pgsql.cpp` is that `pgsql_out_relation` is called with `outtags`, the transformed tags from the relation, so shouldn't it be consistent between the two?

tests/test_output_multi_poly_trivial.osm
```diff
@@ -0,0 +1,39 @@
...
+   5_____6     7_____8
+
+this file defines the above nodes, ways and a relation
+which contains both w1 and w2.
+-->
+  <node id="1" lon="0" lat="1"/>
+  <node id="2" lon="1" lat="1"/>
+  <node id="3" lon="2" lat="1"/>
+  <node id="4" lon="3" lat="1"/>
+  <node id="5" lon="0" lat="0"/>
```
pnorman: 0,0 shouldn't be used in tests since it's a special point to the dense backend (#158)

tests/test-output-multi-poly-trivial.cpp
```diff
@@ -0,0 +1,134 @@
...
+  check_count(test_conn, 1, "select count(*) from pg_catalog.pg_class where relname = 'test_poly'");
+
+  // expect 2 polygons if not in multi(geometry) mode, or 1 if multi(geometry)
+  // mode is enabled.
+  if (enable_multi) {
+    check_count(test_conn, 1, "select count(*) from test_poly");
+    check_count(test_conn, 1, "select count(*) from test_poly where foo='bar'");
+    check_count(test_conn, 1, "select count(*) from test_poly where bar='baz'");
+
+    // this should be a 2-pointed multipolygon
```
pnorman: 2-polygoned multipolygon, not 2-pointed?

output-multi.cpp
pnorman: Ah, I didn't realize it was that way. It's best to be consistent, even though I'm not sure the transformed tags make as much sense

zerebubuth: If it was done that way intentionally, then I assume it's so that you don't have to repeat (potentially complex / time-consuming) tag processing and normalisation that you might have already done in the relation filter function again in the relation members filter function. For example, if you had a complex method for figuring out whether a relation was a particular type of feature based on various tags then set `INT_feature=1`, you could at least re-use that in the members filter function without needing to recompute it.

But it's just as likely to be hysterical raisins :wink:

PR #345, opened 2015-04-28
Add documentation and improve legibility style.lua

math1985:
This adds documentation to style.lua.

It also has some small changes to improve the legibility: it moves the configurable arrays to the top, and it changes some variable names.

### DIFF
--- a/style.lua
+++ b/style.lua
@@ -207,5 +207,5 @@ function filter_tags_relation_member (keyvalues, keyvaluemembers, roles, memberc
    -- Add z_order key/value combination and determine if the object should also be added to planet_osm_roads
    keyvalues, roads = add_z_order(keyvalues)
 
-   return filter, keyvalues, membersuperseded, boundary, polygon, roads
+   return filter, keyvalues, membersuperseded, linestring, polygon, roads
 end
--- a/style.lua
+++ b/style.lua
@@ -1,8 +1,12 @@
+-- For documentation of Lua tag transformations, see docs/lua.md.
+
+-- Objects with any of the following keys will be treated as polygon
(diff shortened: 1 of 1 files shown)


style.lua
```diff
@@ -156,23 +184,28 @@ function filter_tags_relation_member (keyvalues, keyvaluemembers, roles, memberc
...
-         membersuperseeded[i] = superseeded
+         membersuperseded[i] = superseded
       end
    end
 
+   -- Add z_order key/value combination and determine if the object should also be added to planet_osm_roads
    keyvalues, roads = add_z_order(keyvalues)
 
-   return filter, keyvalues, membersuperseeded, boundary, polygon, roads
+   return filter, keyvalues, membersuperseded, boundary, polygon, roads
```
pnorman: boundary needs changing here too

math1985: Thanks, fixed.

style.lua
```diff
@@ -1,8 +1,12 @@
+-- For documentation of Lua tag transformations, see docs/lua.md.
+
+-- Objects with any of the following keys will be treated as polygon
 polygon_keys = { 'building', 'landuse', 'amenity', 'harbour', 'historic', 'leisure', 
       'man_made', 'military', 'natural', 'office', 'place', 'power',
       'public_transport', 'shop', 'sport', 'tourism', 'waterway',
       'wetland', 'water', 'aeroway' }
 
+-- Objects without any of the following keys will be deleted
```
pnorman: with `--hstore-all` is this true?

math1985: Will check.

math1985: Yes, this is true. Objects without any key in this list will be skipped.

Moreover, objects that do have a key in the list, but no column name (for example an object tagged only with capital=yes) will be skipped too.

PR #346, opened 2015-04-28
Set polygon/linestring differently for some key values differently than the key in default style.lua 

math1985:
This allows the decision whether an object is treated as polygon or linestring
to be set in the style.lua file.

In particular, it treats the following values as linestring:
man_made=embankment, man_made=breakwater, man_made=groyne, natural=cliff, natural=tree_row, historic=citywalls,  waterway=derelict_canal, waterway=ditch, waterway=drain, waterway=river, waterway=stream, waterway=wadi, waterway=weir, power=line, power=minor_line. It also treats highway=services as polygon.

Including this in the default style.lua file has the following advantages:
- It serves as an easy-to-extend example of how the polygon/linestring decision
  can be made based on key.
- It provides sane defaults - it can be expected that most users of the style
  want to treat man_made=embankment and natural=cliff as linestring, and
  highway=services as polygon.

This resolves the following downstream bugs in openstreetmap-carto:
- https://github.com/gravitystorm/openstreetmap-carto/issues/892
- https://github.com/gravitystorm/openstreetmap-carto/issues/268
- https://github.com/gravitystorm/openstreetmap-carto/issues/137

This builds on #345, and thus assumes #345 is merged first.

### DIFF
--- a/stamp-h1
+++ b/stamp-h1
@@ -1,0 +1,1 @@
+timestamp for config.h
--- a/stamp-h1
+++ b/stamp-h1
@@ -1,1 +1,0 @@
-timestamp for config.h
--- a/style.lua
+++ b/style.lua
@@ -6,6 +6,12 @@ polygon_keys = { 'building', 'landuse', 'amenity', 'harbour', 'historic', 'leisu
       'public_transport', 'shop', 'sport', 'tourism', 'waterway',
       'wetland', 'water', 'aeroway' }
 
+-- Objects with any of the following key/value combinations will be treated as polygon
+polygon_values = {highway='services'}
+
+-- Objects with any of the following key/value combinations will be treated as linestring
+linestring_values = {man_made='embankment', natural='cliff'}
+
 -- Objects without any of the following keys will be deleted
 generic_keys = {'access','addr:housename','addr:housenumber','addr:interpolation','admin_level','aerialway','aeroway','amenity','area','barrier',
    'bicycle','brand','bridge','boundary','building','capital','construction','covered','culvert','cutting','denomination','disused','ele',
@@ -127,6 +133,22 @@ function filter_tags_way (keyvalues, numberofkeys)
       end
    end
 
+  -- Treat objects with a key/value combination in polygon_values as polygon
+   for k,v in pairs(polygon_values) do
+      if keyvalues[k] == v then
+         polygon=1
+         break
+      end
+   end
+
+  -- Treat objects with a key/value combination in linestring_values not as polygon
+   for k,v in pairs(linestring_values) do
+      if keyvalues[k] == v then
+         polygon=0
+         break
+      end
+   end
+   
    -- Treat objects tagged as area=yes, area=1, or area=true as polygon,
    -- and treat objects tagged as area=no, area=0, or area=false not as polygon
    if ((keyvalues["area"] == "yes") or (keyvalues["area"] == "1") or (keyvalues["area"] == "true")) then
@@ -168,14 +190,29 @@ function filter_tags_relation_member (keyvalues, keyvaluemembers, roles, memberc
       -- Treat as polygon
       polygon = 1
       polytagcount = 0;
-      -- Count the number of polygon tags of the object
+      -- Count the number of polygon tags
+      -- First count keys in polygon_keys
       for i,k in ipairs(polygon_keys) do
          if keyvalues[k] then
             polytagcount = polytagcount + 1
          end
       end
-      -- If there are no polygon tags, add tags from all outer elements to the multipolygon itself
-      if (polytagcount == 0) then
+      -- Then add key/value combinations in polygon_values
+      for k,v in pairs(polygon_values) do
+         if keyvalues[k] == v then
+            polytagcount = polytagcount + 1
+         end
+      end
+      -- Then substract key/value combinations in linestring_values
+      for k,v in pairs(linestring_values) do
+         if keyvalues[k] == v then
+            polytagcount = polytagcount - 1
+         end
+      end
+
+      -- If the multipolygon has no polygon keys or polygon key/value combinations,
+      -- add tags from all outer elements to the multipolygon itself
+      if (polytagcount <= 0) then
          for i = 1,membercount do
             if (roles[i] == "outer") then
                for k,v in pairs(keyvaluemembers[i]) do
--- a/style.lua
+++ b/style.lua
@@ -8,7 +8,8 @@ polygon_keys = { 'building', 'landuse', 'amenity', 'harbour', 'historic', 'leisu
 
 -- Objects with any of the following key/value combinations will be treated as polygon
 polygon_values = {
-      {'highway', 'services'}
+      {'highway', 'services'},
+      {'junction', 'yes'}
    }
 
 -- Objects with any of the following key/value combinations will be treated as linestring
--- a/style.lua
+++ b/style.lua
@@ -14,6 +14,7 @@ polygon_values = {
 
 -- Objects with any of the following key/value combinations will be treated as linestring
 linestring_values = {
+      {'leisure', 'track'},
       {'man_made', 'embankment'},
       {'man_made', 'breakwater'},
       {'man_made', 'groyne'},
--- a/style.lua
+++ b/style.lua
@@ -7,14 +7,28 @@ polygon_keys = { 'building', 'landuse', 'amenity', 'harbour', 'historic', 'leisu
       'wetland', 'water', 'aeroway' }
 
 -- Objects with any of the following key/value combinations will be treated as polygon
-polygon_values = {highway='services'}
+polygon_values = {
+      {'highway', 'services'}
+   }
 
 -- Objects with any of the following key/value combinations will be treated as linestring
-linestring_values = {man_made='embankment', man_made='breakwater', man_made='groyne',
-      natural='cliff', natural='tree_row', historic='citywalls',  waterway='derelict_canal',
-      waterway='ditch', waterway='drain', waterway='river', waterway='stream', waterway='wadi',
-      waterway='weir', power='line', power='minor_line'
-      }
+linestring_values = {
+      {'man_made', 'embankment'},
+      {'man_made', 'breakwater'},
+      {'man_made', 'groyne'},
+      {'natural', 'cliff'},
+      {'natural', 'tree_row'},
+      {'historic', 'citywalls'},
+      {'waterway', 'derelict_canal'},
+      {'waterway', 'ditch'},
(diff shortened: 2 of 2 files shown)


kreed:
`power=line` as linestring would be nice too. Usually power lines aren't closed loop so it isn't an issue but sometimes it comes up with busbars inside substations.

math1985:
Yes, good suggestion.

math1985:
I added ``power=line`` now, as well as some other key/value combinations.

pnorman:
changes to `filter_tags_relation_member` should also be mirrored in multi.lua, as they are basically the same function.

mboeringa:
@math1985 , you do take into account possible secondary **building NOT IN (Null,'no')** tags? Historic citywalls are sometimes mapped as buildings with individual parts as closed loops / ways. So they may have both a **historic = citywalls** and **building = yes** tag.

Anything tagged with a building tag should be excluded from this evaluation I think.

I am also not sure about defaulting closed ways of **man_made = breakwater** and **man_made = groyne** as linestrings. I agree these are "linear" features by nature protruding from the coast or riverbank, ***but if mapped as a closed way***, they almost certainly represent a micro-mapped polygon representation of the feature, so should likely be polygons.

pnorman:
Renamed the issue - the polygon/linestring decision has always been allowed be set by value in Lua.

math1985:
> changes to filter_tags_relation_member should also be mirrored in multi.lua, as they are basically the same function.

I don't really see how that works - there is not even an polygon_keys in multil.lua? What should exactly be changed?

math1985:
Rebased and solved a bug.

AppVeyor seems to fail, but I only touched the Lua files. What could be the cause of that?

> changes to filter_tags_relation_member should also be mirrored in multi.lua, as they are basically the same function.

@pnorman It's not yet totally clear to me how this works. Both style.lua and multi.lua sometimes add tags from the outer polygons to the multipolygons. However, the circumstances under which they do so are different. The file style.lua adds the tags from the outer polygons to the multipolygon if the multipolygon has no polygon tags. However, multi.lua adds the tags from the outer polygons to the multipolygon if the multipolyon is not a building. That's quite a different condition!  The only change I made to filter_tags_relation_member  is the way the number of polygon tags is counted, and as we don't count polygon tags in multi.lua, it does not seem like we need a similar change in multi.lua at the moment.

math1985:
@mboeringa Good point about historic=citywalls.

So basically the problem is that a closed linestring tagged with historic=citywalls can mean two things. Either it is a circular wall, or a wall of which the outline is tagged. Unfortunately, both of them are used commonly in practice. The question is now what we should assume as default.

Unfortunately the presence of a building tag is not sufficient to tell the two apart. For example [Templin](http://www.openstreetmap.org/#map=19/53.12111/13.49609) has a citywall mapped as area, but no building tag. And conversely, I could also imagine that an area tagged with both a building and citywall tag might signify a building with a city wall around it (like a fortress).

To repeat the question: a citywall as an open linestring will in any case be interpreted as line. What should we do with citywalls as closed linestring?
A. Treat as line, unless tagged area=yes.
B. Treat as area, unless tagged area=no.

I think personally I'd prefer option A. It seems counterintuitive to me that closing a linestring would suddenly change the semantics. Also, I don't think we currently have any tags that can refer both to a line or area that do not need an area=yes tag when used as area, except the tags in this PR.

What do others think?

For breakwater and groyne, the situation is similar, except that closed linestrings for these objects in almost all cases do refer to an area. Should we treat these objects differently?

sommerluk:
@math1985 Could you also add junction=yes as polygon? (Downstream gravitystorm/openstreetmap-carto#1253 )

lonvia:
A general remark: You do say that the PR is intended as an example implementation but at the same time I see a tagging discussion unfold here with plenty of references to the carto style. That tells me this might be the wrong repository for that change. Shouldn't you better start maintaining your own lua style file like you already do for the C style file? I'm just afraid that the more fine details of tagging are lost on the osm2pgsql maintainers as evident from the fact that this PR hasn't been merged in almost a year.

math1985:
I have thought a bit more about this, and I came to the following conclusions:
* An object with both a polygon-key and a line-key are currently treated as polygon. Therefore, the logical thing to do is to treat an object with both a polygon-key and a line-tag (e.g. building=yes and historic=citywalls) also as polygon. I have changed the code to handle this correctly. Thanks again @mboeringa.
* Current practice is that objects that can be interpreted as a line, such as highways, are interpreted as lines even when they form a closed linestring. To treat them as an area, the tag area=yes is needed. I'd like to follow that practice, so with this PR, tags like historic=citywalls or waterway=weir will be interpreted as line even for closed linestring. This can be overruled by adding either area=yes or an area tag, such as building=yes.

math1985:
> @math1985 Could you also add junction=yes as polygon? (Downstream gravitystorm/openstreetmap-carto#1253 )

Added.

math1985:
This is now ready to be merged. I can still squash the commits if you like.

> A general remark: You do say that the PR is intended as an example implementation but at the same time I see a tagging discussion unfold here with plenty of references to the carto style. That tells me this might be the wrong repository for that change.

I understand your point, but I'm not sure if I agree. I think it would be good if osm2pgsql provides sane defaults. The decision whether to treat a certain tag, like natural=tree_row, as polygon or line is a data representation decision, not a rendering decision. I would therefore expect this to be solved in osm2pgsql. It doesn't matter whether I want to use PostGIS-database for some kind of rendering or something else, I'd consider the current default, where a closed natural=tree_row is treated as polygon, as an non-sane default.

However, if the osm2pgsql maintainers say this repository is not the right place for this PR, I would understand. Maintaining an openstreetmap-carto specific .lua-file would work for us too.

stamp-h1
```diff
@@ -0,0 +1 @@
+timestamp for config.h
```
lonvia: This file seems accidentally committed, it is an artifact from autoconf times.

style.lua
```diff
@@ -119,14 +144,31 @@ function filter_tags_way (keyvalues, numberofkeys)
...
+         polygontag = 1
+         -- However, if the key/value combination occurs in linestring_values, do not treat the object as polygon
+         for index,tag in pairs(linestring_values) do
+            if k == tag[1] and keyvalues[k] == tag[2] then
+               polygontag = 0
+               break
+            end
+         end
+         if polygontag == 1 then
+            polygon = 1
```
lonvia: I believe you forgot a  break here. Further, I don't quite see what is the use of 'polygontag'. Setting 'polygon' directly should have the same effect.

style.lua
```diff
@@ -119,14 +144,31 @@ function filter_tags_way (keyvalues, numberofkeys)
...
+         end
+      end
+   end
+
+   -- Treat objects with a key/value combination in polygon_values as polygon
+   for index,tag in pairs(polygon_values) do
```
lonvia: Add a 'if polygon == 0' around the for loop.

style.lua
```diff
@@ -168,14 +210,35 @@ function filter_tags_relation_member (keyvalues, keyvaluemembers, roles, memberc
...
+               polytagcount = polytagcount + 1
+            end
+         end
+      end
+
+      -- Treat objects with a key/value combination in polygon_values as polygon
+      for index,tag in pairs(polygon_values) do
+         if keyvalues[tag[1]] == tag[2] then
             polytagcount = polytagcount + 1
+            break
```
lonvia: If you actually mean to count the polygon tags as it was done before, that break must be removed. However, I'd argue that the code was wrong before already. 'polytagcount' is only used to check if there are any polygon tags, so a simple boolean flag will do, just as above in the way function. So instead of removing this break, rather add one in line 226.

lonvia:
Well, I see the argument that style.lua should do something sane. I'm saying that I'm not able to judge, if this PR counts as bug fixing or as a carto-osm-specific change. I've had a look at the code-part now and added a few comments but I would leave the final decision to @pnorman, if this makes sense to have this change in osm2pgsql or not.

From a practical point of view, you'd want your own lua style anyway because otherwise carto devs are forced to compile their own osm2pgsql all the time just to stay in sync with the style file.

style.lua
math1985: @lonvia If I'm not mistaken, setting 'polygon' directly is not quite the same. Consider an object tagged with a polygon key, like building=yes, and a linestring tag, like historic=citywalls. The outer loop will be run at least twice, once for k=building and once for k=historic. The proposed code sets 'polygontag' and 'polygon' to 1 in the loop with k=building. In the loop with k=historic, 'polygontag' will be set to 0 because historic=citywalls is in linestring_values. However, 'polygon' will remain 1. This is not equivalent to setting 'polygon' directly, because in that case, 'polygon' would also be set to 0 in the loop for k=building.

I admit the code is rather hard to read, perhaps there is a cleaner way to accomplish the same.

math1985:
@lonvia Thanks for your feedback, I think I addressed all issues now.

math1985:
@pnorman Would you be able to review this PR? Let me know if you need anything else.

lonvia:
We had a lengthy discussion about the lua style file at the last hackweek end. The general conclusion was that the lua style in its current state is not really ready to be used. The code is unnecessarily complicated and often terribly inefficient. So we are now seriously considering the idea of rewriting it from scratch.  I'm afraid this will leave this PR (and possibly also #532) a bit in the hanging until there is decision on that matter.

I've also noticed that the code is still not quite right. I'll add a comment about that.

lonvia:
Ignore the last part of my comment. The code is probably okay. But that reminds me of the second observation we made at the hackweekend: we need proper tests for the lua code.

math1985:
Thanks for the update. Do you have any idea how long it will take before the new lua file is ready?

Do you think the problems are so serious that you'd recommend openstreetmap-carto (and therefore the openstreetmap.org tile servers) to wait for the new lua file, or do you think we could use the current lua file within openstreetmap-carto?

pnorman:
style.lua is an **example**, not a default. I wouldn't recommend using it as-is for any project.

math1985:
Rebased.

As far as I'm aware, this is still awaiting review from the osm2pgsql team.

pnorman:
Closing as stale. style.lua is an example, and if a style needs a full transform, the osm-carto one is a good starting point.

We should probably rewrite style.lua based on osm-carto's experience, and add tests.

math1985:
> We should probably rewrite style.lua based on osm-carto's experience, and add tests.

Yes, that would be great.

I don't think it make sense that each individual style decides for themselves which osm objects represent areas and which represent linestrings. This is a decision about the semantics of objects in OSM, not a styling decision. Competing styles are great, but competing semantics make no sense. So I think this is really something that should be done in this repository.

Issue #351, opened 2015-05-22
OOM killed after processing phase


pnorman:
On a machine with 32GB RAM, osm2pgsql 0.87.3, command line `~/osm2pgsql/osm2pgsql -d gis --cache 26000 --flat-nodes nodes.bin --slim --drop --
number-processes 8 -S ~/openstreetmap-carto/openstreetmap-carto.style planet-150202.osm.pbf`

```
osm2pgsql SVN version 0.87.3 (64bit id space)

Using built-in tag processing pipeline
Using projection SRS 900913 (Spherical Mercator)
Setting up table: planet_osm_point
Setting up table: planet_osm_line
Setting up table: planet_osm_polygon
Setting up table: planet_osm_roads
Allocating memory for dense node cache
Allocating dense node cache in one big chunk
Allocating memory for sparse node cache
Sharing dense sparse
Node-cache: cache=26000MB, maxblocks=3328000*8192, allocation method=11
Mid: loading persistent node cache from nodes.bin
Allocated space for persistent node cache file
Maximum node in persistent node cache: 0
Mid: pgsql, scale=100 cache=26000
Setting up table: planet_osm_nodes
Setting up table: planet_osm_ways
Setting up table: planet_osm_rels

Reading in file: planet-150202.osm.pbf
Processing: Node(2715748k 2596.3k/s) Way(272021k 51.88k/s) Relation(2630170 269.21/s)
Standard exception processing relation id=3890452: TopologyException: side location conflict at 1038877.48 5779728.6200000001
Processing: Node(2715748k 2596.3k/s) Way(272021k 51.88k/s) Relation(2841950 273.71/s)
Standard exception processing relation id=4136486: TopologyException: side location conflict at 2924456.4100000001 9012036.7799999993
Processing: Node(2715748k 2596.3k/s) Way(272021k 51.88k/s) Relation(2881850 275.04/s)
Standard exception processing relation id=4180656: TopologyException: side location conflict at 272703.48999999999 5093940.6600000001
Processing: Node(2715748k 2596.3k/s) Way(272021k 51.88k/s) Relation(3100780 281.58/s)
Standard exception processing relation id=4542207: TopologyException: side location conflict at 15042893.51 4046892.75
Processing: Node(2715748k 2596.3k/s) Way(272021k 51.88k/s) Relation(3109640 282.05/s)  parse time: 17314s
Node stats: total(2715748725), max(3326469484) in 1046s
[...]

PR #353, opened 2015-05-24
node-cache: fix prefetching

lonvia:
Blocks should be prefetched when they are not in the ram cache.

Might or might not be related to #351.

### DIFF
--- a/node-persistent-cache.cpp
+++ b/node-persistent-cache.cpp
@@ -335,8 +335,7 @@ int node_persistent_cache::load_block(osmid_t block_offset)
                 strerror(errno));
         exit(1);
     }
-    add_to_cache_idx(cache_index_entry(readNodeBlockCache[block_id].block_offset,
-                                       block_id));
+    add_to_cache_idx(cache_index_entry(block_offset, block_id));
 
     return block_id;
 }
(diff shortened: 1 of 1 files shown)


pnorman:
If this fixes #351 it warrants a new release

node-persistent-cache.cpp
```diff
@@ -532,10 +531,10 @@ int node_persistent_cache::get_list(nodelist_t &out, const idlist_t nds)
 {
     out.assign(nds.size(), osmNode());
 
-    bool need_fetch = false;;
+    bool need_fetch = false;
     for (size_t i = 0; i < nds.size(); ++i) {
         /* Check cache first */
-        if (ram_cache && (ram_cache->get(&out[i], nds[i]) != 0)) {
+        if (!ram_cache || (ram_cache->get(&out[i], nds[i]) == 0)) {
```
pnorman: This line is the one causing the problems described in https://github.com/openstreetmap/osm2pgsql/issues/351#issuecomment-105362202

lonvia: So ti turns out, I got the semantics of ram_cache->get() wrong. It actually returns 0 when it finds a node. (Fix will follow in a bit.) The consequence is that all nodes are looked up in the flat node store. If that makes the memory hogging worse, the flat node store must be the culprit. Investigating.

PR #354, opened 2015-05-24
Remove oldmerc option

### LINKED ISSUE #64: 3395 called "broken", but why?
The Readme says:

> Depending on the command-line switches you can select which projection you
> want the database in. You have three choices:
> 
> 4326: The standard lat/long coordinates
> 900913: The spherical Mercator projection, used by TileCache, Google Earth etc.
> 3395: The legacy (broken) WGS84 Mercator projection

It is unclear why [EPSG:3395](http://spatialreference.org/ref/epsg/3395/) is considered "broken" relative to [EPSG:4326](http://spatialreference.org/ref/epsg/4326/). One should clarify that point.

 One should clarify this point.

lonvia:
fixes #64

### DIFF
--- a/docs/osm2pgsql.1
+++ b/docs/osm2pgsql.1
@@ -78,9 +78,6 @@ Store data in degrees of latitude & longitude.
 \fB\-m\fR|\-\-merc
 Store data in proper spherical Mercator (the default).
 .TP
-\fB\-M\fR|\-\-oldmerc
-Store data in the legacy OSM Mercator format.
-.TP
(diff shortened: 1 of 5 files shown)


pnorman:
[osm2pgsql.1](https://github.com/openstreetmap/osm2pgsql/blob/8f59f1b5f2bedcf64132b8167c11b411166079f7/docs/osm2pgsql.1#L80-L82) also needs editing

lonvia:
I've updated osm2pgsq.1.

PR #356, opened 2015-05-25
Cleanup includes and FIXED_POINT define

lonvia:
* cleans up includes as reported by [include-what-you-use](https://code.google.com/p/include-what-you-use)
 * removes rb.cpp/.hpp which is no longer used
 * makes FIXED_POINT configurable (mostly so the define can be moved to config.h where it is much more obvious that this is a global define)


pnorman:
travis is failing the clang builds with
```
In file included from parse.cpp:5:
./parse-pbf.hpp:42:8: error: definition of type 'ProtobufCBinaryData' conflicts
      with typedef of the same name
struct ProtobufCBinaryData;
       ^
/usr/include/google/protobuf-c/protobuf-c.h:94:37: note: 'ProtobufCBinaryData'
      declared here
typedef struct _ProtobufCBinaryData ProtobufCBinaryData;
                                    ^
1 error generated.
```

and the gcc builds with

```
/usr/include/c++/4.6/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the upcoming ISO C++ standard, C++0x. This support is currently experimental, and must be enabled with the -std=c++0x or -std=gnu++0x compiler options.
```

lonvia:
The clang error looks like it was due to an ancient version of protobuf. All fixed now and passing. I've added a configure option for myself, so I can run the compiler without C++11 for testing purposes.

PR #360, opened 2015-06-16
Reduce memory usage of node cache

lonvia:
Reduces preallocation of memory and implements more dense management structures for the ram node cache. For a planet import, the reduction in total memory is around 800MB. This helps a bit with #351 but won't completely fix it.

Passes the tests but still needs to be tested more thoroughly with different variations of ram cache configuration before merging.

### DIFF
--- a/node-persistent-cache.cpp
+++ b/node-persistent-cache.cpp
@@ -235,8 +235,16 @@ int node_persistent_cache::load_block(osmid_t block_offset)
         readNodeBlockCache[block_id].dirty = 0;
     }
 
-    remove_from_cache_idx(readNodeBlockCache[block_id].block_offset);
-    new(readNodeBlockCache[block_id].nodes) ramNode[READ_NODE_BLOCK_SIZE];
+    if (readNodeBlockCache[block_id].nodes) {
+        remove_from_cache_idx(readNodeBlockCache[block_id].block_offset);
+        new(readNodeBlockCache[block_id].nodes) ramNode[READ_NODE_BLOCK_SIZE];
+    } else {
+        readNodeBlockCache[block_id].nodes = new ramNode[READ_NODE_BLOCK_SIZE];
+        if (!readNodeBlockCache[block_id].nodes) {
+            fprintf(stderr, "Out of memory: Failed to allocate node read cache\n");
+            util::exit_nicely();
+        }
+    }
     readNodeBlockCache[block_id].block_offset = block_offset;
     readNodeBlockCache[block_id].used = READ_NODE_CACHE_SIZE;
 
@@ -613,11 +621,7 @@ node_persistent_cache::node_persistent_cache(const options_t *options, int appen
     }
     for (i = 0; i < READ_NODE_CACHE_SIZE; i++)
     {
-        readNodeBlockCache[i].nodes = new ramNode[READ_NODE_BLOCK_SIZE];
-        if (!readNodeBlockCache[i].nodes) {
-            fprintf(stderr, "Out of memory: Failed to allocate node read cache\n");
-            util::exit_nicely();
-        }
+        readNodeBlockCache[i].nodes = 0;
         readNodeBlockCache[i].block_offset = -1;
         readNodeBlockCache[i].used = 0;
         readNodeBlockCache[i].dirty = 0;
@@ -659,7 +663,8 @@ node_persistent_cache::~node_persistent_cache()
     if (readNodeBlockCache) {
         for (int i = 0; i < READ_NODE_CACHE_SIZE; i++)
         {
-            delete[] readNodeBlockCache[i].nodes;
+            if (readNodeBlockCache[i].nodes)
+                delete[] readNodeBlockCache[i].nodes;
         }
         delete[] readNodeBlockCache;
     }
--- a/node-persistent-cache.cpp
+++ b/node-persistent-cache.cpp
@@ -52,7 +52,7 @@ void node_persistent_cache::writeout_dirty_nodes()
         if (readNodeBlockCache[i].dirty())
         {
             if (lseek64(node_cache_fd,
-                    (readNodeBlockCache[i].block_offset
+                    ((osmid_t) readNodeBlockCache[i].block_offset
                             << READ_NODE_BLOCK_SHIFT)
                             * sizeof(ramNode)
                             + sizeof(persistentCacheHeader),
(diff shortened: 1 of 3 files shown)


pnorman:
What kind of tests need running with it?

lonvia:
Basically all possible combinations of --slim, --flat-nodes and --cache-strategy. The latter in particular is badly covered with the regression tests.

pnorman:
I'm not sure we need to wait for what I'm describing to merge, but here are my plans. For valid combinations of slim, flat nodes, cache strategy, and `--cache` being zero or not, do the processing section of an import with the null backend. Then use the C++ api to fetch positions from the middle.

The test file would have a full dense block, a partial dense block, and some sparse nodes.

Out of order and negative nodes would be nice to test, but as the correct behaviour with some combinations of options (`--flat-nodes --slim --cache 0`) is to throw an exception, this needs to be its own file.

lonvia:
For --cache there are actually three possibility: no cache, fits partly, fits all. I wonder if --cache=0 won't trip the code somewhere.

@pnorman  could you run your planet import again using this code in the meantime.

pnorman:
```
LD_LIBRARY_PATH=$HOME/boost/lib time ~/osm2pgsql/osm2pgsql -d gis \
  --cache 12000 --flat-nodes nodes.bin --slim --drop \
  --number-processes 8 -S ~/openstreetmap-carto/openstreetmap-carto.style \
  planet-150202.osm.pbf

...

Processing: Node(2715740k 2538.1k/s) Way(0k 0.00k/s) Relation(0 0.00/s)Switching to read mode
Switching to read mode done
Warning: reading node outside node cache. (18446744072741715967 vs. 200542)
Warning: reading node outside node cache. (18446744072741715967 vs. 274057218)
Warning: reading node outside node cache. (18446744072741715967 vs. 1024965354)
Warning: reading node outside node cache. (18446744072741715967 vs. 200550)
Warning: reading node outside node cache. (18446744072741715967 vs. 1024940305)
Warning: reading node outside node cache. (18446744072741715967 vs. 1024940306)
Warning: reading node outside node cache. (18446744072741715967 vs. 1024940307)
Warning: reading node outside node cache. (18446744072741715967 vs. 200551)
Warning: reading node outside node cache. (18446744072741715967 vs. 200553)
Failed to seek to correct position in node cache: Invalid argument
Error occurred, cleaning up
Command exited with non-zero status 1
```

lonvia:
The error was caused by an integer overflow due to missing conversions from 32 to 64 bit. This is fixed now.

Planet import manages to get past the initial import but in the processing phase it needs a impressive 44GB of RAM (with a total of 25GB of swap being used). So this is far away from fixing anything in #351.

springmeyer:
> The error was caused by an integer overflow due to missing conversions from 32 to 64 bit. This is fixed now.

Nice job fixing! Not sure how you tracked the problem down but just wanted to add that I've been enjoying using `clang++ -Wshorten-64-to-32` to uncover these types of bugs in mapnik (https://github.com/mapnik/mapnik/issues/2907). So if you've not heard of / or are already using that flag you might want to.

pnorman:
How does this look for coverage of possible cases?

```python
#!/usr/bin/env python

# This script generates an OSM XML file for testing node storage.

from lxml import etree
from sys import stdout

# Constants from node-ram-cache
BLOCK_SHIFT = 10
PER_BLOCK = 1 << BLOCK_SHIFT

def node(id, lat=None):
  if lat is None:
    lat=1+1e-6*id
  return etree.Element("node", {'id':('%d' % id), 'visible':'true', 'version':'1', 
                       'changeset':'1', 'timestamp':'2010-01-01T00:00:00Z',
                       'lon':'0.0', 'lat':('%f' % lat)})

root = etree.Element("osm", {'version':'0.6', 'generator':'generate_storage_xml.py'})

root.append(etree.Comment("This file contains nodes arranged to test both dense and sparse node storage in the optimized storage"))

root.append(etree.Comment("2 dense blocks, the second partially filled at the start"))
for id in xrange(1, PER_BLOCK+(PER_BLOCK >> 1) +1):
  root.append(node(id))

# 1 dense block, 75% filled
root.append(etree.Comment("1 dense block, partially filled"))
for id in xrange(PER_BLOCK*2+1, PER_BLOCK*3, 4):
  root.append(node(id))
  root.append(node(id+1))
  root.append(node(id+3))

root.append(etree.Comment("1 dense block worth of sparse nodes"))

for id in xrange(PER_BLOCK*3, PER_BLOCK*4, 4):
  root.append(node(id))

#skip a bit, and a lone node
root.append(etree.Comment("A lone sparse node"))
root.append(node(PER_BLOCK*5))

# A dense block of alternating positions of zero/non-zero
root.append(etree.Comment("A dense block to check 0,0 is stored correctly"))
for id in xrange(PER_BLOCK*6, PER_BLOCK*7, 2):
  root.append(node(id, 0))
  root.append(node(id+1))

root.append(etree.Comment("Sparse nodes to check 0,0 is stored correctly"))

root.append(node(PER_BLOCK*7, 0))
root.append(node(PER_BLOCK*7+1))
  
stdout.write(etree.tostring(root, pretty_print=True))
stdout.write('\n')
```

lonvia:
Looks good.

pnorman:
The import which failed before works now.


```
...
node cache: stored: 1484575472(54.67%), storage efficiency: 94.39% (dense blocks: 1406958, sparse nodes: 66070009), hit rate: -114.53%

Osm2pgsql took 33536s overall
```

lonvia:
That hit rate computation looks wrong. Will check it out.

Any degradation in terms of speed? Also comparison figures for number of dense blocks and sparse nodes might be interesting if you have them.

pnorman:
I'll get numbers after the dentist

pnorman:
I'm testing 7db0ae9 for speed

pnorman:
So, 
```
node cache: stored: 1484575472(54.67%), storage efficiency: 94.39% (dense blocks: 1406958, sparse nodes: 66070009), hit rate: -114.50%

Osm2pgsql took 31555s overall
```

So, the hit rate is marginally different, but this branch is 6% slower.

Looking in more details, 

```
                #360   master
processing n     820      809
processing w    6110     5766
processing r   11284    10568
pending ways    8125     7419
indexes         7083     6907
total          33536    31555
```

Processing node differences are within normal variations.

Because I ran both with `time` I have those numbers too

#360
```
52134.89user 3594.71system 9:18:56elapsed 166%CPU (0avgtext+0avgdata 16419744maxresident)k
294242648inputs+913034600outputs (56448major+4575021minor)pagefaults 0swaps
```

master
```
47005.27user 2260.20system 8:45:55elapsed 156%CPU (0avgtext+0avgdata 16459960maxresident)k
324145584inputs+52018888outputs (28397major+1744337minor)pagefaults 0swaps
```

pnorman:
Rerunning with this branch,

```
node cache: stored: 1484575472(54.67%), storage efficiency: 94.39% (dense blocks: 1406958, sparse nodes: 66070009), hit rate: -114.49%

Osm2pgsql took 33522s overall
52263.25user 3572.23system 9:18:42elapsed 166%CPU (0avgtext+0avgdata 16410072maxresident)k
294721152inputs+913607960outputs (78090major+4679810minor)pagefaults 0swaps
```

Basically the same overall time

lonvia:
I'm in favour of merging this. The performance numbers might be as well due to swapping artifacts.

PR #379, opened 2015-07-09
Initial libosmium port

### LINKED ISSUE #213: middle_ram_t::relations_using_way is unimlpemented
Hello,

We're trying to load north-america's pbf using the following command:
```
osm2pgsql -a -k -E EPSG:4326  -H localhost -U postgres -W -d osm_test1 /tmp/north-america-latest.osm.pbf -C 12000

Upon ways processing, the following message is thrown and the load is aborted:

Using 1 helper-processes
terminate called after throwing an instance of 'std::runtime_error'
  what():  middle_ram_t::relations_using_way is unimlpemented, and should not have been called. This is probably a bug, please report it at https://github.com/openstreetmap/osm2pgsql/issues
Aborted
```

I've looked at middle_ram_t definition and the execption at https://github.com/openstreetmap/osm2pgsql/blob/master/middle-ram.cpp and seems weird to me that it mentions that relations_using_ways is only used in slim mode which in the command above isn't used.

The version of osm2pgsql I'm using is: osm2pgsql SVN version 0.87.0 (64bit id space)

Is this a bug or I'm a missing something?

lonvia:
This introduces libosmium as another parser back-end, replacing the pbf and xml parser. The o5m parser back-end remains for the time being and should still be usable. Once o5m support has been added, the libosmium integration can be taken a bit further and the custom OSM data structures should be replaced with libosmium data structures which would avoid quite a bit of copy/malloc.

I've copied the libosmium sources because none of the distros provide a usable version and won't for some time. Maybe somebody has a better idea how to handle that.

Other notable changes include:

- refactor of parse.hpp/cpp
- retire UTFsanitizer and remove sanitize option
- make C++11 compiler mandatory (required by libosmium)
- update pbf proto descriptions
- resolves #213 by sending all data as 'modify' in append mode (going back to the previous behaviour of the C version)

CC @joto


pnorman:
I'd like to tag 0.88.0 as a new stable, then move to 0.89.0-dev with this merge. That'll give us a stable release/branch without C++11.

Given the current state of libosmium packaging and that it's header only, I agree with copying the sources.

I'll throw some full planet tests at this branch on the weekend, see if it turns up any problems. A huge thanks for this big work, and I owe you a can of maple syrup (I'm Canadian).

pnorman:
`protobuf-c-compiler` and `libprotobuf-c0-dev` are still in configure.ac. Are they still needed, with PBF reading moved to libosmium?

lonvia:
Yes, libprotobuf is a dependency for libosmium. The compiler is needed because pbf proto files are still compiled as part of the osm2pgsql built. The reason is that the libosmpbf packages that are distributed with Ubuntu/Debian are buggy.

pnorman:
`protobuf-c-compiler` is part of the pure C implementation of protobuf. `protobuf-compiler` is the normal compiler for proto files

lonvia:
Ah, you mean the mention in the `README`. True, I forgot to change that. Fixed now. `configure.ac` already checks for the correct C++ versions.

pnorman:
configure says "Look for packages named: libprotobuf-dev protobuf-c-compiler" if PBF support is missing.

pnorman:
:shipit:

pnorman:
I'm getting an error building on my Ubuntu test box

```
autoreconf: running: /usr/bin/autoconf --force
configure.ac:34: error: possibly undefined macro: AC_CHECK_PROG
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
configure.ac:37: error: possibly undefined macro: AC_MSG_ERROR
configure.ac:86: error: possibly undefined macro: AS_IF
autoreconf: /usr/bin/autoconf failed with exit status: 1
```

The relevant part of configure.ac is
```autoconf
dnl Find C++ compiler
AC_PROG_CXX
AC_CHECK_PROG(HAVE_CXX, $CXX, yes, no)
if test "$HAVE_CXX" = "no" 
then
  AC_MSG_ERROR([Could not find a c++ compiler]);
fi
```
https://github.com/lonvia/osm2pgsql/blob/libosmium-port/configure.ac#L32-L38

pnorman:
> I'm getting an error building on my Ubuntu test box

Looks like we've added a dependency on `pkg-config`. http://lists.freedesktop.org/archives/harfbuzz/2011-January/001042.html gave me the clue

PR #394, opened 2015-07-15
use vectors of pointers in middle-ram to reduce memory footprint

### LINKED ISSUE #381: Non-slim mode uses too much memory
8d803dc4d72f155dd89968916da12de297280c7f introduced a regression for middle-ram where it needs much more memory for its management structs than before. I'm no longer able to import Switzerland on a 8GB machine in non-slim mode.

I can provide a fix that reverts to having pointer arrays with dynamically allocated vectors (at the price of being marginally slower) but this will require C++11, so the libosmium branch needs to come first.

lonvia:
fixes #381

### DIFF
--- a/middle-ram.hpp
+++ b/middle-ram.hpp
@@ -13,9 +13,72 @@
 
 #include "middle.hpp"
 #include <vector>
+#include <array>
 
 struct node_ram_cache;
 struct options_t;
+
+template <typename T, size_t N>
+class cache_block_t
+{
+    std::array<std::unique_ptr<T>, N> arr;
+public:
+    void set(size_t idx, T *ele) { arr[idx].reset(ele); }
+
+    T const *get(size_t idx) const { return arr[idx].get(); }
+};
+
+template <typename T, size_t BLOCK_SHIFT>
+class elem_cache_t
+{
+    constexpr static size_t per_block() { return 1 << BLOCK_SHIFT; }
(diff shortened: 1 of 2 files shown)


middle-ram.hpp
```diff
@@ -61,15 +61,19 @@ struct middle_ram_t : public middle_t {
...
     struct ramRel {
         taglist_t tags;
         memberlist_t members;
+
+        ramRel(const taglist_t &t, const memberlist_t &m) : tags(t), members(m) {}
     };
 
-    std::vector<std::vector<ramWay> > ways;
-    std::vector<std::vector<ramRel> > rels;
+    std::vector<std::unique_ptr<std::vector<std::unique_ptr<ramWay> > > > ways;
```
pnorman: this is a vector of pointers to vectors of pointers to ways?

I'm wondering if it makes sense to typedef `std::vector<std::unique_ptr<ramWay> >` for clarity that it's a block of ways

pnorman:
Do you think this needs any full planet testing? It looks okay to me.

lonvia:
I doubt it's possible to still run non-slim on a full planet.

Redoing this with proper classes right now.

pnorman:
> I doubt it's possible to still run non-slim on a full planet

Oh yes, good point.

lonvia:
This is ready now. Tested with `./osm2pgsql -O null -C 16000 germany.osm.pbf`.

PR #440, opened 2015-08-17
Fixes to restore MSVC compatibilty

alex85k:
Recompilng on Visual Studio 2015 on Windows 10 x64 resulted in several errors.
Most of them were from my cmake scripts (now fixed), but two problematic places were found in the code.

### DIFF
--- a/middle-ram.hpp
+++ b/middle-ram.hpp
@@ -45,7 +45,7 @@ class elem_cache_t
         return id & (per_block()-1);
     }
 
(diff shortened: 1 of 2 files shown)


lonvia:
Thanks for testing on MSVC.

alex85k:
Thank you too for support!
All the tests pass now except one json-realted (will post logs later), the cmake build scripts are here:
https://github.com/alex85k/osm2pgsql-cmake

PR #452, opened 2015-09-28
Manpage: fix abbreviation for --output

MaxSem:


pnorman:
backported for next 0.88.x release

PR #454, opened 2015-10-02
Add area reprojection facility

woodpeck:
This adds a command line option --reproject-area which is useful for those who do not import data in EPSG:3857, but a different projection, but who still want to use one of the open styles available for standard EPSG:3857 databases.

These styles will often rely on the way_area column in the polygon table for rendering decisions (e.g. draw larger lakes on smaller zoom levels, and only bring in smaller lakes later). Because way_area was, until now, computed in target projection units, styles written for EPSG:3857 would not work with a database imported, say, with the -l (ell) flag.

You can now use the --reproject-area flag which projects every polygon to EPSG:3857 temporarily to compute its way_area. 

This patch should not negatively affect the standard use case, but it will take more time on import if --reproject-area is used together with -l (ell) or -E.

A simple test is provided, covering the -l (ell) use case but not more exotic projections that should, in theory, also work, provided that the proj library is able to project from the target projection to EPSG:3857 which is not always guaranteed.


geometry-builder.cpp
```diff
@@ -760,8 +759,75 @@ void geometry_builder::set_exclude_broken_polygon(int exclude)
...
+        double area = 0.0;
+        for (std::size_t i=0; i<multi->getNumGeometries(); i++)
+        {
+            area += getArea(multi->getGeometryN(i));
+        }
+        return area;
+    }
+
+    const geos::geom::Polygon *poly = dynamic_cast<const geos::geom::Polygon *>(geom);
+    if (!poly) return 0.0;
```
pnorman:
```c++
if (!poly) {
    return 0.0;
}
```

geometry-builder.cpp
```diff
@@ -760,8 +759,75 @@ void geometry_builder::set_exclude_broken_polygon(int exclude)
...
+
+    // standard polygon - reproject, then compute area of outer - areas of holes
```
pnorman: Why compute the area by doing hole subtraction instead of [getArea from geos::geom::Polygon](http://geos.osgeo.org/doxygen/classgeos_1_1geom_1_1Polygon.html#a7fdaf9550f08f56340b0ce8350501b66)

tests/area-reprojection.osm
```diff
@@ -0,0 +1,27 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<osm version="0.6" generator="Osmosis 0.43.1">
+  <node id="1" version="1" timestamp="1980-01-01T00:00:00Z" uid="1" user="user" changeset="1" lat="83.6397141" lon="0"/>
```
pnorman: You can strip out timestamp, uid, user, and changeset.

tests/test-area-reprojection.cpp
```diff
@@ -0,0 +1,114 @@
...
+        exit(77);
+    }
+
+    std::shared_ptr<middle_pgsql_t> mid_pgsql(new middle_pgsql_t());
+    options_t options;
+    options.database_options = db->database_options;
+    options.num_procs = 1;
+    options.style = "default.style";
+    options.prefix = "osm2pgsql_test";
+    if (latlon) options.projection.reset(new reprojection(PROJ_LATLONG));
```
pnorman: These should also be [1TBS style](https://github.com/openstreetmap/osm2pgsql/blob/master/CONTRIBUTING.md#code-style)

tests/test-area-reprojection.cpp
```diff
@@ -0,0 +1,114 @@
...
+    test_area_base(true, true, 6.66e+10);
+}
+
+int main(int argc, char *argv[]) {
+    RUN_TEST(test_area_latlon);
+    RUN_TEST(test_area_classic);
+    RUN_TEST(test_area_latlon_with_reprojection);
+    return 0;
+}
+
```
pnorman: This looks like an extra endline in the file, but github might be confusing me

reprojection.cpp
```diff
@@ -162,6 +164,42 @@ void reprojection::reproject(double *lat, double *lon) const
...
+        return;
+
+    if (Proj == PROJ_LATLONG)
+    {
+        if (*lat > 85.07)
+            *lat = 85.07;
+        if (*lat < -85.07)
+            *lat = -85.07;
+
+    //std::cout << "builtin=" << Proj << ": " << *lon ;
```
pnorman: The debugging statements should be removed

reprojection.cpp
```diff
@@ -9,6 +9,8 @@
 
 #include <cstdlib>
 #include <cstdio>
+#include <assert.h>
+#include <iostream>
```
pnorman: are either of these headers used in the new code?

reprojection.cpp
```diff
@@ -162,6 +164,42 @@ void reprojection::reproject(double *lat, double *lon) const
     *lon = x[0];
 }
 
+void reprojection::target_to_tile(double *lat, double *lon) const
```
pnorman: Please add a comment above explaining what the function does, see ``reprojection::coords_to_tile`` for an example

geometry-builder.cpp
```diff
@@ -760,8 +759,78 @@ void geometry_builder::set_exclude_broken_polygon(int exclude)
...
+        return area;
+    }
+
+    const geos::geom::Polygon *poly = dynamic_cast<const geos::geom::Polygon *>(geom);
+    if (!poly) {
+        return 0.0;
+    }
+
+    // standard polygon - reproject rings individually, then assemble polygon and
+    // compute area.
```
pnorman: Does reprojecting `poly` and using `poly->getArea()` work?

woodpeck: Thing is we can only reproject individual coordinate pairs, and the wrapper I built will dissect a polygon into rings, the rings into coordinate pairs, reproject them, and reassemble rings and a polygon from that. The wrapper does not have the capability to reproject a multipolygon; this would require iterating through the getGeometryNs of the multipolygon, reprojecting the individual members, and then calling the GEOS factory to build a multipolygon from all that in the end. It would then be possible to call getArea on the multipolygon. But the building of the multipolygon would IMHO be unnecessary extra code and extra memory allocation/release.

reprojection.cpp
```diff
@@ -163,6 +163,46 @@ void reprojection::reproject(double *lat, double *lon) const
...
+        *lon = (*lon) * EARTH_CIRCUMFERENCE / 360.0;
+        *lat = log(tan(M_PI/4.0 + (*lat) * DEG_TO_RAD / 2.0)) * EARTH_CIRCUMFERENCE/(M_PI*2);
+        return;
+    }
+
+    x[0] = *lon;
+    y[0] = *lat;
+    z[0] = 0;
+
+    /** end of "caution" section. */
```
pnorman: Where's the start of the caution section? This might be refering to something in `reprojection::reproject`

pnorman:
did woodpeck#1 to fix some code formatting/style issues.

geometry-builder.cpp
```diff
@@ -760,8 +759,71 @@ void geometry_builder::set_exclude_broken_polygon(int exclude)
...
+ */
+double geometry_builder::getArea(const geos::geom::Geometry *geom) const
+{
+    // reprojection is not necessary, or has not been asked for.
+    if (!reprojection) {
+        return geom->getArea();
+    }
+
+    // MultiPolygon - return sum of individual areas
+    if (const geos::geom::MultiPolygon* multi = dynamic_cast<const geos::geom::MultiPolygon *>(geom)) {
```
pnorman: This code path doesn't seem to be tested

woodpeck: The "if !reprojection" path is tested in https://github.com/openstreetmap/osm2pgsql/pull/454/files#diff-d5b71a720f93c25bd9295e207018acdcR101 and https://github.com/openstreetmap/osm2pgsql/pull/454/files#diff-d5b71a720f93c25bd9295e207018acdcR105 - is it not?

pnorman: this is the multipolygon case, https://github.com/openstreetmap/osm2pgsql/pull/454/files#diff-853d1c6f0a63324f82abcb98373e8bc5 only contains a simple closed way.

pnorman:
Although I do find the test rather clever :star: it might be easier to just have a rectangle and a rectangle with a rectangular hole as tests. This would allow calculation of the appropriate areas by hand.

geometry-builder.cpp
```diff
@@ -760,8 +759,71 @@ void geometry_builder::set_exclude_broken_polygon(int exclude)
...
+        double area = 0.0;
+        for (std::size_t i=0; i<multi->getNumGeometries(); i++) {
+            area += getArea(multi->getGeometryN(i));
+        }
+        return area;
```
pnorman: @lonvia we don't currently have any usage of `std::accumulate` or lambdas, but do you think this should be 
```c++
return std::accumulate(multi->begin(), multi->end(), 0.0,
                       [=](double a, const geos::geom::Geometry *geom) { return a + getArea(geom); });
```

lonvia: In this particular case, a loop would mean considerably less writing:

    for (auto g : multi) {
      area += getArea(g);
    }

Personally, I find that more readable. However, in general, using the standard library function and lambdas in place of hand-crafted for loops is a good thing. So sure, go ahead and use it.

(Also: please use auto inside the if condition.)

reprojection.cpp
```diff
@@ -163,6 +163,43 @@ void reprojection::reproject(double *lat, double *lon) const
...
+
+    pj_transform(pj_target, pj_tile, 1, 1, x, y, z);
+
+    *lat = y[0];
+    *lon = x[0];
```
lonvia: x and y are unnecessary here. Simply use lat and lon as parameters to pj_transform.

Also: the stray 'caution' comment still needs to be deleted.

geometry-builder.cpp
```diff
@@ -760,8 +760,68 @@ void geometry_builder::set_exclude_broken_polygon(int exclude)
...
+        return 0.0;
+    }
+
+    // standard polygon - reproject rings individually, then assemble polygon and
+    // compute area.
+
+    const auto *ext = poly->getExteriorRing();
+    auto *projectedExt = reproject_linearring(ext);
+    auto nholes = poly->getNumInteriorRing();
+    std::vector<geos::geom::Geometry *> *projectedHoles = new std::vector<geos::geom::Geometry *>(nholes);
```
lonvia: Wrap in std::unique_ptr<> and use release() below in createPolygon() to ensure memory gets clean up when an exception is thrown.

geometry-builder.cpp
```diff
@@ -760,8 +760,68 @@ void geometry_builder::set_exclude_broken_polygon(int exclude)
...
+    return area;
+}
+
+/**
+ * Reprojects given Linear Ring from target projection to spherical mercator.
+ * Caller takes ownership of return value.
+ */
+geos::geom::LinearRing* geometry_builder::reproject_linearring(const geos::geom::LineString *ls) const
+{
+    std::vector<geos::geom::Coordinate> *projectedCoords = new std::vector<geos::geom::Coordinate>();
```
lonvia: Dito, use unique_ptr.

pnorman:
Replaced by #506.

PR #455, opened 2015-10-08
Ignore nodes with invalid coordinates.

zerebubuth:
I hit this while processing a snapshot extracted from history which happened to contain [version 3 of node 7226378](http://www.openstreetmap.org/node/7226378/history), which has coordinates (83.4542914, 214.7483647). The API at the time didn't check this, and it has since been fixed, but it means that [the node parsing code](https://github.com/openstreetmap/osm2pgsql/blob/master/parse-osmium.cpp#L64-L65) will throw an exception.

(Notably, this doesn't kill the process, as it means the consumer thread waits on the Queue to finish processing, while the producer thread is waiting on the Queue to empty. But this is a separate issue.)

In this case it seems simple enough to just ignore any node which has an invalid location, which is effectively the same behaviour as setting the bbox to the world extent.

### DIFF
--- a/parse-osmium.cpp
+++ b/parse-osmium.cpp
@@ -67,7 +67,7 @@ void parse_osmium_t::node(osmium::Node& node)
     if (!node.location().valid()) {
       fprintf(stderr, "WARNING: Node %" PRIdOSMID " (version %ud) has an invalid "
               "location and has been ignored. This is not expected to happen with "
-              "recent planet files, so please check that your input is correct.",
+              "recent planet files, so please check that your input is correct.\n",
               node.id(), node.version());
 
       return;
--- a/parse-osmium.cpp
+++ b/parse-osmium.cpp
@@ -61,6 +61,11 @@ parse_osmium_t::stream_file(const std::string &filename, osmdata_t *osmdata)
 
 void parse_osmium_t::node(osmium::Node& node)
 {
+    // if the node is not valid, then node.location.lat/lon() can throw.
+    // we probably ought to treat invalid locations as if they were
+    // deleted and ignore them.
+    if (!node.location().valid()) { return; }
+
     double lat = node.location().lat();
     double lon = node.location().lon();
     if (bbox.inside(lat, lon)) {
--- a/parse-osmium.cpp
+++ b/parse-osmium.cpp
@@ -64,7 +64,14 @@ void parse_osmium_t::node(osmium::Node& node)
     // if the node is not valid, then node.location.lat/lon() can throw.
     // we probably ought to treat invalid locations as if they were
     // deleted and ignore them.
-    if (!node.location().valid()) { return; }
+    if (!node.location().valid()) {
+      fprintf(stderr, "WARNING: Node %" PRIdOSMID " (version %ud) has an invalid "
+              "location and has been ignored. This is not expected to happen with "
+              "recent planet files, so please check that your input is correct.",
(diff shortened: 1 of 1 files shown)


pnorman:
@joto, is this a check that should make it into libosmium?

zerebubuth:
IMO, libosmium is doing the right thing; presenting the node to the application, but allowing the application to decide the policy of what to do with nodes that have invalid locations. For some applications which don't care about location (i.e: just tags or users or something) it would be incorrect for libosmium to exclude those nodes.

joto:
@pnorman: @zerebubuth is right. This is why I specifically added this `valid()` function, so the user can decide what to do with the messed up data. (This is one of the many strange corner cases you run into with OSM data that's unfortunately not documented anywhere.)

lonvia:
Given that invalid nodes don't happen in current planets, I'd prefer the solution where you explicitly give a whole-world bbox when working with a buggy planet instead of always forcing the check on every single node. Naturally that means that osm2pgsql should give a meaningful error instead of just hanging.

zerebubuth:
@lonvia, sorry, I didn't explain that very well... I meant it's _conceptually_ the same as having a whole-world bounding box. The libosmium code already checks for an invalid location on each call to `lat()` and `lon()`, and it's exactly that check throwing an exception which was causing osm2pgsql to fail for me.

In 382c5f3 I've bypassed the check on `lat()` and `lon()`, as the code already explicitly checks `valid()`, and this saves us one check (although the check is almost free). However, I'm not as confident that current planets or extracts aren't buggy and, in my opinion, it's better to be defensive and keep the checks.

Possibly it ought to have a warning; "Oops, this input is buggy, you might not get the results you expect.", though? What do you think?

lonvia:
@zerebubuth I see. I'm not sure how to understand your comment about the process not being killed then. Node parsing runs in a single thread, so that would have to be a libosmium bug then.

The new version looks ok. Side note: we really should replace the custom bbox class with osmium::box. It's faster and would even allow us to unify the valid() and inside() check.

lonvia:
And +1 for adding a warning.

zerebubuth:
Added a warning in 2ee1955.

@lonvia what's happening is that libosmium is starting some threads to handle input, on my system it looks like there's 3 in addition to the main thread. It seems that one is reading PBF blocks from disk and putting them onto a `Queue` while the main thread is popping - possibly from a different `Queue`, I'm not quite sure. When the exception occurs in the parsing method on the main thread, it causes the stack to unwind and the `Reader`'s destructor to be called, which sets a flag and waits for the threads it has started to finish. However, the thread which is reading PBF blocks is blocked on pushing to the `Queue`, so doesn't terminate and the system deadlocks.

It's not clear to me whether the responsibility for handling exceptions lies with libosmium (i.e: `Reader::close()` handles early termination), or whether there's something that osm2pgsql should be doing to catch the exception and indicate a fatal condition. In either case, it's a separate issue from ignoring those nodes with invalid locations. @joto - should I file something against libosmium for this, or should osm2pgsql be doing something to handle early termination?

joto:
@zerebubuth libosmium should handle this magically behind the scenes. If it doesn't, than that is a bug. Please open a bug against libosmium with a way to reproduce this, if you can.

lonvia:
Merged now. For further discussion on the issue with hanging threads I've opened https://github.com/osmcode/libosmium/issues/131.

PR #458, opened 2015-10-11
Clean up after flat nodes tests

zerebubuth:
The flat nodes tests each create 20GB flat nodes files and didn't remove them when the tests were finished, which would fail on my system which has "only" 35GB free disk space. This PR adds an RAII cleanup object to remove the files when the tests are complete, reducing the amount of disk space needed to "only" 20GB if the tests are run serially.


pnorman:
With ^C I was able to interrupt tests/test-middle-flat and leave behind the flat nodes file.

zerebubuth:
Yeah, RAII cleanup won't catch things which are "out of band" and kill the process from the outside, like ^C, SIGKILL or the OOM killer. To deal with those, we'd need to have a test wrapper script to clean up resources. The RAII approach, which we're using for the test databases at the moment too, works in 99% of cases. The other 1% might need more work.

Issue #398, opened 2015-07-17
Try cppformat as a replacement for boost:format


lonvia:
The boost::format library is known to have a very bad performance. A recommended replacement is the [cppformat library](https://github.com/cppformat/cppformat). We should try that.

pnorman:
I thought that we weren't using it anywhere that mattered for performance, but found a lot more than I thought with a grep.

It looks like cppformat **just** made it into Debian. Would we be looking at including it like libosmium and compiling it in header-only mode?

lonvia:
Given that it is relatively new and still in development, including it as a contrib library will give as more control. So I would prefer that.

If the performance turns out well, it would be nice to be able to replace the remaining hacky printfs in the code that creates SQL statements.

pnorman:
Including an unrelated third-party library often causes problems with package maintainers, so we should check with them before planning that

Issue #202, opened 2014-10-25
Use a logging library


pnorman:
Currently everything prints out with fprintf. This makes it hard to change the verbosity.

Any thoughts?

@zerebubuth, do you think [MapQuest/logging](https://github.com/MapQuest/logging) would be suited here?

pnorman:
boost::log was only introduced in 1.54. This is a big minus.

@zerebubuth, I can't remember the options we came up with at the hack weekend.

zerebubuth:
There's google's logging library too. I've never used it, though, so I
can't say whether it's any good.

The MQ logging library is probably no good for now, as it currently depends
on mapnik and that's too big of a dep to pull in (imho).

pnorman:
I forgot our library depends on Mapnik. I agree, that rules it out.

[log4cxx](http://logging.apache.org/log4cxx/) is another library mentioned, patterned after log4j. log4cxx is packaged on Ubuntu 12.04, 14.04 and FreeBSD

[google-glog](https://code.google.com/p/google-glog/) is Google's. It is packaged on Ubuntu 14.04 and FreeBSD, but not Ubuntu 12.04.

alex85k:
Maybe use something lighweight and header-only like https://github.com/gabime/spdlog , if you are going to use cppformat anyway https://github.com/openstreetmap/osm2pgsql/issues/398
It would be great to replace boost completely, to my opinion. It is the biggest dependency, after all and used only in few places.

alex85k:
Another way is to use boost whenever possible, like https://github.com/alacarte-maps/alacarte and https://github.com/Project-OSRM/osrm-backend (including logging, formatting, testing etc.)

pnorman:
osmium::util::VerboseOutput might be of use here

pnorman:
Per previous discussion, we need support for three levels of output - details (off by default), standard, and errors.

We might also have additional debugging output with some compile-time flags (e.g. all COPY lines to postgres)

PR #483, opened 2015-10-29
fix printf formats for 64 bit values

alex85k:
Tried to fix https://github.com/openstreetmap/osm2pgsql/issues/468 . Is this correct?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ matrix:
       env: RUNTEST="-L NoDB" CXXFLAGS="-pedantic -Werror"
     - os: osx
       compiler: clang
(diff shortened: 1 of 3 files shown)


alex85k:
Incorrect. Fixing :)

alex85k:
Strange. Travis does not build it at all...

alex85k:
Seems to pass now.

pnorman:
boost::format has bad performance (#398) but all these fprintfs are only run once. I'd be okay with changing these to boost::format, which I believe would avoid the formatting string issues completely.

We still need to find a logging library (#202), but just printing to stderr like before is good enough for now.

pnorman:
Merging, and I'll look at doing a backport for 0.88.x. Pulling out fprintf can wait.

PR #484, opened 2015-10-29
explicit type conversions to reduce MSVC warnings

alex85k:
Here is the cosmetic fix that reduces the number of MSVC warnings
https://github.com/openstreetmap/osm2pgsql/issues/470
(all warnings in hpps)

### DIFF
--- a/pgsql.hpp
+++ b/pgsql.hpp
@@ -25,10 +25,10 @@ void escape(const std::string &src, std::string& dst);
 
 
 inline void pgsql_CopyData(const char *context, PGconn *sql_conn, const char *sql) {
-    pgsql_CopyData(context, sql_conn, sql, strlen(sql));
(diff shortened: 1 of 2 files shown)


pgsql.hpp
```diff
@@ -25,10 +25,10 @@ void escape(const std::string &src, std::string& dst);
 
 
 inline void pgsql_CopyData(const char *context, PGconn *sql_conn, const char *sql) {
-    pgsql_CopyData(context, sql_conn, sql, strlen(sql));
+    pgsql_CopyData(context, sql_conn, sql, (int) strlen(sql));
```
alex85k: We are not going to have 2Gb SQL, I hope.

pnorman: the length goes to `PQputCopyData(PGconn *conn, const char *buffer, int nbytes);` so we're fine

PR #485, opened 2015-10-29
Fix more MSVC warnings related to size_t

alex85k:
Here are the additions to https://github.com/openstreetmap/osm2pgsql/pull/484
I tried to patch only the most obvious parts anfd touch only local variables. 
https://github.com/openstreetmap/osm2pgsql/issues/470

Tried not to touch files from other PRs.

### DIFF
--- a/parse-o5m.cpp
+++ b/parse-o5m.cpp
@@ -576,7 +576,7 @@ static void str_read(byte** pp,char** s1p,char** s2p) {
        return:
        *s1p,*s2p: pointers to the strings which have been read; */
   char* p;
-  int len1,len2;
+  size_t len1,len2;
   int ref;
 
   p= (char*)*pp;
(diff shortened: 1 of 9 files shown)


parse-o5m.cpp
```diff
@@ -576,7 +576,7 @@ static void str_read(byte** pp,char** s1p,char** s2p) {
        return:
        *s1p,*s2p: pointers to the strings which have been read; */
   char* p;
-  int len1,len2;
+  size_t len1,len2;
```
pnorman: This looks okay, lengths are supposed to be <256

alex85k: OK, will replace with (int) conversion.

pnorman: I was saying the change you did looked okay.

PR #487, opened 2015-10-30
Fix three format warnings.

landryb:
node-ram-cache.cpp:372:162: warning: format '%ld' expects argument of type 'long int', but argument 3 has type 'int64_t {aka long long int}' [-Wformat=]
node-ram-cache.cpp:372:162: warning: format '%zd' expects argument of type 'signed size_t', but argument 5 has type 'long long unsigned int' [-Wformat=]
node-ram-cache.cpp:380:52: warning: format '%li' expects argument of type 'long int', but argument 7 has type 'int64_t {aka long long int}' [-Wformat=]

Found while porting/testing on OpenBSD by patrick keshishian


pnorman:
Appears to duplicate #483, but this version is failing travis.

alex85k:
I have changed it 3 times before Travis stopped complaining :)

landryb:
Your solution also works for me, building git HEAD on OpenBSD doesnt produce warnings anymore on node-ram-cache.cpp

PR #488, opened 2015-10-30
Include config.h for HAVE_TERMIOS_H define

landryb:


alex85k:
Could you check also #ifdef HAVE_LIBGEN_H in options.cpp and HAVE_FORK in node-persistent-cache-reader.hpp? (maybe it is better to include confih.h directly)

Issue #490, opened 2015-11-01
Regression tests missing default.style


pnorman:
```
test_regression_simple
NOTICE:  database "osm2pgsql-test-25469-1443121178" does not exist, skipping
Couldn't open style file 'default.style': No such file or directory
FAIL
```

Perhaps related to #478

cc @alex85k

alex85k:
Checking...

alex85k:
Seem to pass on Windows from ctest. 

Did you run it manually? Seems it can work only from osm2pgsql folder now (and will not find executable file).

alex85k:
I made it pass manually only by running
```
python tests/regression-test.py -f tests/liechtenstein-2013-08-03.osm.pbf -x G:/winbuilds/osm2pgsql/build/Debug/osm2pgsql.exe
```
from th main osm2pgsql folder.

pnorman:
You're looking at the python test, I was looking at a C++ one.

It failed with ctest and running it from the command line.

alex85k:
Sorry for misunderstanding. I can reproduce only when calling from build folder without ctest:
```
 ./tests/test-output-pgsql-tablespace 
```
ctest explicitly says to use working directory:
```
 add_test(NAME ${test_name} COMMAND ${test_name}
    WORKING_DIRECTORY ${osm2pgsql_SOURCE_DIR})
 ``` 

However, I am not sure if osm2pgsql_SOURCE_DIR is set correctly on old cmake. Which version do you use?

Maybe replacing  ${osm2pgsql_SOURCE_DIR} with  ${PROJECT_SOURCE_DIR} or absolute path can help (then the changes will be needed).

**Comment by pnorman** [comment]
> Sorry for misunderstanding. I can reproduce only when calling from build folder without ctest:
> ```
 ./tests/test-output-pgsql-tablespace 
```

It's important to be able to do this, otherwise it's a huge pain to attach a debugger to the process.

This was also with `tests/test-output-pgsql`, but the problem probably applies to any tests that use default.style

Perhaps as part of the build process we could copy default.style to the build directory?

> However, I am not sure if osm2pgsql_SOURCE_DIR is set correctly on old cmake. Which version do you use?

My testing VM has 2.8.12.2.

Doing a full `git clean -xfd && git reset --hard HEAD` fixed this particular test failing in ctest, but we still need to be able to run a single test in isolation.

alex85k:
Running tests also open many files from tests folder. For now single test can be run from main folder as
[...]

PR #496, opened 2015-11-05
change directory for tests if TEST_DATA_DIR is present

alex85k:
This patch solves the problem of missing files in tests.https://github.com/openstreetmap/osm2pgsql/issues/490

To run tests from any directory one need to add environment variable

```
export TEST_DATA_DIR=/path/to/osm2pgsql_sources
```

(if the variable is not present, the tests will get data from current directory)

ctest did not need it, it had WORKING_DIRECTORY in it.
If we want to just run executable without setting TEST_DATA_DIR, only auto-hardcoding in source can help. Will do it if really needed.
Python test also needs similar patching. Will do if solution is accepted.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,7 +1,7 @@
 add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} ${ARGS})
 add_dependencies(check osm2pgsql)
 
-add_library(common-pg STATIC common-pg.cpp common-pg.hpp common-cleanup.cpp common-cleanup.hpp)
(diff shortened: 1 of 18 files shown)


lonvia:
I don't think changing the working directory is a good idea in general because it might have unintended side effects, for example, when temporary files are created. So I would prefer to hand in the test directory during compile time and use absolute paths for all files inside the tests.

The other thing this PR (and also #497) shows is that the tests urgently need a test framework. This may be as little as a home-grown base test class all the tests inherit from or it may be an external test library. The gist is that a change like this should certainly not have to touch every single source file in tests/.

lonvia:
Closing this PR as it has gone stale. The topic remains open though, see #490.

Issue #473, opened 2015-10-26
.o5m really slow compared to .osm.pbf


leblowl:
So with o5m I only get node processing speeds of around 50k/s. Converting the same dataset to osm.pbf with osmconvert results in speeds of around 1300k/s for nodes. I also get this funky ```osm2pgsql Warning: unknown .o5m dataset id: 0xdb``` warning but it never seems to effect the overall import result. I have tried importing various filtered subsets of planet_osm, filtered using osmfilter. Is this normal behavior? I was expected o5m to be just as fast. Please let me know if ya'll need any more info, thank you!

leblowl:
Using a command like this: ```osm2pgsql -d gis -S osm2pgsql.style ../../data/streets.o5m --slim --cache 36000 --flat-nodes node.cache --hstore --number-processes 16```

pnorman:
What version? Also, how many hardware threads does your hardware have, and why 36000 cache? The latter two are *unlikely* to be the cause of any problems, but that much cache is unnecessary and not what --help recommends.

leblowl:
Version ```0.89.0-dev (64bit id space)```

CPU info:
```
$ lscpu
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                24
On-line CPU(s) list:   0-23
Thread(s) per core:    2
Core(s) per socket:    6
Socket(s):             2
NUMA node(s):          2
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 45
Model name:            Intel(R) Xeon(R) CPU E5-2620 0 @ 2.00GHz
Stepping:              7
CPU MHz:               2148.203
CPU max MHz:           2500.0000
CPU min MHz:           1200.0000
BogoMIPS:              3991.36
Virtualization:        VT-x
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
L3 cache:              15360K
NUMA node0 CPU(s):     0-5,12-17
NUMA node1 CPU(s):     6-11,18-23
```

I was unsure if extra cache made a difference as I saw quite a few other examples around the web with larger cache sizes

leblowl:
I just pulled the latest changes & rebuilt. With osm.pbf of a streets extract I am now getting speeds up to 3500 k/s on nodes so far. With o5m, it still sits around 50 k/s...

pnorman:
[...]

Issue #244, opened 2014-12-30
o5m import throws an exception


TobWen:
When using the o5m reader inside of osm2pgsql SVN version 0.87.1 (64bit id space) on Debian 7 Wheezy, this warning gets shown immediately:

`osm2pgsql Warning: unknown .o5m dataset id: 0xdb`

Seems like the o5m format got updated, but the reader wasn't.

pnorman:
I doubt this issue was ever as described, but the message is gone with #497

PR #497, opened 2015-11-06
Drop custom o5m parser in favour of libosmium parser

lonvia:
This commit includes some larger structural changes:

- parse_t and parse_delegate_t removed, osm2pgsql now directly
  creates a parse_osmium_t object
- bbox class replaced with osmium::Box
- code duplication in parse_stats_t cleaned up
- add generic parse() function for tests, also to avoid code duplication


pnorman:
Fixes #244, might fix #473 

o5m node parsing speed increases from 1915k/s to 3878k/s. PBF speed remains about the same.

Issue #464, opened 2015-10-18
Multithreading on Windows


alex85k:
Currently when running a Windows build tets, I get many warnings
```
1: WARNING: osm2pgsql was compiled without fork, only using one process!
```
But I see that there is no actual fork() in code, only in broken node-persistent-cache-reader.cpp
There are also unused fork-related defines here: 
https://github.com/openstreetmap/osm2pgsql/blob/master/middle-pgsql.cpp#L87

Maybe it is time to remove #ifdef HAVE_FORK from options.hpp and allow Windows users to use multithreading too?
 ( Or maybe it does strictly require Pthread etc.? )

lonvia:
Fixed in #498.

PR #498, opened 2015-11-08
Replace pthread and boost::thread with std::async

### LINKED ISSUE #466: Maybe we can use std::thread only?
Since osm2pgsql is now fully C++11, maybe it is a time to switch from boost::thread and pthread fragments to std::thread from standard library (like libosmium)? @pnorman  @lonvia 

Its functionality should be enough (I have seen nothing complex). The advantages
  
   - Can get rid of boost::thread dependency (+boost_date_time and boost_chrono on some systems) and related bugs like https://github.com/openstreetmap/osm2pgsql/pull/460#issuecomment-149125399 on clang
   - Make multithreading work on Windows https://github.com/openstreetmap/osm2pgsql/issues/464 :) (pthread should be enough?)
   - There are only 3 files to adapt and 2 of them are just "create n threads -> join":
https://github.com/openstreetmap/osm2pgsql/blob/eece9ac7fd73bf0a922ff98c140953e957de6670/output-pgsql.cpp#L395
https://github.com/openstreetmap/osm2pgsql/blob/eece9ac7fd73bf0a922ff98c140953e957de6670/middle-pgsql.cpp#L1191
https://github.com/openstreetmap/osm2pgsql/blob/eece9ac7fd73bf0a922ff98c140953e957de6670/osmdata.cpp#L244

lonvia:
Fixes #466 and #464.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,7 +116,7 @@ if (NOT Boost_ADDITIONAL_VERSIONS)
 endif()
 
 # first try to find the version
-find_package(Boost 1.50 REQUIRED COMPONENTS system filesystem thread ${BOOST_EXTRA})
+find_package(Boost 1.50 REQUIRED COMPONENTS system filesystem ${BOOST_EXTRA})
 include_directories(${Boost_INCLUDE_DIR})
 
 find_package(PostgreSQL REQUIRED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,7 +164,6 @@ message("Active compiler flags:" ${CMAKE_CXX_FLAGS})
 # Build the library and executable file
 #############################################################
 
-set(HAVE_PTHREAD "${CMAKE_USE_PTHREADS_INIT}")
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/cmake/config.h.in ${CMAKE_CURRENT_BINARY_DIR}/config.h)
 
 if (NOT HAVE_UNISTD_H AND NOT EXISTS ${CMAKE_CURRENT_BINARY_DIR}/unistd.h)
--- a/osmdata.cpp
+++ b/osmdata.cpp
@@ -1,9 +1,10 @@
+#include <cstdio>
+#include <functional>
+#include <future>
+#include <mutex>
 #include <stdexcept>
 #include <utility>
-
-#include <cstdio>
-
-#include <boost/thread.hpp>
+#include <vector>
 
 #include "middle.hpp"
 #include "node-ram-cache.hpp"
@@ -168,7 +169,7 @@ struct pending_threaded_processor : public middle_t::pending_processor {
     typedef std::vector<std::shared_ptr<output_t>> output_vec_t;
     typedef std::pair<std::shared_ptr<const middle_query_t>, output_vec_t> clone_t;
 
-    static void do_jobs(output_vec_t const& outputs, pending_queue_t& queue, size_t& ids_done, boost::mutex& mutex, int append, bool ways) {
+    static void do_jobs(output_vec_t const& outputs, pending_queue_t& queue, size_t& ids_done, std::mutex& mutex, int append, bool ways) {
         while (true) {
             //get the job off the queue synchronously
             pending_job_t job;
@@ -239,14 +240,29 @@ struct pending_threaded_processor : public middle_t::pending_processor {
 
 
         //make the threads and start them
+        std::vector<std::future<void>> workers;
         for (size_t i = 0; i < clones.size(); ++i) {
-            workers.create_thread(boost::bind(do_jobs, boost::cref(clones[i].second), boost::ref(queue), boost::ref(ids_done), boost::ref(mutex), append, true));
+            workers.push_back(std::async(std::launch::async,
+                                         do_jobs, std::cref(clones[i].second),
+                                         std::ref(queue), std::ref(ids_done),
+                                         std::ref(mutex), append, true));
         }
 
         //TODO: print out partial progress
 
-        //wait for them to really be done
-        workers.join_all();
+        for (auto& w: workers) {
+            try {
+                w.get();
+            } catch (...) {
+                // drain the queue, so that the other workers finish
+                mutex.lock();
+                while (!queue.empty()) {
+                    queue.pop();
+                }
+                mutex.unlock();
+                throw;
+            }
+        }
 
         time_t finish = time(nullptr);
         fprintf(stderr, "\rFinished processing %zu ways in %i sec\n\n", ids_queued, (int)(finish - start));
@@ -286,14 +302,27 @@ struct pending_threaded_processor : public middle_t::pending_processor {
         time_t start = time(nullptr);
 
         //make the threads and start them
+        std::vector<std::future<void>> workers;
         for (size_t i = 0; i < clones.size(); ++i) {
-            workers.create_thread(boost::bind(do_jobs, boost::cref(clones[i].second), boost::ref(queue), boost::ref(ids_done), boost::ref(mutex), append, false));
+            workers.push_back(std::async(std::launch::async,
+                                         do_jobs, std::cref(clones[i].second),
+                                         std::ref(queue), std::ref(ids_done),
+                                         std::ref(mutex), append, false));
         }
 
-        //TODO: print out partial progress
-
-        //wait for them to really be done
-        workers.join_all();
+        for (auto& w: workers) {
+            try {
+                w.get();
+            } catch (...) {
+                // drain the queue, so the other worker finish immediately
+                mutex.lock();
+                while (!queue.empty()) {
+                    queue.pop();
+                }
+                mutex.unlock();
+                throw;
+            }
+        }
 
         time_t finish = time(nullptr);
         fprintf(stderr, "\rFinished processing %zu relations in %i sec\n\n", ids_queued, (int)(finish - start));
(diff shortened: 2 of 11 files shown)


CMakeLists.txt
```diff
@@ -116,7 +116,7 @@ if (NOT Boost_ADDITIONAL_VERSIONS)
 endif()
 
 # first try to find the version
-find_package(Boost 1.50 REQUIRED COMPONENTS system filesystem thread ${BOOST_EXTRA})
+find_package(Boost 1.50 REQUIRED COMPONENTS system filesystem ${BOOST_EXTRA})
```
pnorman: If we don't need boost thread then `libboost-thread-dev` should be dropped Ubuntu/Debian and `libboost-thread1.55-dev` from travis.yml

osmdata.cpp
```diff
@@ -1,9 +1,10 @@
+#include <cstdio>
```
pnorman: By the order we've been using, this should be where it was

lonvia:
@alex85k Could you run this on Windows to make sure it actually works as expected?

pnorman:
node-persistent-cache-reader.cpp is now the only user of the HAVE_FORK define

osmdata.cpp
lonvia: I would consider <cstdio> a C++ header. The C header is called <stdio.h>. There is a subtle difference between two, for example, the former exports namespaces.

alex85k:
@lonvia  It's great, thatnk you!
Will run the tests with debug version now and then try some benchmarks.

alex85k:
The usual tests seem to pass. (actually, there are same 3 failing muti-backend tests with master branch, but it is for another ticket).

However, the Python regression test (which is multithreaded, I guess), breaks each osm2pgsql with some "heap memory corruption" error deep inside geos call. Will investigate more and put the details here.

CMakeLists.txt
lonvia: Done.

alex85k:
I am afraid Python regression tests still do not work. There are access violation errors in async call:
image

alex85k:
Will post additional info if I find the exact variables and objects that are broken.

alex85k:
It fails if I run with 4 threads only
```
osm2pgsql.exe RU-TY.osm.bz2 --number-processes 4
```
(1 or 2 processes => no errors).
And, as classical threading error, it differs for each run:
Run 1:
image
Run 2:
image
Run 3:
image
The last was "vector iterator not dereferencable" exception in finding tag:
big screenshot is here https://db.tt/O5DcclDt .

lonvia:
These errors seem to happen with stack-allocated data. So my uneducated guess would be: stack overflow. Is there a way to increase the stacksize for std::threads in MSVC, just to check if it runs through with more stack.

alex85k:
Tried increasing stack size from 30Mb to 300Mb (/STACK linker option) - memory consumption is twice higher, but the errors are the same - always inside or after get_wkt_split . Maybe my geos 3.5.0 from git is buggy... 

Debbuging showed that the node and tag lists are small at the breaking point (5-8 node ways with ~4 tags) and coordinate sequences seem to be OK. I can not see any common data used. All the workers seem to be still running (tried to add  pritnf and breakpoint after w.get() in process_ways()), so there are no destruction-related errors too.

lonvia:
Thanks, so we can exclude stack problems. It might be possible that your geos library is somehow not thread-safe (either because the C++ interface of geos doesn't use the thread-safe functions of geos or because it was built wrongly). I'm not sure how the tag list corruption would fit, although it is possible that geos corrupts other memory if it happens to use a thread-unsafe version of malloc.

alex85k:
Checked the cmake scripts and dll dependencies - all seems to be correct, /MD or /MDd flags are used in geos and osm2pgsql  (multi-threaded runtime with thread-safe malloc). Tried older Geos 3.4.2 from git and also geos from https://github.com/mapbox/windows-builds (built without cmake, release build).

C++ geos seems to be not thread-safe, only the C functions have _r variants. Are there any converter objects that we are using together from multiple threads?

I thought it may be cross-thread stack access problems https://software.intel.com/en-us/node/595554 but the stack variables we pass by references seem to be created in same thread we use it: https://github.com/openstreetmap/osm2pgsql/blob/70b27884eabbdf43ccaf71d4a92d1daa5e30556f/output-pgsql.cpp#L309  

Will try to run at work under Intel Inspector. It should tell if there are any thread-related problems left.

alex85k:
Intel tools finds nothing particular (few memory leaks in middle-ram, new ... as a parameters). 
The program does not crash or show critical error while running with analysis tools (it is slowing the threads down very much), so there are no new ideas... 

The only find: the errors does not appear if I do not call get_wkt_split or if I insert ``return nullptr`` before 
```c++
wkts->back().geom = writer.write(geom.get());
```
and after get_wkt_split call. But the errors resume (with some pif I move ``return nullptr`` after that line. It usually happens in geom.isValid() call  or inside writer.write.

lonvia:
How about definining a global mutex in geometry_builder.cpp and adding a lock_guard around get_wkt_split (and potentially other functions that use writer).

alex85k:
Thank you, mutex in all writers helped :) Running all tests & benchmarking Release to check...

Was it really using some common objects? Was the mutex really necessary or it is just a workaround for now?

lonvia:
Jay, that's good news. Let me dig into the geos source code a bit and see if there isn't something to make the calls thread-safe. It's just strange that this problem never appeared on Linux, but I guess that's just the nature of multi-threading bugs.

alex85k:
Tested: some speedup is possible even with mutex. 625s -> 425s (pending ways 98s -> 56s). Used 4 threads on moble Intel i7 Haswell 4200u (2 real cores).

alex85k:
By the way, one lock_guard in  get_wkt_split  seemed to be enough...
https://github.com/alex85k/osm2pgsql/commits/replace-pthread-with-stdthread

alex85k:
The tests passes with only one lock_guard. Strange result is that 1 processes is faster that 2 or 4 in that case (this time bulgaria-latest.osm.pbf was used). Time of processing pending ways increases from 34 to 43 s (1->2 processes).

lonvia:
Those are odd numbers. With only one thread I would expect numbers to be approximately equal, no matter how many lock_guards there are.

I've added a fix for some variable shadowing on this branch. I doubt it makes a difference but could you give it a try anyways? I have the nagging feeling that somewhere in the code a freed pointer is accessed. Will stare at the code some more tomorrow.

alex85k:
The lock_guard tests were not good because of different data formats used.

Tried with latest commit https://github.com/lonvia/osm2pgsql/commit/68e328d9962aa97212d01973e846e2a454088cb4 - the errors are the same for now :)

alex85k:
By the way, one possible threading isuue:
image
std::setlocale is set and restored from multiple threads.

alex85k:
And when I deleted that locale switching code from geos (adding std::setlocale(LC_NUMERIC,"C") into main() ), the bug was gone. However, that is not a working solution...

alex85k:
https://trac.osgeo.org/geos/ticket/260

It would be good to use stringstream.imbue for setting locale in geos instead of CLocalizer but I do not know how to make they accept it.

lonvia:
Good catch. Sadly enough the geos code is so bad that there is no way to get around the setlocale() call. I wonder if we shouldn't simply just switch to WKB. It's CLocalizer free and also produces shorter strings.

pnorman:
[a comment in output-pgsql.cpp](https://github.com/openstreetmap/osm2pgsql/blob/master/output-pgsql.cpp#L59-L70) discusses WKB, but for table_t::write_node a WKT is constructed by hand with boost::format so the comment might just be wrong. Something similar is done for ways.

I do recall issues with locale stuff not being thread-safe in Geos coming up previously, but can't find a reference. If this the issue, then it might be reproducible on linux by changing the locale to non-C before starting osm2pgsql.

There's a difference between WKB as text, EWKB as text, WKB, and EWKB, and I'm not sure which we'd want to send with either COPY or INSERT

lonvia:
Postgres requires hex-encoded EWKB (with srid) iirc. Looks like geos is now able to produce that natively, see https://github.com/libgeos/libgeos/commit/cd9a22e7d6c8b7eab0813a9862c6a465bccce7f9

pnorman:
But needs 3.5.0 for that, which is a problem.

lonvia:
Available since 3.0.0 (i.e. since 2007) if I read that commit right.

pnorman:
Both master (0e3d5e2) and 68e328d9962aa97212d01973e846e2a454088cb4 take the same time for a full planet import within the margin of error. Command line was `osm2pgsql -d gis --cache 28000 --slim --drop --flat-nodes nodes.bin --number-processes 8 ~/planet-150202.osm.pbf`

alex85k:
I have found an "official workaround" of locale problem by Microsoft:
https://msdn.microsoft.com/en-us/library/ms235302.aspx

Here is the patch: https://github.com/lonvia/osm2pgsql/pull/3
If included, the bug is gone even without switching to WKB (however WKB may be cleaner and faster).

lonvia:
Reading through the article you provided, wouldn't an even better fix be to run `_configthreadlocale(_ENABLE_PER_THREAD_LOCALE)` when starting each worker thread on Windows? That should stop setlocale from interfering between threads.

alex85k:
I tried to add  exactly this into https://github.com/lonvia/osm2pgsql/pull/3 . Maybe the place of insertion was not exact? Or there any more parallel WKT-related calls?

lonvia:
Sorry, my bad. I only read as far as `setlocale("C")`. Looks good.

alex85k:
Shortened the workaround patch as discussed in comments.

lonvia:
I've merged _configthreadlocal change and AppVeyor is still happy. From my side this PR is good to merge now.

Huge thanks to @alex85k for patiently debugging the problem and finding the nice 1-line solution.

PR #506, opened 2015-11-18
Add area reprojection facility

lonvia:
Supersedes #454 which is no longer mergable because of cmake and geometry-builder refactoring.

@woodpeck could you check that it still works for you? I've done some minor cleanup especially wrt resource management but functionally it should be the same.


woodpeck:
works for me! thanks for adopting this @łonvia, guess it would otherwise have been rotting away ;)

PR #511, opened 2015-11-27
Convert geometries to WKB instead of WKT

lonvia:
Changes geometry-builder to emit WKBs. Nodes are still built as WKTs at the moment.
Improves performance by 15-20%.

Planet import with `./osm2pgsql -d gis -S ../default.style --flat-nodes osm2pgsql.flatnodes -s -C 24000 --drop planet-150624.osm.pbf` on 32GB RAM all-SSD system with 6x2 CPUs yields:

master:
```
Node stats: total(2926993226), max(3610010794) in 1064s
Way stats: total(295111882), max(355378385) in 12100s
Relation stats: total(3493735), max(5307354) in 15387s
Pending ways took 9523s at a rate of 20389.88/s
Osm2pgsql took 55724s overall
```

use-wkb branch:
```
Node stats: total(2926993226), max(3610010794) in 1065s
Way stats: total(295111882), max(355378385) in 9093s
Relation stats: total(3493735), max(5307354) in 11710s
Pending ways took 8320s at a rate of 23338.07/s
Osm2pgsql took 48452s overall
```

### DIFF
--- a/geometry-builder.cpp
+++ b/geometry-builder.cpp
@@ -24,6 +24,7 @@
 #include <algorithm>
 #include <cmath>
 #include <cstddef>
+#include <sstream>
 #include <stdexcept>
 #include <memory>
 #include <new>
@@ -47,7 +48,7 @@
 #include <geos/geom/Polygon.h>
 #include <geos/geom/MultiPolygon.h>
 #include <geos/io/WKTReader.h>
-#include <geos/io/WKTWriter.h>
+#include <geos/io/WKBWriter.h>
 #include <geos/util/GEOSException.h>
 #include <geos/opLinemerge.h>
 using namespace geos::geom;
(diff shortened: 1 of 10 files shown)


geometry-builder.cpp
```diff
@@ -224,12 +227,12 @@ geometry_builder::maybe_wkt_t geometry_builder::get_wkt_simple(const nodelist_t
         auto coords = nodes2coords(gf, nodes);
         if (polygon && is_polygon_line(coords.get())) {
             auto geom = create_simple_poly(gf, std::move(coords));
-            wkt = std::make_shared<wkt_t>(geom.get(), projection);
+            wkt = std::make_shared<wkt_t>(geom.get(), true, projection);
```
pnorman: should this be changed to `wkb = ... <wkb_t> ...`?

pnorman:
Results look okay, but there's a lot of variables with wkt in the name that now seem to refer to wkbs

lonvia:
If rename we must, I'd prefer a more neutral name. Currently wkt_t can contain both WKT (for nodes) and WKB (for everything else). I'll think about it.

lonvia:
Looking through the code, the `maybe_wkt(s)_t` types need a major rewrite anyway. There is no need that they are shared_ptr. They can be simple values. They functions in geometry_builder are already written in a way that RVO should be possible. I would suggest to do all this in a separate PR and do the renaming at that point.

PR #512, opened 2015-11-29
Fix build without FIXED_POINT

dperpeet:
* fixed an error in `node-ram-cache.hpp`
* make the headers work with msvc on windows

These changes have only been tested on Windows x64 with Visual Studio 2015.

### DIFF
--- a/node-ram-cache.cpp
+++ b/node-ram-cache.cpp
@@ -62,7 +62,9 @@
 
 #define SAFETY_MARGIN 1024*PER_BLOCK*sizeof(ramNode)
 
+#ifdef FIXED_POINT
 int ramNode::scale;
+#endif
 
 static int32_t id2block(osmid_t id)
 {
@@ -305,8 +307,9 @@ node_ram_cache::node_ram_cache( int strategy, int cacheSizeMB, int fixpointscale
       maxSparseTuples(0), sizeSparseTuples(0), maxSparseId(0), cacheUsed(0),
       cacheSize(0), storedNodes(0), totalNodes(0), nodesCacheHits(0),
       nodesCacheLookups(0), warn_node_order(0) {
-
+#ifdef FIXED_POINT
     ramNode::scale = fixpointscale;
+#endif
     blockCache = 0;
     blockCachePos = 0;
     cacheUsed = 0;
--- a/node-ram-cache.hpp
+++ b/node-ram-cache.hpp
@@ -68,9 +68,9 @@ private:
 #else
 public:
     ramNode() : _lat(NAN), _lon(NAN) {}
-    ramNode(double _lon, double _lat) : _lon(lon), _lat(lat) {}
+    ramNode(double lon, double lat) : _lon(lon), _lat(lat) {}
 
-    bool is_valid() const ( return !std::isnan(_lon); }
+    bool is_valid() const { return !std::isnan(_lon); }
     double lon() const { return _lon; }
     double lat() const { return _lat; }
 private:


martin-s:
I would rather change node-ram-cache.hpp to:
```
ramNode(double lon, double lat) : _lon(lon), _lat(lat) {}
```

dperpeet:
> I would rather change node-ram-cache.hpp to:
> 
> ramNode(double lon, double lat) : _lon(lon), _lat(lat) {}

done

parse-osmium.cpp
```diff
@@ -21,7 +21,10 @@
 */
 
 #include <boost/format.hpp>
-
+#ifdef _WIN32
+#    define NOMINMAX
+#    include <WinSock2.h>
```
lonvia: Could you explain what error you get here that requires the include? This looks like something that should go into libosmium.

lonvia:
The changes are ok but I wonder: did you intentionally try to compile without -DFIXED_POINT or is something going wrong with the cmake configuration for you?

node-persistent-cache.cpp
```diff
@@ -15,7 +15,9 @@
 #include <string.h>
 #include <sys/stat.h>
 #include <sys/types.h>
-#include <unistd.h>
+#ifndef _WIN32
+#    include <unistd.h>
+#endif
```
alex85k: This should not be needed. There are many other unistd includes, so cmake just generates empty unistd.h.

sprompt.cpp
```diff
@@ -57,7 +57,9 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
-#include <unistd.h>
+#ifndef _WIN32
+#    include <unistd.h>
+#endif
```
alex85k: Should work without it (cmake should generate it). Many other places with unistd.h.

parse-osmium.cpp
alex85k: We already have -DNOMINMAX in cmake. 
Why WinSock.h?

alex85k:
Adding -DFIXED_POINT is really giving errors, so changes in node-ram-cache.hpp are really needed.

pnorman:
> Adding -DFIXED_POINT is giving errors, so changes in node-ram-cache.hpp are really needed.

We should check this option in CI.

Or switch to storing unprojected coordinates in slim tables, which would eliminate the option and fix a few bugs.

lonvia:
We surely should fix the syntax errors but further than that, I'm more for tackling the projection issue and getting rid of the non-fixed-point options. That is pretty high on my todo list anyways.

From @alex85k comments I take it that the other issues really are caused by some cmake misconfiguration and should be ignored.

parse-osmium.cpp
dperpeet: This was my mistake. A consistent, solution wide use of `WIN32_LEAN_AND_MEAN` takes care of this. The extra include isn't necessary.

dperpeet:
Thanks for the feedback!
@lonvia 
> The changes are ok but I wonder: did you intentionally try to compile without -DFIXED_POINT or is something going wrong with the cmake configuration for you?
The truth will probably sound even worse to you: I created my project without cmake, because I used [mapbox/windows-builds](https://github.com/mapbox/windows-builds/) for building mapnik and I wanted to reuse the dependencies as much as possible. I also had an existing solution that I only had to tweak a bit.

I reduce this pull request to the minimum set of fixes.

dperpeet:
addressed issues and updated

PR #516, opened 2015-12-13
avoid double-filtering tags on ways in pgsql output

lonvia:
In way_add() tags have been filtered and then forwarded to pgsql_out_way() where they will filtered again. Moved filtering code out of pgsql_out_way() to avoid filtering twice altogether.

### DIFF
--- a/output-pgsql.cpp
+++ b/output-pgsql.cpp
@@ -83,29 +83,12 @@ E4C1421D5BF24D06053E7DF4940
 212696  Oswald Road     \N      \N      \N      \N      \N      \N      minor   \N      \N      \N      \N      \N      \N      \N    0102000020E610000004000000467D923B6C22D5BFA359D93EE4DF4940B3976DA7AD11D5BF84BBB376DBDF4940997FF44D9A06D5BF4223D8B8FEDF49404D158C4AEA04D
 5BF5BB39597FCDF4940
 */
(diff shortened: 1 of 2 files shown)


output-pgsql.cpp
```diff
@@ -83,29 +83,12 @@ E4C1421D5BF24D06053E7DF4940
 212696  Oswald Road     \N      \N      \N      \N      \N      \N      minor   \N      \N      \N      \N      \N      \N      \N    0102000020E610000004000000467D923B6C22D5BFA359D93EE4DF4940B3976DA7AD11D5BF84BBB376DBDF4940997FF44D9A06D5BF4223D8B8FEDF49404D158C4AEA04D
 5BF5BB39597FCDF4940
 */
-int output_pgsql_t::pgsql_out_way(osmid_t id, const taglist_t &tags, const nodelist_t &nodes, int exists)
+int output_pgsql_t::pgsql_out_way(osmid_t id, taglist_t &outtags,
```
pnorman: We're modifying `outtags` later on, is this an issue?

lonvia: It is not an issue at the two places where the function is currently used.

Th whole area hack is a bit dirty anyway. In the long run I would prefer to move the special-casing of the area tag into table_t::write_wkt() to avoid tag modification and remove all that code duplication. It would also make the area available for output-multi.

PR #527, opened 2016-01-12
rework projection code

### LINKED ISSUE #514: some multipolygons don't get created when using --proj 4326
osm2pgsql is not building some multipolygons when using --proj 4326.

Have a look at this relation:
```wget -O nrw.osm http://www.openstreetmap.org/api/0.6/relation/62761/full```
or its simiplied version (only have the size and faster to download):
 ```wget -O nrw.osm https://clbin.com/nbVgx```

I've imported it like this using the default.style with and without --multi-geometry using osm2pgsql c8265f809e3c809638040db750477ba2963cc0ee:
```
osm2pgsql --create --database testcase --proj 4326 nrw.osm
```

The resulting administrive area can't be found in planet_osm_polygon, it can be found splittet in more than 20 pieces in planet_osm_line.

Import works as expected when using native mercator projection:
```
osm2pgsql --create --database testcase nrw.osm
```

lonvia:
- make use of osmium's proj wrapper CRS
- use virtual functions instead of switching over projection
- remove duplicate code (mainly WSG84 to mercator transformation)
- use custom projections also with --proj 4326 and 900913 (fixes #514)

### DIFF
--- a/geometry-builder.hpp
+++ b/geometry-builder.hpp
@@ -35,7 +35,7 @@ class GeometryFactory;
 class CoordinateSequence;
 }}
 
-struct reprojection;
+class reprojection;
 
 class geometry_builder
 {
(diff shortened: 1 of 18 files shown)


pnorman:
Did some performance testing and it's a wash

PR #532, opened 2016-02-22
Add more tags from imports to list of deleted tags

math1985:


pnorman:
Some or all of these should go upstream to osm2pgsql

math1985:
I'm not sure if I understand what you mean?

pnorman:
whoops - reading out of context on the laptop. nevermind

default.style
```diff
@@ -226,8 +226,96 @@ node,way    naptan:*                text    delete
...
+node,way    naptan:*                text    delete
+
+# Various imports (PL)
+node,way    addr:city:simc          text    delete
+node,way    addr:street:sym_ul      text    delete
+node,way    building:usage:pl       text    delete
+node,way    building:use:pl         text    delete
+
+# National Hydrography Dataset (US)
+node,way    NHD:*                   text    delete
```
pnorman: NHD:* is already present

default.style
```diff
@@ -226,8 +226,96 @@ node,way    naptan:*                text    delete
...
+# DIBAVOD (CZ)
+node,way    dibavod:id              text    delete
+
+# Maa-amet (EE)
+node,way    maaamet:ETAK            text    delete
+
+# FANTOIR (FR)
+node,way    ref:FR:FANTOIR          text    delete
+
+# MassGIS (Massachusetts, US)
```
pnorman: This should be with the other US stuff

math1985:
Sorted by country and alphabetised.

Ready for review again.

pnorman:
I'm not sure about adding some of the ref tags - they're beyond what we've normally had in default.style. That doesn't mean I think it's a bad idea, but I'm not sure, and could use comments from others.

pnorman:
The same changes to the delete list should be applied to empty.style

math1985:
> The same changes to the delete list should be applied to empty.style

Done.

ImreSamu:
my candidate for deleting :  
* `"ref:bag"`     "BAG reference number for properties"  
  * taginfo: https://taginfo.openstreetmap.org/keys/ref%3Abag 
     *  Number of objects:  (now)  **9 966 798**
  * wiki:  https://wiki.openstreetmap.org/wiki/Key%3Aref%3Abag
     *   _"This key contains the reference number (ID) from the BAG database containing addresses and buildings in the Netherlands. ..."_

pnorman:
There's a tradeoff here - most of these tags aren't useful for most rendering, but might be needed for some purposes with osm2pgsql. This is different than tags like `source` which would only be used by very specialized data consumers.

OpenStreetMap Carto has moved to Lua transforms, which removes some of the need for this.

PR #544, opened 2016-02-27
Install node-persistent-cache-reader executable & manpage too.

sebastic:
The executable previously named `nodecachefilereader`.


pnorman:
We've decided its better to remove nodecachefilereader, as it serves no useful purpose anymore, is not maintained, and was broken for over a year before anyone noticed.

sebastic:
Why is the executable still built, just for testing?

sebastic:
Nevermind, I just saw the removal in 479ef507a4d8208a1410b15ae4a052086a147abe

PR #548, opened 2016-02-28
Rework wildcard code

lonvia:
- replace with simpler recursive implementation
- make function case-sensitive as per documentation
- only use wildcard for columns with the delete flag,
  the bahaviour is unpredictable for other tags
- add simple test for wild card function


pnorman:
This is possibly 6% faster on pending ways and 1.7% faster overall. As it's clearly a speed improvement and also has cleaner code, I'm not motivated to get more precise numbers.

PR #551, opened 2016-03-03
Change the section about `--flat-nodes`

StyXman:
The original paragraph mentions that `--flat-nodes` in unsuitable for "small extracts", but a few test on my own show that it's actually unsuitable for small _imports_. For instance, I'm importing a 5x3 degrees bbox from a whole Europe extract (~80x51 degrees bbox) and it's _way_ faster without `--flat-nodes`.


pnorman:
fixed in 3e949ce3ed13c65f52887ed24a405db9de2fae3b

PR #570, opened 2016-04-14
add command line option for EXPIRE_TILES_MAX_BBOX instead of hard coded value of 20000

JesseCrocker:
Default value is still 20000

new lines in --help --verbose output"
```
       -B|--expire-bbox-size Max size for a polygon to expire the whole polygon,
                             not just the boundary.
```

### DIFF
--- a/options.cpp
+++ b/options.cpp
@@ -16,7 +16,7 @@
 
 namespace
 {
-    const char * short_options = "ab:cd:KhlmMp:suvU:WH:P:i:IE:C:S:e:o:O:xkjGz:r:V";
+    const char * short_options = "ab:cd:KhlmMp:suvU:WH:P:i:IE:C:S:e:o:B:O:xkjGz:r:V";
     const struct option long_options[] =
     {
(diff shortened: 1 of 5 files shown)


lonvia:
Looks good to me. @pnorman any objection at adding yet another command line parameter?

pnorman:
I'd like to avoid using up a short letter for something that's going to be infrequently used

JesseCrocker:
I have removed the single letter option.

options.cpp
```diff
@@ -168,6 +169,8 @@ namespace
     Expiry options:\n\
        -e|--expire-tiles [min_zoom-]max_zoom    Create a tile expiry list.\n\
        -o|--expire-output filename  Output filename for expired tiles list.\n\
+       -B|--expire-bbox-size Max size for a polygon to expire the whole polygon,\n\
+                             not just the boundary.\n\
```
lonvia: You forgot to adapt the documentation. While you change that, can you please also adapt the number of options in README.md and docs/usage.md.

PR #572, opened 2016-04-20
fix projection info for generic projection

### LINKED ISSUE #571: Problem with flag -E
Hi, 
If I try to set different projection system with flag -E I receive the following error. I'm running osm2pgsql on Debian testing OS, and I tried the package 0.90.0 and this master repo but I got the same error.
No error is reported without -E flag or us -l 

```
osm2pgsql -s -C 4000 -k -d osm nord-italy.pbf -E EPSG:32632
osm2pgsql SVN version 0.90.0 (64 bit id space)

Osm2pgsql failed due to ERROR: creation of CRS failed: no options found in 'init' file
```

lonvia:
fixes #571

### DIFF
--- a/reprojection.cpp
+++ b/reprojection.cpp
@@ -75,7 +75,8 @@ class generic_reprojection_t : public reprojection
 {
 public:
     generic_reprojection_t(int srs)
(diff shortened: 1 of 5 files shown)


lonvia:
Now even comes with a test.

PR #580, opened 2016-05-14
add flag to process new-style polygons only

lonvia:
With this flag set, the logic that checks for tags of outer
rings on multipolygons is disabled and such ways also cannot
be superseded and will always be rendered. Note that for
efficiency reasons we still process polygon ways in the
second stage where we can make use of multiple threads.

This is an experimental change for exploring the effects of old-style multipolygons in the OSM database. I haven't extensively tested yet if it does what it says.

### DIFF
--- a/options.cpp
+++ b/options.cpp
@@ -64,6 +64,7 @@ namespace
         {"exclude-invalid-polygon",0,0,210},
         {"tag-transform-script",1,0,212},
         {"reproject-area",0,0,213},
+        {"ignore-oldstyle-polygons", 0, 0, 215},
         {0, 0, 0, 0}
     };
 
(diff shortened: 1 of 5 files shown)


pnorman:
> Note that for efficiency reasons we still process polygon ways in the second stage where we can make use of multiple threads

I wonder about this - my past thought has been that even though pending ways is parallel, because it has to hit the middle for each way's nodelist it is slower than handling something in the initial processing, where only node locations need to be fetched.

lonvia:
My quick test against a Germany extract on a 32GB RAM machine with SSDs and 8 threads showed that it takes about twice as long doing all ways in the 1st stage.

lonvia:
With the [progress on old-style MP cleanup](http://area.jochentopf.com/stats/) we should be rid of those MPs within a month. So I would suggest to get this PR cleaned up and make a switch that **enables** old-style MP processing (i.e. make new-style only the default).

PR #585, opened 2016-06-01
Code cleanups

DerDakon:
* removes some unused defines and feature tests
 * replaces C-style casts with C++ ones
 * replace 2 malloc() calls


pnorman:
Rather than going to `new` and bare pointers, would it be much additional effort going to `unique_ptr`?

lonvia:
Indeed, if you can make all those malloc/free pairs disappear in `unique_ptr`, that would be really cool.

Other observation: alloca() with a constant should really just be a variable, e.g. char *buf = alloca(64) -> char buf[64];

DerDakon:
unique_ptr sounds indeed like a good idea. Regarding the alloca: I had the same thing in mind but did not notice that I can indeed move the variable declarations into those scopes. Followup patch coming soon.

PR #592, opened 2016-06-12
Update libosmium to version 2.7.2

lonvia:
I had some test failures because the multipolygon test files contained longitude values in E notation, which is no longer supported by libosmium. Looks like the file was produced by a buggy JOSM. After loading and saving the file in a current version of JOSM, the E notation was gone.


pnorman:
> I had some test failures because the multipolygon test files contained longitude values in E notation, which is no longer supported by libosmium. Looks like the file was produced by a buggy JOSM. After loading and saving the file in a current version of JOSM, the E notation was gone.

Is it just E notation or also e that libosmium doesn't support? If the latter, we should report that upstream as a bug.

The latter is valid OSM XML produced by the rails port, e.g. http://www.openstreetmap.org/api/0.6/node/3286696168/1

```xml
<osm version="0.6" generator="OpenStreetMap server" copyright="OpenStreetMap and contributors" attribution="http://www.openstreetmap.org/copyright" license="http://opendatacommons.org/licenses/odbl/1-0/">
<node id="3286696168" changeset="28121089" timestamp="2015-01-13T21:04:17Z" version="1" visible="true" user="c2r" uid="5926" lat="51.8163125" lon="3.2e-05"/>
</osm>
```

lonvia:
It does not support scientific notation at all. I'm tempted to open a bug against the rails port here.

Holding off with this PR for the moment.

pnorman:
see openstreetmap/openstreetmap-website#341

lonvia:
Will be fixed in libosmium soon. I'll do a new PR against the next version. Closing for now.

Issue #437, opened 2015-08-10
Cache kept in indexing step


StyXman:
I noticed that `osm2pgsql` version 0.88.0 keeps the cache when it has finished processing the ways and relations and starts indexing the tables. Is it needed? Could that memory be freed to the kernel's page cache?

pnorman:
Yes, freeing memory once out of pending would be good, since if the DB is on the same machine as osm2pgsql it would give more memory for index creation.

hholzgra:
I actually spent a bit of time on this earlier today

https://github.com/hholzgra/osm2pgsql/commit/36c397f3e694075d0a74371fb3cc0a21819a3aaf

Maybe I should have checked pending open issues and pull requests earlier ...

hholzgra:
So far I've just added freeing of mid/out clones as part of the process_polygons method in middle-pgsql.cpp, the mid clones hold copies of the cache auto_ptr, so these have to be cleaned up to bring the reference counter down to 1, and eventually 0, in the final middle_pgsql::stop()

PR #612, opened 2016-07-21
Drop slim mode tables just after finishing processing pending ways an…

StyXman:
…d relations, freeing space for clustering and indexing.


lonvia:
It might be a good idea to unconditionally move the mid.stop() up in the code as it also frees the memory from the node cache (see #437). Needs some testing.

StyXman:
This could be tested this as is, and if it works properly, then see if it can be done unconditionally.

lonvia:
I've done a few test runs with the UK excerpt. This works ok with drop but there is a 15% performance penalty for non-drop imports. So I would go with this PR for now. @pnorman?

pnorman:
I ran a test import with this PR, and the commit before it

Import flags were `--cache 34500 --flat-nodes nodes.bin --slim` with a 161212 planet. The machine had 64GB RAM, vm.swappiness=20, and very fast NVMe drives

```
                master          PR
time            15:54:26        15:48:21
processing time 17650s          16641s
pending ways    6803s           6822s
indexing poly   14238s          13028s
indexing ways   32795s          33414s
Max res size    36.90 GB        36.87
```

No big differences, but but Munin is more interesting. Master is to the left, PR to the right.

image

Just for reference, DB size is below on the same time scale
image

I could retry with swap disabled, but there's not much indication of memory differences with this testing.

Because with this much ram the node cache is unused, it probably all went out to swap.

pnorman:
image
PR to the left in this one.

I see no particularly significant differences between the two runs. So I'm fine merging or not, but see no real differences in a full planet import.

pnorman:
It's been pointed out to me that I tested code that works with `--drop` on an import without `--drop`. so the above numbers are not meaningful.

StyXman:
@pnorman: for me the big difference is disk space, but I understand that you *have* to look at performance impact. If anybody runs more tests, a disk usage graph would be nice. I could do it, but I will have to wait until I can buy a new SSD, because imports do not finish for me (because of disk space issues :).

pnorman:
> @pnorman: for me the big difference is disk space, but I understand that you have to look at performance impact. If anybody runs more tests, a disk usage graph would be nice.

I'll do it after the current tests finish.

Somewhat related is @hholzgra's 36c397f3e694075d0a74371fb3cc0a21819a3aaf

pnorman:
> @pnorman: for me the big difference is disk space

I don't see a disk space difference

image

The big DB size spike is at the end of processing, right before the slim tables are dropped.

Left is **current master**, right is this PR.

Testing this PR vs HEAD~ for memory

StyXman:
@pnorman: the database is not the only thing using disk; the flat nodes files also does and is quite big. My patch aims to reclaim that disk space before it starts creating the indexes. This is useful for those who are short of space and put all in a single disk/partition like me. That's why I said disk usage and not db size :)

pnorman:
> the flat nodes files also does and is quite big.

But the flat nodes file isn't deleted.

> My patch aims to reclaim that disk space before it starts creating the indexes

It doesn't seem to be reclaim the slim table space.

For disk usage, this branch is actually **worse**

this PR on the left, HEAD~ on the right.

image

StyXman:
(One year later...)

I was rechecking this code again and I noticed that yes, indeed, the file is not removed, so I have two questions:

* Would you say that, given that the `--drop` option removes the tables needed for update, it could also remove the flat file?
* I don't seem to find the definition of `node_persistent_cache.reset()`, which I think is the right function to implement such a thing.

StyXman:
I found how to. I tested locally with this command:

`osm2pgsql --username mdione --database gis --cache 1024 --number-processes 4 --verbose \
    --slim --flat-nodes /home/mdione/src/projects/osm/nodes.cache --hstore \
    --multi-geometry --style /home/mdione/src/projects/osm/openstreetmap-carto.style --tag-transform-script /home/mdione/src/projects/osm/openstreetmap-carto.lua \
    --drop`

And it works, the file `nodes.cache` is deleted after `Going over pending relations...`, which frees a lot of space for indexing.

StyXman:
I will do a full Europe test import this weekend and see if nothing relevant breaks. I al ready tested with Andorra, Croatia and Argentina extracts from geofabrik.

node-persistent-cache.cpp
```diff
@@ -76,3 +77,22 @@ node_persistent_cache::~node_persistent_cache()
...
+        m_fd = -1;
+    }
+
+    // but as the file is mmap'ed, we also have to release the shared_ptr
+    // which deletes the index_t, which unmmaps
+    m_index.reset();
+
+    if (remove_file) {
+        unlink(fname);
+    }
```
lonvia: You are duplicating destructor code here. If you just move the three last lines to the end of the destructor, you can get rid of this entire function.

node-persistent-cache.hpp
```diff
@@ -32,6 +33,8 @@ class node_persistent_cache
     std::shared_ptr<node_ram_cache> m_ram_cache;
     int m_fd;
     std::unique_ptr<index_t> m_index;
+    bool remove_file;
+    const char *fname;
```
lonvia: Coding style: follow the example and prefix members with m_.

middle-pgsql.cpp
```diff
@@ -1072,7 +1072,10 @@ void middle_pgsql_t::pgsql_stop_one(table_desc *table)
 void middle_pgsql_t::stop(void)
 {
     cache.reset();
-    if (out_options->flat_node_cache_enabled) persistent_cache.reset();
+    if (out_options->flat_node_cache_enabled) {
+        persistent_cache->clean_up();
+        persistent_cache.reset();
+    }
```
lonvia: I still think that the solution should be generalized, so the benefit is there even without --drop. To achieve that the stop() function needs to be split at this point, the upper part executed before the indexing and the lower part needs to remain where the stop() call was before this change.

And if we do that, then the future vector below can even be merged with the one with the indexing functions in osmdata.cpp below, meaning that we get a much better parallelization.

StyXman: From what I understand, there are 5 modes of import:

1. Normal mode, where intermediate data is created in RAM. It's not clear to me from the doc if this mode is updatable or not, but I guess the answer is yes.
2. `--slim` mode, where that data is written in the database. This mode is updatable.
3. `--slim` + `--flat-nodes`, where par of the data if written in a spaces file and `mmap()`'ed, so no db queries. This mode is also updatable.
4. `--slim --drop` and
5. `--slim --flat-nodes --drop`, were the intermediate data is erased. With this flag, the database is not upgradable anymore.

There are two problems with `--slim --drop`:

* The flat nodes file, if used (mode 5), is not erased. This PR fixes this.
* The intermediate data is removed at the end, even when they're not longer used way before that moment (mainly, just before indexing). This PR fixes this only for `--flat-nodes`, mode 5.

So what you mean is that we should do the second item for both modes 5 and 3. I think it makes sense, and it seems that what you mention about splitting `stop()` is actually making this slower to run: I'm getting 500s+ with this patch vs 230s+ without for importing Argentina in my machine.

I'll give the `stop()` splitting a try tomorrow, but you will have to explain a little bit more about the last paragraph. Can I drop by any IRC channel to discuss it? I think I'll be faster than just ping-pong through GitHub.

woodpeck: The "normal mode" (your #1) is not updateable. Updating requires intermediate data to be persistent.

StyXman:
Tested with Argentina. Looks good. @lonvia, is this what you had in mind?

middle-pgsql.cpp
lonvia: In all modes a cache of node locations is kept in memory (the -C parameter). Currently it is kept around in the indexing phase, making the memory unavailable for postgres. But I think you are right, just splitting as I suggested again leaves the drop case in a worse state. This clearly needs some more thinking. So I would suggest to leave it for this PR and tackle the cache dropping later.

StyXman:
ACK. I'll have some time at noon, I'l try to catch you there then.

lonvia:
Merged manually.

I think I'll have a quick look at dropping the node caches myself.

PR #613, opened 2016-07-21
Allow no columns for HSTORE_ALL mode

rmarianski:
When running with `-j` or `HSTORE_ALL`, it can make sense not to have any other columns present.


pnorman:
Looks good and passes tests. I'll backport it too

PR #615, opened 2016-08-05
Add link to readme to find OSM data

tpitale:
As a newbie, I went looking for the OSM data to load. This seems to be canonical place. I could be wrong. I _think_ a link would be helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # osm2pgsql #
 
-osm2pgsql is a tool for loading OpenStreetMap data into a PostgreSQL / PostGIS
+osm2pgsql is a tool for loading [OpenStreetMap data](http://planet.openstreetmap.org/) into a PostgreSQL / PostGIS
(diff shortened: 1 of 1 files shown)


pnorman:
planet.osm.org is the canonical place, but I'm not sure we should link to it. Most people will want to start with a third-party extract.

tpitale:
Could we link to a wiki page or something to describe what third-party extracts are, and maybe some examples? I'm completely ignorant of these other data extracts.

Just trying to help out other newcomers, in the readme or wiki.

lonvia:
planet.osm.org has links to extract providers on its main page. I think it's fair to add a link to it. Personal preference would be to add it in the 'Usage' section (which, by the way, confusingly talks about data.osm.pbf and planet-latest.osm.pbf).

tpitale:
Not sure if the comment surrounding the link is appropriate …

PR #619, opened 2016-08-24
Schema support

asbjornst:
This series adds schema support to the prefix option eg. "gisdata.osm_". Default behavior is
preserved thus placing the tables in in the default public schema.

### DIFF
--- a/middle-pgsql.cpp
+++ b/middle-pgsql.cpp
@@ -1203,7 +1203,7 @@ middle_pgsql_t::middle_pgsql_t()
             /*copy*/ "COPY %p_ways FROM STDIN;\n",
          /*analyze*/ "ANALYZE %p_ways;\n",
             /*stop*/  "COMMIT;\n",
-   /*array_indexes*/ "CREATE INDEX %p_ways_nodes ON %p_ways USING gin (nodes) WITH (FASTUPDATE=OFF) {TABLESPACE %i};\n"
+   /*array_indexes*/ "CREATE INDEX %P_ways_nodes ON %p_ways USING gin (nodes) WITH (FASTUPDATE=OFF) {TABLESPACE %i};\n"
                          ));
     tables.push_back(table_desc(
         /*table = t_rel,*/
@@ -1223,7 +1223,7 @@ middle_pgsql_t::middle_pgsql_t()
             /*copy*/ "COPY %p_rels FROM STDIN;\n",
          /*analyze*/ "ANALYZE %p_rels;\n",
             /*stop*/  "COMMIT;\n",
-   /*array_indexes*/ "CREATE INDEX %p_rels_parts ON %p_rels USING gin (parts) WITH (FASTUPDATE=OFF) {TABLESPACE %i};\n"
+   /*array_indexes*/ "CREATE INDEX %P_rels_parts ON %p_rels USING gin (parts) WITH (FASTUPDATE=OFF) {TABLESPACE %i};\n"
                          ));
 
     // set up the rest of the variables from the tables.
--- a/middle-pgsql.cpp
+++ b/middle-pgsql.cpp
@@ -905,13 +905,14 @@ void middle_pgsql_t::end(void)
  *
  * The input string is mangled as follows:
  * %p replaced by the content of the "prefix" option
+ * %P replaced by the content of the table part of the "prefix" option
  * %i replaced by the content of the "tblsslim_data" option
  * %t replaced by the content of the "tblssslim_index" option
  * %m replaced by "UNLOGGED" if the "unlogged" option is set
  * other occurrences of the "%" char are treated normally.
  * any occurrence of { or } will be ignored (not copied to output string);
  * anything inside {} is only copied if it contained at least one of
- * %p, %i, %t, %m that was not NULL.
+ * %p, %P, %i, %t, %m that was not NULL.
  *
  * So, the input string
  *    Hello{ dear %i}!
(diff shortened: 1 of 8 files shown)


middle-pgsql.cpp
```diff
@@ -1063,6 +1073,7 @@ void middle_pgsql_t::start(const options_t *out_options_)
         pgsql_exec(sql_conn, PGRES_COMMAND_OK, "SET synchronous_commit TO off;");
 
         pgsql_exec(sql_conn, PGRES_COMMAND_OK, "SET client_min_messages = WARNING");
+        pgsql_exec(sql_conn, PGRES_COMMAND_OK, "CREATE SCHEMA IF NOT EXISTS %s", schema.c_str());
```
pnorman: We still support 9.1 which doesn't support the `IF NOT EXISTS` parameter. I recommend dropping this as we already rely on the database having appropriate extensions and this is similar.

pnorman:
cross-ref #173 #279

middle-pgsql.cpp
asbjornst: Ok, I have pushed the branch without it.

pnorman:
I want schema support (see past PRs) but it's surprisingly tricky to do

* This changes the default behavior from using the default schema to using public.

* There are lots of test failures. e.g.

  ```
2/25 Testing: test-hstore-match-only
2/25 Test: test-hstore-match-only
Command: "/home/pnorman/osm/osm2pgsql/build/tests/test-hstore-match-only"
Directory: /home/pnorman/osm/osm2pgsql
"test-hstore-match-only" start time: Aug 24 16:25 PDT
Output:
----------------------------------------------------------
NOTICE:  database "osm2pgsql-test-32244-1472081104" does not exist, skipping
Using built-in tag processing pipeline
Setting up table: ._point
ERROR: CREATE SCHEMA IF NOT EXISTS  failed: ERROR:  syntax error at end of input
LINE 1: CREATE SCHEMA IF NOT EXISTS
                                    ^
```

* We need to carefully test the schema option when there are already osm2pgsql tables in the search path.

**Comment by asbjornst** [comment]
I have changed the code to only behave differently if theres a schema name in the prefix option, so no assumption of default schema being "public".

The "CREATE SCHEMA IF NOT EXISTS" usage was removed in the push I made yesterday, and was not a part of this PR, at the time of your last message.

I have deliberately kept gazetteer output out of this, since it currently doesn't prefix the "place" table. One way to introduce prefix support there, without breaking current usage, could be if prefix starts with "=" then it's also used for the "place" table.

**Comment by pnorman** [comment]
I'm still getting test failures. on `tests/test-output-pgsql-area` I get

```
Expected PGRES_TUPLES_OK but got PGRES_FATAL_ERROR. ERROR:  relation "osm2pgsql_test_polygon" does not exist
LINE 1: SELECT COUNT(*) FROM osm2pgsql_test_polygon
                             ^
 Query was SELECT COUNT(*) FROM osm2pgsql_test_polygon.
FAIL
```

It looks like it created the table as `_polygon`.

To get the tests working locally you *might* need to copy some files from the source dir to the build dir with `cp ../tests/*.{osm,json,style} tests/ && cp ../default.style ./`

Shouldn't there also be some changes to the tests, probably `test-output-pgsql-schema`?

I'll leave it to @lonvia to decide what to do about schemas and Nominatim.

Something that just occurred to me is that the approach of setting schema via a prefix isn't clearest withte multi backend. With that backend the slim tables still have a prefix but the rendering tables don't.

lonvia:
I would prefer to have an additional option to set the schema instead of hacking it into the prefix. This is more verbose and solves the issues with the Nominatim and multibackend.

asbjornst:
I currently see four main mutations of this:
* using prefix option and specifying the schema explicitly in the SQL statements (current code)
* using prefix option and setting the search path if schema is given
* using schema option and specifying the schema explicitly in the SQL statements
* using schema option and setting the search path if schema is given (approach used in #279)

Both option strategies are backwards compatible. 

With regards to Nominatim it doesn't completely support the prefix option at the moment, which makes it hard to introduce it without breaking backwards compatibility. I'm not sure it makes sense to add schema support to the <code>place</code> table, without also add prefix support to it. <code>--prefix</code> currently affects the slim tables in Nominatim mode.

multibackend doesn't currently support the prefix option, so a prefix currently have to be put in the style.

I would prefer <code>--prefix</code> and a possible future <code>--schema</code> to have the same limitations, which was why I implemented it as an extension of <code>--prefix</code>.

The term "schema" can mean either object namespace or the object definitions, due to all databases supporting the latter, and fewer supporting the former, the latter use is more widespread.
Therefore my main objection to <code>--schema</code>, is that by extending the <code>--prefix</code> option, it clearly communicates in which context "schema" should be understood.

By using the schema explicitly in the SQL statements - when specified - it is backwards compatible, and so works as expected even if the default search path has been altered.
The search path approach seams to lead to a few more issues.

Regarding tests, they all pass on my system, but I will write a new test, since the current schema test is good for testing backward compatibility.

@lonvia @pnorman Which mutation should I rewrite this patchset into? Do you have any other things I should take into account before rewriting it?

asbjornst:
@lonvia @pnorman ping

pnorman:
I think an additional option is a requirement.

For an implementation, I'd go with explicitly setting the schema in the SQL because in #279 I ran into

> Fix behavior when using a database/user search_path which has osm2pgsql tables in it, but not in the first table in the list

And the best fix for that is to specify full schema paths if you're using a non-default schema

kachkaev:
Ping guys :–) Any news?

pnorman:
> Ping guys :–) Any news?

There have been no updates by @asbjornst since dec, so nothing has changed.

lonvia:
This PR has change requests which have not been addressed and is not mergable any more. Closing here. The topic remains open though, see #768.

middle-pgsql.cpp
mmd-osm: The README file still lists 9.1 as minimum version, although it isn't supported anymore since more than 2 years now. Should this updated?

https://www.postgresql.org/support/versioning/

PR #630, opened 2016-10-22
allow more numeric datatypes for table columns in C tagtransform

### LINKED ISSUE #621: Crash when integer contains non-integer value
If there is non-integer value in OSM data in tag, that is supposed to be integer in the style file, Osm2pgsql throws error similar to following:

```
Osm2pgsql failed due to ERROR: result COPY_END for planet_osm_point failed: ERROR:  invalid input syntax for integer: "bis 1946"
```

The error could be printed as warning, but Osm2pgsql should deal with cases of misedited OSM key/value and finish processing of the OSM file.

I tested it with on current main branch (and also with version 0.88.1) with following command:

```
./osm2pgsql test.osm -U mtbmap -d gisczech -s -S mtbmap.style
```

[test.osm.txt](https://github.com/openstreetmap/osm2pgsql/files/450965/test.osm.txt)
[mtbmap.style.txt](https://github.com/openstreetmap/osm2pgsql/files/450966/mtbmap.style.txt)

lonvia:
Scans the style file for additional int and real data types
and mappes them all to the existing int and real conversion
routines.

Fixes #621.

### DIFF
--- a/taginfo.cpp
+++ b/taginfo.cpp
@@ -1,5 +1,4 @@
 #include "taginfo_impl.hpp"
-#include "table.hpp"
 #include "util.hpp"
 
 #include <cassert>
@@ -15,13 +14,24 @@
 #endif
 #endif
 
-static const std::map<std::string, int> tagflags = {
+static const std::map<std::string, unsigned> tagflags = {
     {"polygon", FLAG_POLYGON},
     {"linear",  FLAG_LINEAR},
     {"nocache", FLAG_NOCACHE},
     {"delete",  FLAG_DELETE},
     {"phstore", FLAG_PHSTORE},
     {"nocolumn", FLAG_NOCOLUMN}
+};
+
+static const std::map<std::string, unsigned> tagtypes = {
+    {"smallint", FLAG_INT_TYPE},
+    {"integer", FLAG_INT_TYPE},
+    {"bigint", FLAG_INT_TYPE},
+    {"int2", FLAG_INT_TYPE},
+    {"int4", FLAG_INT_TYPE},
+    {"int8", FLAG_INT_TYPE},
+    {"real", FLAG_REAL_TYPE},
+    {"double precision", FLAG_REAL_TYPE}
 };
 
 taginfo::taginfo()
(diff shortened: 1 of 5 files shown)


taginfo_impl.hpp
```diff
@@ -13,15 +13,25 @@ enum column_flags {
...
 
 struct taginfo {
     taginfo();
     taginfo(const taginfo &);
 
+    ColumnType column_type() const {
+        if (flags & FLAG_INT_TYPE)
+            return COLUMN_TYPE_INT;
+        if (flags & FLAG_REAL_TYPE)
+            return COLUMN_TYPE_REAL;
```
pnorman: code style issue: missing {} after ifs

taginfo.cpp
```diff
@@ -24,6 +23,17 @@ static const std::map<std::string, int> tagflags = {
...
 };
 
+static const std::map<std::string, int> tagtypes = {
+    {"smallint", FLAG_INT_TYPE},
+    {"integer", FLAG_INT_TYPE},
+    {"bigint", FLAG_INT_TYPE},
+    {"int2", FLAG_INT_TYPE},
+    {"int4", FLAG_INT_TYPE},
+    {"int8", FLAG_INT_TYPE},
+    {"real", FLAG_REAL_TYPE},
```
pnorman: `double precision` is another real type

taginfo.cpp
```diff
@@ -145,6 +153,13 @@ int read_style_file( const std::string &filename, export_list *exlist )
     temp.type.assign(datatype);
     temp.flags = parse_tag_flags(flags, lineno);
 
+    try {
+        temp.flags |= tagtypes.at(temp.type);
+    } catch (std::out_of_range const &e) {
+        fprintf(stderr, "Bad datatype '%s' in style file line %d\n", datatype, lineno);
```
pnorman: An unknown datatype should be a warning, not an error. e.g. `character varying`, `varchar`, `char(255)` are all valid types for arbitrary OSM data.

Also, it shouldn't warn if using Lua transforms because any column type can be valid there, including stranger ones like enumerated types, `text[]`, `hstore`, and `jsonb`

lonvia:
I've added 'double precision', removed the error on unknown types and fixed the style in all lines touched by this PR.

PR #638, opened 2016-10-31
update to upstream version 2.8.0 of libosmium

lonvia:
Turns out that 2.9.0 still has a bug with parsing coordinates in scientific notation. This updates to 2.8.0 for the moment. The newer version is required for the upcoming integration of libosmium types.


lonvia:
After some discussion with @joto, we decided that a new release that fixes the parsing issue is the better solution here.

PR #641, opened 2016-11-11
This is just for appveyor comes from tilzen

lygstate:


pnorman:
This PR seems to be a mistake - it's trying to merge the development version into the 0.90.x series.

PR #662, opened 2016-12-22
Add config for clang-format

lonvia:
Running clang-format on the current code touches about 5k lines, which is a bit excessive. So I would suggest to require running git-clang-format for all future PRs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,36 +32,23 @@ for easy bug fixes, or if a patch backporting a fix is provided.
 
 ## Code style
 
-The current codebase is a mix of styles, but new code should be written in the
+Code must be written in the
 [K&R 1TBS style](https://en.wikipedia.org/wiki/Indent_style#Variant:_1TBS) with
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -32,37 +32,24 @@ for easy bug fixes, or if a patch backporting a fix is provided.
...
 Names should use underscores, not camel case, with class/struct names ending in `_t`.
 
 Headers should be included in the order `config.h`, C++ standard library headers,
 C library headers, Boost headers, and last osm2pgsql files.
 
+There is a .clang-format configuration avialable and all code must be run through
+clang-format before submitting. You can use git-clang-format after staging all
+your changes:
+
+    git-clang-format -style=file *pp tests/*pp
```
pnorman: `git-clang-format-3.8 -style=file *pp tests/*pp` reports `git-clang-format-3.8: error: unrecognized arguments: -style=file`, but it seems to work with `--style=file`

PR #668, opened 2017-01-04
Save unprojected coordinates in node caches and use osmium dense file array

### LINKED ISSUE #96: Excessive maximum node messages
With an import with `--slim --drop --number-processes 32` (and other flags) I get the following excessive messages about the maximum node ID

```
Going over pending ways...
Maximum node in persistent node cache: 2501902335
        91162940 ways are pending

Using 32 helper-processes
Mid: loading persistent node cache from /mnt/flat/nodes.bin
Maximum node in persistent node cache: 2501902335
Mid: loading persistent node cache from /mnt/flat/nodes.bin
Maximum node in persistent node cache: 2501902335
Mid: loading persistent node cache from /mnt/flat/nodes.bin
Maximum node in persistent node cache: 2501902335
Mid: loading persistent node cache from /mnt/flat/nodes.bin
Maximum node in persistent node cache: 2501902335
Mid: loading persistent node cache from /mnt/flat/nodes.bin
Maximum node in persistent node cache: 2501902335
Mid: loading persistent node cache from /mnt/flat/nodes.bin
Maximum node in persistent node cache: 2501902335
Mid: loading persistent node cache from /mnt/flat/nodes.bin
Maximum node in persistent node cache: 2501902335

...

Process 11 finished processing 2848842 ways in 2377 sec
Process 13 finished processing 2848842 ways in 2377 sec
Process 21 finished processing 2848842 ways in 2377 sec
Process 22 finished processing 2848842 ways in 2377 sec
Maximum node in persistent node cache: 2501902335
Maximum node in persistent node cache: 2501902335
Process 15 finished processing 2848842 ways in 2377 sec
Maximum node in persistent node cache: 2501902335
Maximum node in persistent node cache: 2501902335
Maximum node in persistent node cache: 2501902335
Maximum node in persistent node cache: 2501902335
Maximum node in persistent node cache: 2501902335
Maximum node in persistent node cache: 2501902335
Maximum node in persistent node cache: 2501902335

...

Going over pending relations...
Maximum node in persistent node cache: 2501902335
        0 relations are pending

Using 32 helper-processes
Mid: loading persistent node cache from /mnt/flat/nodes.bin
Maximum node in persistent node cache: 2501902335
Mid: loading persistent node cache from /mnt/flat/nodes.bin
Maximum node in persistent node cache: 2501902335
...
```

I think I got over 72 messages.

### LINKED ISSUE #121: Broken geometries with -E 4326 projection.
Hi, I am using osm2pgsql but I want to import data in EPSG:4326 projection. Therefore, I use the -E 4326 flag. However I get all geometries broken and converted to a big square.

osm2pgsql -c -s -d app -U user -W -H localhost -P 5432 -E 4326 file.osm

I think this is due to having planet_osm_nodes lat / lon columns as integer, which truncates lat / lon decimals and so geometries get wrong coordinates.

Am I doing something wrong? Thanks in advance.

### LINKED ISSUE #542: Store slim coordinates in unprojected units
Storing coordinates as scaled projected units means that it is necessary to scale differently for different projection, leading to bugs like #121 and untested code in #357

Instead, we should store unprojected WGS84 units as they are in the OSM file.

lonvia:
*This is a DB-breaking change.* DBs imported with older code can no longer be updated with this code and there will be **no migration** to the new format. 

Unprojected coordinates in node caches
-------------------------------------------------------

Reprojection of coordinates into the tile coordinate system is delayed until the point when nodes are retrieved from the middle caches (see in particular the `nodes_get_list()` functions). This has a few advantages:

 * OSM coordinates can be saved directly, no need for fix-point magic on reprojected coordinates, which avoids a few rounding errors.
 * no more custom scale factor needed for custom projections (fixes #121)
 * no need for floating-point implementation of caches (supersedes #357) 
 * avoid expensive floating point conversation until last minute

The main drawback is that some coordinates now may need to be reprojected multiple times. This is for example the case for nodes where roads cross each other. OTOH unused nodes are no longer converted.

Osmium Dense File Array
-----------------------------------

The custom flatnode file implementation has been replaces with osmium's dense file array index. Its implementation uses mmap, avoiding custom block management. You can also use osmium-based tools to inspect/update the file outside osm2pgsql.

Furthermore, the dense file array should be thread-safe for reading. The 4th commit in this PR removes the parts where the file is reopened and uses one flatnode store object throughout the import.  (fixes #96)

There are a couple of additional formatting changes created by git-clang-format.

Resolves #542.


middle-pgsql.cpp
```diff
@@ -378,8 +367,9 @@ size_t middle_pgsql_t::local_nodes_get_list(nodelist_t &out, osmium::WayNodeList
     size_t wrtidx = 0;
     for (size_t i = 0; i < nds.size(); ++i) {
         if (std::isnan(out[i].lat)) {
-            std::unordered_map<osmid_t, osmNode>::iterator found = pg_nodes.find(nds[i].ref());
-            if(found != pg_nodes.end()) {
+            std::unordered_map<osmid_t, osmNode>::iterator found =
+                pg_nodes.find(nds[i].ref());
+            if (found != pg_nodes.end()) {
```
pnorman: could this entire loop be an `auto nd:nds`?

node-ram-cache.cpp
```diff
@@ -281,157 +299,169 @@ int node_ram_cache::get_sparse(osmNode *out, osmid_t id) {
...
     }
 
-    return 1;
+    if (coord.valid()) {
+        nodesCacheHits++;
+    }
+    nodesCacheLookups++;
+
+    return coord;
 }
```
pnorman: If any reformatting has been missed by only doing the diffs, I'd just as soon reformat the entire file. Of course, you might have already thought of this ;)

pnorman:
> The main drawback is that some coordinates now may need to be reprojected multiple times. This is for example the case for nodes where roads cross each other. OTOH unused nodes are no longer converted

This should be an overall savings by my estimates.

pnorman:
Should we change something about the slim tables to force an error if you try to update old ones? Right now the tables have gone from storing int4 for each coordinate to storing int4 for each coordinate, so the table DDL is the same.

middle-pgsql.cpp
lonvia: This is part of the code that was just formatted by clang-format. I don't want to touch it at the moment.

node-ram-cache.cpp
lonvia: I can do that in a separate commit immediately before submitting this PR.

lonvia:
> Should we change something about the slim tables to force an error if you try to update old ones?

Yes, I would like that but I'm not sure how to do it. How about adding an additional table that holds a schema version? This would allow us to create better error reporting and also to write automatic migrations in the future.

pnorman:
> Yes, I would like that but I'm not sure how to do it. How about adding an additional table that holds a schema version? This would allow us to create better error reporting and also to write automatic migrations in the future.

I'd be in favour of this going forwards, but I'm not sure it's the best option for **this** issue. Looking at the schema, we have the column `tags text[]` with node tags. We don't need this information, and flat nodes doesn't have it. If we dropped that column it would break compatibility (and save space).

lonvia:
> I'd be in favour of this going forwards, but I'm not sure it's the best option for this issue. Looking at the schema, we have the column tags text[] with node tags. 

It's a bit hacky but good enough for me for the moment. I can add that.

pnorman:
> It's a bit hacky but good enough for me for the moment. I can add that.

Well, I'd be in favour of dropping tags from nodes *anyways* ;)

With that the changes look good to me, and performance is about the same. We should add something to migrations that the slim tables have changed in 0.93.0-dev. Is there anything else we should do pre-merge?

lonvia:
I've added commits for removing the tags column in the nodes table and reformatting node-ram-cache.

I've also had to add a try/catch around the get() function for the dense node cache as unknown ids throw an exception. There is a small penalty attached to that but not enough to block this PR. @joto will look into providing an exception free get() function on the libosmium side at some point.

If rendering results look ok, then this is good to go for me.

joto:
New `get_noexcept()` functions are now available in master for the index maps. https://github.com/osmcode/libosmium/commit/d353993b41f76a4228e5c3849ae77531d8d9367a

PR #670, opened 2017-01-09
Replace osmNode with osmium::geom::Coordinates

lonvia:
Simplifies code around reprojection a little bit.

Also includes a fix for a null-pointer access in output in middle tests.


pnorman:
Looks good to me, but I haven't tested it yet.

PR #679, opened 2017-01-26
Gentoo ebuild

andrew-aladev:
Added gentoo ebuild.


pnorman:
We don't have other distro-specific build information in the repo, I don't think we should have gentoo either.

It's also not clear how we'd maintain this with no test coverage, and with the maintainers on Debian, no way to test it.

lonvia:
I agree with @pnorman. Distro-specific information should go to repos maintained by the distros as there is no way for the maintainers here to keep this information current.

PR #636, opened 2016-10-28
Geos36

darkblue-b:
this built and tests passed, with the following Ubuntu 1404 Trusty config  : 

```
osm2pgsql_geos36/build$ cmake .. -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTS=ON
-- The C compiler identification is GNU 4.8.4
-- The CXX compiler identification is GNU 4.8.4
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Building osm2pgsql 0.91.0-dev
-- Looking for include file termios.h
-- Looking for include file termios.h - found
-- Looking for include file libgen.h
-- Looking for include file libgen.h - found
-- Looking for include file unistd.h
-- Looking for include file unistd.h - found
-- Looking for lseek64
-- Looking for lseek64 - found
-- Looking for posix_fallocate
-- Looking for posix_fallocate - found
-- Looking for posix_fadvise
-- Looking for posix_fadvise - found
-- Looking for sync_file_range
-- Looking for sync_file_range - found
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for stddef.h
-- Looking for stddef.h - found
-- Check size of off_t
-- Check size of off_t - done
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found version "1.2.8") 
-- Looking for include file pthread.h
-- Looking for include file pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- Found EXPAT: /usr/lib/x86_64-linux-gnu/libexpat.so (found version "2.1.0") 
-- Found BZip2: /usr/lib/x86_64-linux-gnu/libbz2.so (found version "1.0.6") 
-- Looking for BZ2_bzCompressInit in /usr/lib/x86_64-linux-gnu/libbz2.so
[...]

PR #684, opened 2017-02-04
Use libosmium for geometry building

lonvia:
This PR replaces the libgeos-based geometry builder with the libosmium area assembly. As libosmium approaches multipolygon assembly in a slightly different way, there are some visible changes to the polygon handling, as well as some secondary changes.

Changes in Polygon Handling
---------------------------

 * *Geometry type is determined by tagging only.* Previously osm2pgsql would create polygons only when a polygon was requested by the tagtransform and the geometry could actually assembled into rings. Otherwise a linestring was added to the *_ways table. This could lead to the odd situation where a partially broken multipolygon had some of its rings in the polygon table and the unclosed ways in the ways table. New behaviour is that only the result of tagtransform (or for output-multi: the requested format for the table) is taken into account. In some corner cases, this can lead to a regression with the C-transform when keys may describe a polygon or a way.
 * *Broken geometries are not recovered.* The libgeos geometry builder was very forgiving with broken polygons and would create a partial solution. The libosmium assembler uses an all or nothing approach. If the geometry is bad, the entire object is discarded. There is also no more buffering of geometries to recover them. On the plus side, libosmium can handle geometries where rings touch in more than one point, which is permissible in OSM but not for OGR polygons.
 * *`--exclude-invalid-polygon` removed.* libosmium never creates invalid geometries.
 * *Line merging has different results.* For route relations a custom line merging algorithm has been added which seems to be much more aggressive than the one employed by libgeos.

Other Changes
-------------

 * Reprojection of geometries now moved into geometry assembler because libosmium can build geometries only in WSG84.
 * Expiry works now on binary wkb only. This actually results in much cleaner code.
 * Some streamlining of middle interface (mostly working more with the native libosmium types).
 * now unused custom nodelist_t and multinodelist_t removed.

Todo
----

- [x] get libosmium area assembler changes into mainline (see https://github.com/osmcode/libosmium/issues/194)
- [x] fix regression tests (need to go each of the differences and check if the results are expected)
- [x] test with larger imports and visually check results
- [x] update documentation


imagico:
Regarding broken geometries:

In the current state of libosmium w.r.t. assembling multipolygons, i.e. without https://github.com/osmcode/libosmium/issues/152 being addressed, this will have a huge impact on users in map rendering.  I have used libosmium quite extensively for data preparation for landcover rendering and there are areas where 20-50 percent of the big landcover polygons are broken in ways currently not recoverable by libosmium, mostly self intersections.  The visual impact on maps will be boldly visible.

That would be great for the standard style because it would bring down the number of broken geometries in the database quite quickly, faster than any QA tool can do, but it would certainly also be problematic for many osm2pgsql users.

Ideally libosmium should be able to handle these errors and you should be able to choose within osm2pgsql how strict you want to be about them.

Additional note:  If you assemble polygons in geographic coordinates and reproject them into other coordinate systems you can get invalid geometries.  This is rare practically with mercator but if you want to make sure you only have valid geometries you need a validity check afterwards.

lonvia:
> Additional note: If you assemble polygons in geographic coordinates and reproject them into other coordinate systems you can get invalid geometries. This is rare practically with mercator but if you want to make sure you only have valid geometries you need a validity check afterwards.

Reintroducing the validity check is something I'd like to avoid. It's expensive and we'd essentially have to reimplement whatever libgeos is doing. So the question is, does this have any practical implications for normal osm2pgsql use cases? If it only happens when using some obscure projections, then post-filtering is probably the better option. If there are practical cases for mercator projection we should check if we can catch these particular ones without going through the full-blown validation process.

lonvia:
> In the current state of libosmium w.r.t. assembling multipolygons, i.e. without osmcode/libosmium#152 being addressed, this will have a huge impact on users in map rendering. I have used libosmium quite extensively for data preparation for landcover rendering and there are areas where 20-50 percent of the big landcover polygons are broken in ways currently not recoverable by libosmium, mostly self intersections. The visual impact on maps will be boldly visible.

[joto's area statistics](http://area.jochentopf.com/stats/) report that approximately 50k out of 242.000k have intersections. @joto tells me that this figure actually includes polygons that are reported multiple times and the real figure is more like 20k self-intersecting polygons of which 6k are relations. That's 0.001% of polygons. It's probably true that some areas where the polygons are concentrated but with this small number, they cannot be large and deserve being fixed. I also think that it will make life easier for mappers and data consumers alike if osm2pgsql takes a more consistent approach to creating polygons. 

> That would be great for the standard style because it would bring down the number of broken geometries in the database quite quickly, faster than any QA tool can do, but it would certainly also be problematic for many osm2pgsql users.

Not necessarily. The average data consumer expects to get the data they see on osm.org when they process the planet. They don't need what they are not aware is not even there. And fixing broken data always comes with the danger of fixing it the wrong way.

imagico:
Regarding reprojection and validity - It is probably not a big deal in most applications - normal rendering for example does not really care if the geometry is valid or not but with

* --exclude-invalid-polygon removed. libosmium never creates invalid geometries.

i wanted to point out that this would not imply osm2pgsql never creates invalid geometries with coordinate systems other than geographic.  The most important thing is to make this clear in the documentation.

Ultimately this is a more generic problem anyway, i.e. the need for a reprojection library that is capable of validity preserving reprojection of linestring/polygon geometries.

> joto's area statistics report that approximately 50k out of 242.000k have intersections. @joto tells me that this figure actually includes polygons that are reported multiple times and the real figure is more like 20k self-intersecting polygons of which 6k are relations.

I was talking about visual impact here - large polygons have a much higher likeliness of being broken than small ones - a bit like ceramic components in engineering :smile:.  At the same time large polygons have a high visual impact in map rendering.  The vast majority of polygons in the OSM database are buildings which are obviously rarely broken.

Note i am not against being stricter with polygons in osm2pgsql, on the contrary.  But you need to expect complaints from users because of this.  To get an idea of the visual impact - if you take the difference in Jochen's comparison map with the removal of old style MPs: http://area.jochentopf.com/map/index.html - the impact of removing all invalid polygons osmium currently does not handle is larger than that at z6-8 (that is for water and wood/forest) in many areas, in particular northern Europe.

woodpeck:
One of the things I do at $DAYJOB is sell shape files generated from OSM data, where I typically discard invalid polygons. I'd like to support imagico's assessment - nobody cares if a house goes missing, but everyone notices if a 50km segment of the Rhone is lacking its riverbank area. So yes, it might be "only" a few thousand relations but there will be many large ones and there will be complaints. On the other hand people who post-process OSM data like me will finally be free from user complaints like "but this forest is visible on the OSM web site, why is it not in your shapefile?", so +1 for dropping invalid stuff. Incidentally I made an attempt to do that 5 years ago in 8be0cfc3caabaac3a48b9202eabefcfaced4647d, maybe this time it gets through ;)

pnorman:
Appveyor is failing with 
```
C:\osm2pgsql\contrib\libosmium\osmium/tags/filter.hpp(41): fatal error C1083: Cannot open include file: 'boost/iterator/filter_iterator.hpp': No such file or directory
```

@alex85k, could this be something that's missing in the binary blobs used to supply the osm2pgsql dependencies?

pnorman:
> Geometry type is determined by tagging only. Previously osm2pgsql would create polygons only when a polygon was requested by the tagtransform and the geometry could actually assembled into rings. Otherwise a linestring was added to the *_ways table. This could lead to the odd situation where a partially broken multipolygon had some of its rings in the polygon table and the unclosed ways in the ways table. New behaviour is that only the result of tagtransform (or for output-multi: the requested format for the table) is taken into account. In some corner cases, this can lead to a regression with the C-transform when keys may describe a polygon or a way.

How will this work with `man_made=pier`?

pnorman:
I've set up https://osmium.osm2pgsql.paulnorman.ca. The data on the right is from 170125 and imported with `./osm2pgsql -d osm2pgsql_test -S ../default.style --slim --drop --flat-nodes nodes.bin --cache 34000 ~/planet-170125.osm.pbf`

lonvia:
The missing dependency for Appveyor is something we probably can (and would want to) get rid of in the code. So it's okay to leave it for now.

lonvia:
> How will this work with man_made=pier?

That's the regression. It will be fine for the lua transform, as polygon can be set according to value not only key but it won't work for the C transform anymore. If that is not acceptable than we need to reintroduce the old behaviour for ways.

joto:
Regarding the fixing of broken multipolygons: The Mapbox data team is actively fixing broken multipolygons that have a large visible impact on the map. They have some kind of prioritized list. I am also working on Maproulette challenges to get the community involved in fixing.

imagico:
> https://osmium.osm2pgsql.paulnorman.ca

Great, that reflects my experiences - large errors particularly exist in southern Sweden and Finland:

https://osmium.osm2pgsql.paulnorman.ca/#8/61.090/27.339
https://osmium.osm2pgsql.paulnorman.ca/#8/59.043/9.470

But also elsewhere:

https://osmium.osm2pgsql.paulnorman.ca/#8/30.121/114.456

As further example you can also find the geometries of all water areas not processed by libosmium processed with https://github.com/imagico/osmium/tree/multipolygons-fix

on http://www.imagico.de/files/waterareas_error.db.gz

which includes geometries current osm2pgsql fails with like here http://www.openstreetmap.org/#map=7/67.902/123.563

All in all not a big deal, this can be handled and should be handled but it is big enough a problem so that it cannot be simply ignored.

pnorman:
> That's the regression. It will be fine for the lua transform, as polygon can be set according to value not only key

That doesn't really help. man_made=pier doesn't have a clear meaning of area vs line based on tags alone

> but it won't work for the C transform anymore. If that is not acceptable than we need to reintroduce the old behaviour for ways.

Because osmium.osm2pgsql.paulnorman.ca is working, I wonder if I built the wrong osm2pgsql branch.

We need unclosed waterway=river to go to the lines table and closed waterway=riverbank to go to the polygon table. If we can't do that, we might as well get rid of the C transforms. Not that I wouldn't welcome that for other reasons.

lonvia:
> > but it won't work for the C transform anymore. If that is not acceptable than we need to reintroduce the old behaviour for ways.
> 
> Because osmium.osm2pgsql.paulnorman.ca is working, I wonder if I built the wrong osm2pgsql branch.

No, the branch is right, I was badly remembering my own code. For ways the rules is: it goes into the polygon table if tagtransform makes it a polygon and the way is closed. So the main change is that closed ways with a broken geometry (e.g. a self-intersecting way) won't end up in the line table anymore. They are discarded. I guess that's an acceptable loss.

wkb.hpp
```diff
@@ -0,0 +1,361 @@
...
+    T read_data()
+    {
+        auto *data = reinterpret_cast<T const *>(m_wkb + m_pos);
+        m_pos += sizeof(T);
+
+        return *data;
+    }
+
+    char const *m_wkb;
+    unsigned m_pos;
```
pnorman: should this be `size_t`? msvc is generating warnings about conversion from `size_t` to `unsigned int` elsewhere?

lonvia:
libosmium master branch now contains the necessary changes for this PR. I've updated our version in contrib to current master. We should remember though to update to a released version of libosmium before doing a new osm2pgsql release.

pnorman:
> libosmium master branch now contains the necessary changes for this PR. I've updated our version in contrib to current master. We should remember though to update to a released version of libosmium before doing a new osm2pgsql release.

Do we need to wait for a new release for this PR?

Do I need to reimport on my test server?

lonvia:
> Do we need to wait for a new release for this PR?

No. It's just an requirement for the release because at least Debian builds against it's external libosmium package.

> Do I need to reimport on my test server?

Yes, that would be a good time to do it. I've fixed the regression tests and a few bugs that they uncovered, so from a technical side this PR is mostly done. (Still left to do: running updates and the Nominatim tests and @pnorman's comments.)

That leaves the issue of changing behaviour for invalid polygons. I agree that @imagico's examples are quite visible on the map but as far as I can see, they are caused by relatively few errors. I still think that the problem solves itself quickly once the errors are visible. What would be needed is to give ample warning that this change will happen and why, and that osm.org switches to the newest version of osm2pgsql as soon as it is released. We should have a good plan for both these points before merging the PR.

wkb.hpp
lonvia: Done.

pnorman:
Numbers from a test with master (b9d7fcc162aabda0e972fba4a6a616cb0da5eaa0) and this PR (0333e11adff2a7e09fb3ae00a0de69ae5dc87ad6)

pr
```
Node stats: total(3706912087), max(4630874436) in 548s
Way stats: total(389670834), max(468563887) in 2308s
Relation stats: total(4734167), max(6908001) in 3582s
Finished processing 267057024 ways in 4428 s
All indexes on planet_osm_roads created in 903s
All indexes on planet_osm_point created in 2347s
All indexes on planet_osm_line created in 5859s
All indexes on planet_osm_polygon created in 9432s
Osm2pgsql took 20310s overall
User time (seconds): 11826.25
Maximum resident set size (kbytes): 40146944
Involuntary context switches: 69054
```

master
```
Node stats: total(3706912087), max(4630874436) in 639s
Way stats: total(389670834), max(468563887) in 2864s
Relation stats: total(4734167), max(6908001) in 4775s
Finished processing 267057051 ways in 6517 s
All indexes on planet_osm_roads created in 916s
All indexes on planet_osm_point created in 2468s
All indexes on planet_osm_line created in 5876s
All indexes on planet_osm_polygon created in 9605s
Osm2pgsql took 24412s overall
User time (seconds): 29804.55
Maximum resident set size (kbytes): 39976848
Involuntary context switches: 2654796
```

This is saving over an *hour* on a 6 hour import. Interestingly, master is more parallel, probably because it has more work to do.

jburgess777:
That is an impressive reduction of CPU from 30k to 12k seconds. Can that be explained by the simpler geometry processing?
Given that the overall time is 20k seconds, the bottleneck must be somewhere other than osm2pgsql.

lonvia:
> That is an impressive reduction of CPU from 30k to 12k seconds. Can that be explained by the simpler geometry processing?

Yes. Using geos functions to build OSM geometries was not a straightforward process while libosmiums algorithm is tailored for the peculiarities of OSM data. There is also a lot less memory allocation going on when preparing the data as more data is processed directly in place in the libosmium buffers.

> Given that the overall time is 20k seconds, the bottleneck must be somewhere other than osm2pgsql.

Our biggest bottleneck is still postgres. Given that all queries (including inserts) are done synchronous, osm2pgsql probably spends a lot of time waiting for results.

Looking at the numbers it might be worth testing if reducing the number of parallel threads won't actually be faster.

lonvia:
Performance numbers for a planet import on a 32GB RAM machine with SSDs (-C 24000 --drop -s):

PR
```
Processing: Node(3467864k 5376.5k/s) Way(361033k 31.52k/s) Relation(4392840 337.86/s)  parse time: 25102s
Node stats: total(3467864331), max(4318974389) in 645s
Way stats: total(361033871), max(432971616) in 11455s
Relation stats: total(4393429), max(6435395) in 13002s
Finished processing 244505022 ways in 7390 s
```

master 
```
Processing: Node(3467864k 4051.2k/s) Way(361033k 25.98k/s) Relation(4393000 254.80/s)  parse time: 31992s
Node stats: total(3467864331), max(4318974389) in 856s
Way stats: total(361033871), max(432971616) in 13894s
Relation stats: total(4393429), max(6435395) in 17242s
Finished processing 244505051 ways in 10367 s
```

The PR database has about 66k polygons less coming from about 20k objects. Summing up the areas of all polygons, there is about 1% less. So as suspected, there are in particular large areas missing.

pnorman:
On the planet the PR is 
- 0.028% (71k/255M) fewer rows in planet_osm_polygon;
- 0.25% (6.94e+12/2.81749e+15) less ST_Area(way) in planet_osm_polygon;
- 0.28% (8.02e+12/2.83212457458888e+15) less way_area in planet_osm_polygon;
- 0.8% (1M/131M) fewer rows in planet_osm_line;
- 0.016% (2k/11M) more rows in planet_osm_roads; and
- unchanged for planet_osm_point

lonvia:
I have finished all testing. Code-wise this is good to go.

pnorman:
I'm also okay with it, it's now just a question of release management.

lonvia:
@joto informed me this morning that I might have broken the expiry code. It assumes that the WKBs that are returned from the database have native endianess. That is not necessarily the case. It is trivial to fix but needs to be done before the PR can be merged.

pnorman:
> It assumes that the WKBs that are returned from the database have native endianess

Would this also be an issue in putting geoms into the DB?

lonvia:
> Would this also be an issue in putting geoms into the DB?

No. Each WKB geometry contains a flag that indicates the endiness used, so the DB must make sure to interpret the string correctly.

lonvia:
It wasn't trivial after all, so I've only added a check that the endianess is as expected the native endianess of the machine. We can add an endian flip if somebody should ever hit the exception.

wkb.hpp
```diff
@@ -230,6 +230,10 @@ class parser_t
             }
         }
 
+        if (out[0] != Endian)
+            throw std::runtime_error(
+                "Endian setting of database WKB not supported.");
+
```
pnorman: Can we say what the endianness of osm2pgsql is in the error message?

pnorman: Or better yet, something like "Different endian settings of database and osm2pgsql are not supported. Database is big-endian, osm2pgsql is little-endian."

lonvia: Done.

ilovezfs:
@pnorman is there a new release in the offing then?

pnorman:
There aren't plans for an immediate release. First we want to let any bugs after a big change like this have a a chance to shake out, but we'd like to get any other changes that involve schema changes into the same release as #668

ilovezfs:
@pnorman thanks. Any suggestion on how we should handle the Homebrew formula at this point? See https://github.com/Homebrew/homebrew-core/blob/master/Formula/osm2pgsql.rb#L21-L26

It's using
```
  patch do
    url "https://github.com/openstreetmap/osm2pgsql/pull/636.patch"
    sha256 "54aa12fe5a3ebbc9ecc02b7e5771939b99f6437f5a55b67d8835df6d8d58619a"
  end
```
but that seems somewhat inappropriate at this point since #636 will never be merged.

pnorman:
osm2pgsql has no support for geos 3.6 in the 0.92.x branch and is not likely to get it with the need to maintain 3.5 compatibility and the need to verify a patch, particularly around memory leaks. Because we've moved away from geos, it's not something many people are likely to work on.

ilovezfs:
@pnorman Understood. A new release would be much appreciated so we can drop the patch.

PR #712, opened 2017-03-10
split tagtransform into two classes and an interface

lonvia:
Splits tagtransform.cpp into the lua and c transform part. The code is exactly the same with some clang reformatting and a couple of explicit int conversions added to suppress type warnings.


pnorman:
clang on OS X doesn't like part of this

```
/Users/travis/build/openstreetmap/osm2pgsql/output-pgsql.cpp:583:13: error: moving a temporary object prevents copy elision [-Werror,-Wpessimizing-move]
            std::move(tagtransform_t::make_tagtransform(&m_options));
            ^
/Users/travis/build/openstreetmap/osm2pgsql/output-pgsql.cpp:583:13: note: remove std::move call here
            std::move(tagtransform_t::make_tagtransform(&m_options));
            ^~~~~~~~~~                                             ~
```

tagtransform.hpp
```diff
@@ -16,41 +16,75 @@ extern "C" {
...
+    virtual unsigned filter_rel_member_tags(
+        taglist_t const &rel_tags, multitaglist_t const &member_tags,
+        rolelist_t const &member_roles, int *member_superseded,
+        int *make_boundary, int *make_polygon, int *roads,
+        export_list const &exlist, taglist_t &out_tags,
+        bool allow_typeless = false) = 0;
+};
 
-class tagtransform {
+class c_tagtransform_t : public tagtransform_t
```
pnorman: Should these be in `tagtransform-c.hpp`/`tagtransform-lua.hpp` to match the cpp files?

lonvia:
std::move fixed.

> but I'm wondering, do we want to change the C++ interface for tagtransform?

I'm not sure what you mean.

If you are worried that this PR changes the interface, then that is not the case. I've only changed the name of the class to match our naming conventions, otherwise the public interface is the same.

If you mean that we should change the C++ interface further than what this PR does, then the answer is: absolutely. I just would like to have proper C++ code before doing any semantic changes.

tagtransform.hpp
lonvia: I can't say I'm very happy about spreading the code over so many files but it would be more consistent with the rest of the code. I will change it.

PR #716, opened 2017-03-13
Ignore relations with more than 32k members

### LINKED ISSUE #713: pgsql-mid fails on relations with more than 32767 members
pgsql-mid stores the member offset for relations in the planet_osm_rels table as smallints (way_off, rell_off columns). That means that the number of members it can safely handle is restricted to 32767. However, the limit is nowhere checked. Trying to import a larger relation leads to the error:

     Osm2pgsql failed due to ERROR: insert_rel failed: ERROR:  value "37945" is out of range for type smallint

This happened recently with relation 7066589.

osm2pgsql should drop relations that are too large early in the processing. What would a reasonable upper limit be?

lonvia:
There is a hard restriction in the pgsql middle tables, where relation members are indexed using a smallint. This change just restricts to the technical limit to fix the issue with the diff file.

Fixes #713.

### DIFF
--- a/parse-osmium.cpp
+++ b/parse-osmium.cpp
@@ -176,6 +176,9 @@ void parse_osmium_t::relation(osmium::Relation const &rel)
     if (rel.deleted()) {
         m_data->relation_delete(rel.id());
     } else {
+        if (rel.members().size() > 32767) {
+            return;
+        }
         if (m_append) {
             m_data->relation_modify(rel);
         } else {


mboeringa:
@lonvia ,

Just a minor remark: It may be difficult or impossible to implement at this particular stage of processing or code, but wouldn't it be better to reject the entire changeset containing the offending relation wholesale, instead of just dropping the relation?

If I understand the code right, this will just drop the relation, but potentially leave an inconsistently imported changeset in the database.

pnorman:
> Just a minor remark: It may be difficult or impossible to implement at this particular stage of processing or code, but wouldn't it be better to reject the entire changeset containing the offending relation wholesale, instead of just dropping the relation?

osm2pgsql has nothing to do with changesets and really knows nothing about them. It doesn't import changesets, which are intended to be useful to humans. Unless `-x` is used it doesn't store this information in the DB.

Even if it were possible, I'd still do it this way.

pnorman:
It's also worth noting that this PR **doesn't** introduce a restriction, the restriction was always there.

Issue #659, opened 2016-12-21
Lua relation members gets post-way transform member tags


pnorman:
I am working on a transform which creates a table with wood areas, a name column, and no other transform-generated columns.

This means I want a way function like

```lua
function wood_ways(tags, num_keys)
    if tags["natural"] == wood then
        local cols = {}
        cols.name = tags["name"]
        return 0, cols, 1, 0
    end
    return 1, {}, 0, 0
end
```

This works properly, turning a way with `natural=wood` into a row with no transform-generated columns

If I am only concerned with old-style MPs, I can then add 

```lua
function wood_rels (tags, num_keys)
    if (tags["type"] == "multipolygon" and tags["natural"] == wood) then
        return 0, tags
    end
    return 1, {}
end

--- test function
-- @param tags OSM tags
-- @param member_tags OSM tags of relation members
-- @param member_roles OSM roles of relation members
-- @param membercount number of members
-- @return filter, cols, member_superseded, boundary, polygon, roads
function wood_rel_members (tags, member_tags, member_roles, membercount)
    assert(membercount == 2) -- hard-coded to avoid loops to generate member_tags
    if tags["natural"] == wood then
        local cols = {}
        cols.name = tags["name"]
        return 0, {}, {0, 0}, 0, 1, 0
    end
    return 1, {}, {0, 0}, 0, 0, 0
end
```

It's not obvious, but member_tags is member tags **after** transformation by the way function, i.e. columns to be output to PostgreSQL, not original tags.

This causes a problem when I ~~want~~need to handle old-style multipolygons, like a relation with `type=multipolygon` and two members, one with `natural=wood` and the other with no tags.

The output of `wood_ways` for the first way is `0, {}` and the second is `1, {}`. This means the function called for rel members is
```lua
wood_rel_members({type="multipolygon"}, {{}, {}}, {"",""}, 2)
```

This is the same as any two-member MP which does not have any tags on ways, which does not have natural=wood on any ways, or that does have natural=wood on the ways, but also has different tags (e.g. `leaf_type` key). Because of this, it's impossible to correctly handle old-style MPs with the above way transform.

Although I made this example with the multi backend, it shows up if you do the same thing with the pgsql backend. The difference is people tend to either **modify** existing keys when generating columns (e.g. @SomeoneElseOSM) or **add** new columns (e.g. @systemed), but seldom replace all the columns with the pgsql backend like I am doing.

[...]

PR #721, opened 2017-03-18
Don't pre-filter relation member tags

### LINKED ISSUE #605: OSM relations and hstore
Hello, I have a question regarding the style file, OSM Relations and the hstore option in osm2pgsql.
I would like to filter my imported data before it gets in my database. For that I use the stylefile and define the tags I would like to keep. I also would like to keep the other tags of the elements which I set in the stylefile thats why I use the hstore option --hstore-match-only. It seems to work for my node table but my line and polygon tables are full with data which I don't need.
Unfortunately it seems as if osm2pgsql pushes all the OSM-Relations in my db wheter there is a connection to my tags in the stylefile or not. 
Is it possible to use a filter function for the relations similar to the filter in the style file? I actually thought the relations would be filtered as well like the nodes and ways.
I also read about the luna file but it seems a little bit complicated so I thought there might be an easier way which I haven't discovered yet. Any help is appreciated.

stylefile
stylefile

Output
relationproblem

lonvia:
Tags for ways of multipolygons are no longer prefiltered before handing them into the relation member filter function. For new-style MPs the tags are not relevant anyways. For old-style MPs, the filter now first creates a list of common tags and then filters the result with the standard relation tags filter.

I have adapted style.lua but noticed that old-style MPs are handled slightly differently. The tags of the MP are computed as the union of all tags on the out ways while the C transform computes them as the intersection of tags. I haven't touched multi.lua yet.

Tests pass but it could possibly do with visual verification that results are the same.    

Fixes #605 and #659.

### DIFF
--- a/geometry-processor.cpp
+++ b/geometry-processor.cpp
@@ -94,19 +94,6 @@ size_t relation_helper::set(osmium::Relation const &rel, middle_t const *mid)
     return num_ways;
 }
 
-multitaglist_t relation_helper::get_filtered_tags(tagtransform_t *transform,
-                                                  export_list const &el) const
-{
-    multitaglist_t filtered(roles.size());
-
-    size_t i = 0;
-    for (auto const &w : data.select<osmium::Way>()) {
-        transform->filter_tags(w, nullptr, nullptr, el, filtered[i++]);
-    }
-
-    return filtered;
-}
-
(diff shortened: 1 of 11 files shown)


pnorman:
This is on my list to review, but I'm debugging #735 first.

PR #736, opened 2017-04-07
Include wikidata and wikipedia columns as well

dersvenhesse:
Since most of the tags are included in the default style, I feel like `wikidata` and `wikipedia` should be included as well.


pnorman:
There are two problems with this. The first is that the default style file (and C transforms) is unlikely to change without a great detail of consideration and very good reasons. The second is that default.style is targeted at rendering, and I'm not aware of anyone doing rendering the wikipedia or wikidata tags except as a debugging layer. It certainly isn't common.

PR #737, opened 2017-04-08
Gazetteer: use hstore column for address parts

lonvia:
Removes all address columns and adds a generic hstore instead. This way, we have the origin for all address parts later in the geocoder.

### DIFF
--- a/output-gazetteer.cpp
+++ b/output-gazetteer.cpp
@@ -15,17 +15,12 @@
 #include <iostream>
 #include <memory>
 
-#define ADMINLEVEL_NONE 100
+enum : int { MAX_ADMINLEVEL = 100 };
 
 void place_tag_processor::clear()
 {
     // set members to sane defaults
     admin_level = ADMINLEVEL_NONE;
-    countrycode = 0;
-    housenumber.assign("\\N");
-    street = 0;
-    addr_place = 0;
-    postcode = 0;
 
     places.clear();
     extratags.clear();
@@ -64,9 +59,6 @@ void place_tag_processor::process_tags(osmium::OSMObject const &o)
     osmium::Tag const *landuse = 0;
     bool isnamed = false;
     bool isinterpolation = false;
-    char const *house_nr = 0;
-    char const *conscr_nr = 0;
-    char const *street_nr = 0;
 
     clear();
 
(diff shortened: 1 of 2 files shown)


lonvia:
The name column already uses hstore. I really wouldn't recommend using gazetteer outside the nominatim import. That's why it is imo okay that the documentation is sparse.

roques:
I'm not convinced that this is a desirable change. Yes, having the whole address as one `hstore` value is probably nicer to pass around, but is this comfort worth the cost?
I've spent some time looking into the innards of Nominatim and reducing its storage requirements and cache footprint by choosing smaller types and playing "column Tetris". These changes had a nice effect on freshly loaded, indexed, and VACUUM FULLed extracts of Germany, Switzerland, and Italy:

table | original | tetris | tetris + hstore-address
----- | -------- | ----- | ----------------------
place | 8522 MB | 8310 MB | 9367 MB
placex | 16693 MB | 15519 MB | 16757 MB

But, as you can see the space wasted by using `hstore` for address outweighs my savings.
An `hstore` uses 1 + 4 + 2×N×4 + size(keys) + size(values) (+3 (if more than 63)) bytes to store N key/value pairs. This sums up to 23 wasted bytes for each housenumber alone. Whereas adding an additional column just takes one bit (in the null-map) + size(value) if not NULL (when properly aligned).

lonvia:
The change is not for making things easier to pass around, it's for retaining the information about the key, to fix issues like https://github.com/openstreetmap/Nominatim/issues/148. Using hstore instead of columns for each address part gives us a lot more freedom down the line to change search and display strategies for placex entries without having to change the database schema again.

Maybe you would like to share the findings of you 'column tetris'? From the numbers it looks like there is still a lot to gain for the placex table even with the new address hstore. If so, please open a new issue in the Nominatim issue tracker about it.

roques:
I can understand why we don't want to ignore / throw away any of the `addr:*` information; After all, there probably was a reason why it was entered in the first place. However, can't we have both, efficient storage for the common case and no lost information? You could use columns for the "usual" address information which is already explicitly interpreted, like `addr:housenumber`, `addr:street`, `addr:place`, `isin` and store all other, currently uninterpreted `addr:*` fields in the new HSTORE column, making it NULL in the common case?

PR #738, opened 2017-04-08
JSONB tags support for slim tables.

gardster:
OSM tags are very similar to JSON fields. So I think that it will be great to have JSON data representation in database. Postgresql allows indexes on JSON fields and store this data more effectively than plain text arrays.
I made some tests on my dataset (USA East Coast) and have this numbers:

|method|tags size in _ways|tags size in _rels|
|--------|-----------------:|---------------:|
| array   | 4422 mb               |      36 mb         |
| jsonb   | 3805 mb              |      32 mb         |


middle-pgsql.cpp
```diff
@@ -75,10 +75,11 @@ namespace {
 // Argument should point to beginning of literal, on return points to delimiter */
```
pnorman: The comments need updating as this function no longer takes arrays, but takes json.

You also need to handle arbitrary JSON inputs. It's acceptable to exit on json which is not a bunch of string key:value mappings, but you need to handle any json within that.

I'm not sure *what* differences that entails, but one obvious one is escaping. JSON allows more escape sequences than just `\n` and `\t`

pnorman:
ping @gardster

gardster:
@pnorman I really want to add support of a JSONB support to the osm2pgsql not only for slim tables. So I think, that it will be good to make an option that will enable this support. Now I'm investigating possibilities to support this but it requires huge refactoring in the code of database query generation. I'm stuck in it a little bit. I think, that if I could make a query generator with different tags data type, I will easily support a hstore datatype for a fallback.
P.S. Do you think that we must keep 9.2 and 9.3 support when main OSM database were upgraded?

pnorman:
> P.S. Do you think that we must keep 9.2 and 9.3 support when main OSM database were upgraded?

osm2pgsql is more than just tile.osm.org. We need to maintain compatibility with supported PostgreSQL versions.

pnorman:
I did some work at https://github.com/pnorman/osm2pgsql/tree/hstore_middle at converting. It works, but deserialization isn't yet changed over and I'm doing that in postgres temporarily.

woodpeck:
@pnorman, https://github.com/Nakaner/pg-array-hstore-parser might help with deserialization.

gardster:
@pnorman it takes from me more time than I expected.
I made some refactoring in SQL query generation functions. Also I move all middle files to the separate folder and extract tags encoding routines to separate files. Now we can hide tags encoding logic behind the unified interface. I add a special flag to enable the jsonb tags column. Please take a look on this changes.

lonvia:
We really should have a strategic discussion about JSONB again but that should go into an issue. Closing this PR as it has been stale for more than a year.

Issue #709, opened 2017-03-10
Rewrite tile expiry (management of expired tiles)


Nakaner:
I am currently rewriting the tile expiry of osm2pgsql because its code is not nice, difficult to understand and it seems to me that it does not what it is expected to do. As part of my master thesis I wrote a [database import tool](https://github.com/Nakaner/Cerepso) which has a tile expiry. @lonvia asked me a few months ago to port its tile expiry to osm2pgsql.

The work is done in my fork of this repository in the `expire-as-set` branch.

I have already replaced the self-made tree structure by an unordered set of `uint64_t`.

Because the new tile expiry is not fully backward-compatible, I'm opening this issue. The format of the written expiry file is the same – one tile per line, each line follows the pattern `zoom/x/y`. In difference to the old code, my code writes out all tiles on all requested zoom levels if something in the tile[1] has changed. However, the old tile expiry writes out a tile on a zoom level if all its subtiles are expired or it is a tile on the maximum zoom level.

If you [convert](https://github.com/Nakaner/expiries2shp) the expiry log files into shape files and visualize them using QGIS, it will look like this (area fill is semi-transparent):

![osm2pgsql's current tile expiry converted to a shape file and visualized using QGIS](https://cloud.githubusercontent.com/assets/3611273/23796488/7287ade2-059b-11e7-8ede-75c9d56480d7.png)

My code would generate an expiry list which will look like this (rendered without an area fill):

![asset-expiry](https://cloud.githubusercontent.com/assets/3611273/23796487/7286f9c4-059b-11e7-9917-c71448cdc612.png)

EDIT: The lower image does not show all zoom levels because otherwise the image would only contain black lines and it would be almost opaque.

The raw data of the images can be [downloaded here](https://michreichert.de/share/public/osm2pgsql/expiry-example-shapes.zip).

I can see only one advantage in this behaviour – reducing the size of the file. But this behaviour is not documented anywhere and there is a disadvantage: Every program which reads the tile expiry list, has to derive the list of subtiles on its own. I think that the size of the file does not matter that much because you usually don't transfer it over a network.

[1] i.e. the method `expire_tiles::expire_tile(uint32_t x, uint32_t y)` has been called. The selection of tiles is also handled by my rewrite but out of scope of this issue.

SomeoneElseOSM:
[...]

PR #747, opened 2017-04-20
Rewrite tile expiry (management of expired tiles)

Nakaner:
This pull request contains the changes announced in #709.

This pull request is the first part of a rewrite of the tile expiry. It is a port of the tile expiry of [my own](https://github.com/Nakaner/Cerepso) PostgreSQL import tool to osm2pgsql which I wrote as part of my master thesis. I needed an importer with a tile expiry which handles relations in much different way than osm2pgsql does and the tile expiry had to be more exact.

Each tile is represented by a 64-bit integer (bits of Y and X index interleaved). The set only contains the tiles at the highest zoom level where the list of expired tiles should be generated. All lower requested zoom levels are calculated during writing the tile IDs to the output file by shifting the bits of the tile IDs.

I did a performance test with an import of the Europe extract by Geofabrik to ensure that it has no negative impact on the performance. Time and memory consumption was measured using `/usr/bin/time -v`. The input data was located on a hard disk, flat nodes file and the database was located on SSDs.

1. Check out the `master` branch.
2. Import using `osm2pgsql --merc --slim --flat-nodes /ssd/michael-flatnodes/michael-flat-nodes.cache -C 40000 --style ../default.style --unlogged --number-processes 2 --database gis europe-170217.osm.pbf` took 24 hours and 35 minutes and needed 26,765,608 kB. During these 24 hours the machine was busy producing a subset of Geofabrik's extracts offered on download.geofabrik.de for internal purposes, i.e. don't take these numbers serious.
4. Save a copy of the flat nodes file.
5. Shut down the PostgreSQL database.
6. Copy the tablespace of the database `gis` to a backup location.
7. Start the database.
8. Apply a diff using `osm2pgsql -d gis --slim --merc -S ../default.style --flat-nodes /ssd/michael-flatnodes/michael-flat-nodes.cache --append --number-processes 2 -C 25000 -o master-433.log --expire-bbox-size 20000 -e 10-16 433.osc.gz`. (56 minutes 28 seconds, 9,143,644 kB RAM)
9. Check out branch `expiry-as-set` and build osm2pgsql. The original import could be reused because the differences between both branches were small.
10. Restore the backup of the flat nodes file.
11. Shut down the PostgreSQL database.
12. Restore the tablespace of the database `gis` from the backup.
13. Start the database.
14. Apply a diff using `osm2pgsql -d gis --slim --merc -S ../default.style --flat-nodes /ssd/michael-flatnodes/michael-flat-nodes.cache --append --number-processes 2 -C 25000 -o master-433.log --expire-bbox-size 20000 -e 10-16 433.osc.gz`. (57 minutes 34 seconds, 9,389,628 kB RAM)

The cronjob which produces the OSM planet extracts day by day on this machine did not run during step 8 and 14. The result is within the measurement accuracy. The code line responsible for writing the expiry file was removed in both test cases to ensure that writing larger output files (see #709) does not influence the measurements.

If you request tile expiry for smaller zoom levels (e.g. 8 to 12), both branches consume about 500 MB RAM.

This is the first part of a large rewrite of the tile expiry. If you are interested in the ongoing work, have a look at the branch [`tile-selection-rewrite`](https://github.com/Nakaner/osm2pgsql/tree/tile-selection-rewrite) at my fork of this repository. That branch will rewrite the selection of the tiles which will be marked as expired.

### DIFF
--- a/expire-tiles.cpp
+++ b/expire-tiles.cpp
@@ -41,20 +41,25 @@ tile_output_t::~tile_output_t()
     }
 }
 
-void tile_output_t::output_dirty_tile(int x, int y, int zoom)
+void tile_output_t::output_dirty_tile(uint32_t x, uint32_t y, uint32_t zoom)
 {
     if (outfile) {
         fprintf(outfile, "%i/%i/%i\n", zoom, x, y);
+        ++outcount;
+        if (outcount % 1000 == 0) {
+            fprintf(stderr, "\rWriting dirty tile list (%iK)", outcount / 1000);
+        }
     }
 }
 
-void expire_tiles::output_and_destroy(const char *filename, int minzoom)
+void expire_tiles::output_and_destroy(const char *filename, uint32_t minzoom)
 {
     tile_output_t output_writer(filename);
     output_and_destroy<tile_output_t>(output_writer, minzoom);
 }
 
-expire_tiles::expire_tiles(int max, double bbox, const std::shared_ptr<reprojection> &proj)
+expire_tiles::expire_tiles(uint32_t max, double bbox,
+                           const std::shared_ptr<reprojection> &proj)
 : max_bbox(bbox), maxzoom(max), projection(proj)
 {
     if (maxzoom >= 0) {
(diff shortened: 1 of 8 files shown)


lonvia:
Do you mind rebasing on the latest master version? That will fix the appveyor build.

Nakaner:
@lonvia wrote:
> Do you mind rebasing on the latest master version? That will fix the appveyor build.

I rebased my work yesterday on master and found out today that some tests failed. That failure was not (only) caused by a misconfiguration of AppVeyor. I fixed that today and rebased onto master again.

pnorman:
cc @zerebubuth @woodpeck as people who I believe are using osm2pgsql tile expiry

lonvia:
Code looks good and simplifies the tile expiry a lot. So overall I'm in favour of merging this.

There is still the open question if the changed output format (outputting all changed tiles vs. only outputting the lowest zoom level). #709 did not yield any comments strongly in favour or against changing the output format. Given that, I'm leaning slightly towards taking this PR as is with the output of all tiles. It simplifies the code in osm2pgsql and potentially also on the consumer side. Any other opinions? @pnorman?

andrew-aladev:
Hello. Do you plan to add tile expiry to `output-gazetter`?

Nakaner:
@andrew-aladev wrote:
> Hello. Do you plan to add tile expiry to output-gazetter?

No. Nominatim is not tile-based, is it?

andrew-aladev:
Yes, But how to determine whether anything in database has been updated? Today nominatim uses `update --index` after any update.

PR #751, opened 2017-04-22
Reorder columns for gazetteer place table

lonvia:
This is part of  https://github.com/openstreetmap/Nominatim/pull/703 by @roques. See PR for details.

### DIFF
--- a/output-gazetteer.cpp
+++ b/output-gazetteer.cpp
@@ -559,12 +559,12 @@ int output_gazetteer_t::start()
 
         std::string sql =
             "CREATE TABLE place ("
-            "  osm_type CHAR(1) NOT NULL,"
             "  osm_id " POSTGRES_OSMID_TYPE " NOT NULL,"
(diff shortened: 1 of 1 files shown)


output-gazetteer.cpp
```diff
@@ -559,12 +559,12 @@ int output_gazetteer_t::start()
 
         std::string sql =
             "CREATE TABLE place ("
-            "  osm_type CHAR(1) NOT NULL,"
             "  osm_id " POSTGRES_OSMID_TYPE " NOT NULL,"
+            "  osm_type \"char\" NOT NULL,"
```
pnorman: quotes aren't required, and we should keep in caps to be consistent `CHAR`

lonvia: From https://www.postgresql.org/docs/current/static/datatype-character.html:

> The type "char" (note the quotes) is different from char(1) in that it only uses one byte of storage. It is internally used in the system catalogs as a simplistic enumeration type.

PR #752, opened 2017-04-26
go back to SRID 900913

jbuyl:
this was needed to keep updating because the original osm2pgsql uses 3857 instead of the old 900913, wasn't sure where to replace so just replaced 3857 with 900913 everywhere


jbuyl:
please just delete, didn't intend to create a PR in original fork

pnorman:
See migrations for documentation on this: https://github.com/openstreetmap/osm2pgsql/blob/master/docs/migrations.md#091-default-projection

jbuyl:
Thank you, it was more desirable in our case to make a change in the code if it works.

PR #755, opened 2017-05-06
multi-output: reintroduce line spliting

### LINKED ISSUE #726: Multilinestrings built for relations with multi backend
I had to apply #725, but with a much simplified cleartables I get

```
pnorman@pippin:~/osm/ClearTables/ClearTables$ ~/osm/osm2pgsql/build/osm2pgsql -d osm2pgsql_test --number-processes 1 --output multi --style cleartables.json --flat-nodes nodes.bin ~/osm/osm2pgsql/tests/liechtenstein-2013-08-03.osm.pbf
osm2pgsql version 0.93.0-dev (64 bit id space)

Using lua based tag processing pipeline with script admin.lua
Using lua based tag processing pipeline with script admin.lua
Using lua based tag processing pipeline with script place.lua
Using projection SRS 3857 (Spherical Mercator)
Setting up table: admin_area
Setting up table: admin_line
Setting up table: place_point
Allocating memory for dense node cache
Allocating dense node cache in one big chunk
Allocating memory for sparse node cache
Sharing dense sparse
Node-cache: cache=800MB, maxblocks=12800*65536, allocation method=3

Reading in file: /home/pnorman/osm/osm2pgsql/tests/liechtenstein-2013-08-03.osm.pbf
Using PBF parser.
  parse time: 0s
Node stats: total(65733), max(65733) in 0s
Way stats: total(7121), max(7121) in 0s
Relation stats: total(113), max(113) in 0s
Committing transaction for admin_area
node cache: stored: 65733(100.00%), storage efficiency: 89.16% (dense blocks: 9, sparse nodes: 0), hit rate: 100.00%
Osm2pgsql failed due to ERROR: result COPY_END for admin_line failed: ERROR:  Geometry type (MultiLineString) does not match column type (LineString)
CONTEXT:  COPY admin_line, line 1, column way: "0105000020110F0000010000000102000020110F000045020000AD5AF7DB5E5E3041601BDF590DB35641F5C905C4385E3041..."
```

Looking at it with gdb, I *think* it's entering `get_wkb_multiline` from https://github.com/openstreetmap/osm2pgsql/commit/e1b75a5775d410b4e66cdef76d20c335b43706b3?diff=unified#diff-84169822394671b5c318a60145ba54dfR22, while before it used to call build_both

lonvia:
fixes #726

### DIFF
--- a/osmium-builder.cpp
+++ b/osmium-builder.cpp
@@ -54,11 +54,11 @@ osmium_builder_t::get_wkb_node(osmium::Location const &loc) const
 }
 
 osmium_builder_t::wkbs_t
-osmium_builder_t::get_wkb_line(osmium::WayNodeList const &nodes, bool do_split)
+osmium_builder_t::get_wkb_line(osmium::WayNodeList const &nodes, double split_at)
 {
     wkbs_t ret;
 
-    double const split_at = m_proj->target_latlon() ? 1 : 100 * 1000;
+    bool do_split = split_at > 0.0;
 
     double dist = 0;
     osmium::geom::Coordinates prev_pt;
@@ -167,7 +167,7 @@ osmium_builder_t::get_wkb_multipolygon(osmium::Relation const &rel,
 
 osmium_builder_t::wkbs_t
 osmium_builder_t::get_wkb_multiline(osmium::memory::Buffer const &ways,
-                                    bool do_split)
+                                    double split_at)
 {
     // make a list of all endpoints
     using endpoint_t = std::tuple<osmium::object_id_type, size_t, bool>;
@@ -258,7 +258,7 @@ osmium_builder_t::get_wkb_multiline(osmium::memory::Buffer const &ways,
         // found a line end, create the wkbs
         m_buffer.commit();
         auto linewkbs =
-            get_wkb_line(m_buffer.get<osmium::WayNodeList>(0), do_split);
+            get_wkb_line(m_buffer.get<osmium::WayNodeList>(0), split_at);
         std::move(linewkbs.begin(), linewkbs.end(),
                   std::inserter(ret, ret.end()));
     }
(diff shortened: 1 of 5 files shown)


osmium-builder.cpp
```diff
@@ -54,11 +54,11 @@ osmium_builder_t::get_wkb_node(osmium::Location const &loc) const
 }
 
 osmium_builder_t::wkbs_t
-osmium_builder_t::get_wkb_line(osmium::WayNodeList const &nodes, bool do_split)
+osmium_builder_t::get_wkb_line(osmium::WayNodeList const &nodes, double split_at)
 {
     wkbs_t ret;
 
-    double const split_at = m_proj->target_latlon() ? 1 : 100 * 1000;
+    bool do_split = split_at > 0.0;
```
pnorman: Are we setting the stage for any floating point comparison issues later on?

processor-line.cpp
```diff
@@ -18,6 +18,5 @@ geometry_processor::wkbs_t
 processor_line::process_relation(osmium::Relation const &,
                                  osmium::memory::Buffer const &ways)
 {
-    // XXX are multilines really acceptable?
-    return m_builder.get_wkb_multiline(ways, false);
+    return m_builder.get_wkb_multiline(ways, 1000000);
```
pnorman: Given the size of North American road relations and the Mercator scaling, it's safer to use 10 000 000.

osmium-builder.cpp
lonvia: I would expect that comparison against zero is reasonably safe. I was considering using NaN but it did feel like a misusing the value. The other option is to add do_split as an additional parameter, which would be fine to me as well.

processor-line.cpp
lonvia: This was the value used before the regression from the libosmium change. I can add another 0 though.

lonvia:
The Appveyor error is odd given that I haven't changed anything in the CMake file. More likely a configuration change on the Appveyor side.

pnorman:
Looks good to me then, but I want to test it.

lonvia:
Rebased and appveyor is happy now. I'll leave it as is then and for you to merge after you have tested your stuff.

pnorman:
I get a segfault with this, but it's not the fault of any code here. Instead, the import progresses farther and reveals a bug that occurs elsewhere. The patch I used to test master for the segfault was

```patch
diff --git a/table.cpp b/table.cpp
index 87ccf7d..152075c 100644
--- a/table.cpp
+++ b/table.cpp
@@ -137,7 +137,7 @@ void table_t::start()
             sql += "\"tags\" hstore,";
         }

-        sql += (fmt("way geometry(%1%,%2%) )") % type % srid).str();
+        sql += (fmt("way geometry(Geometry,%1%) )") % srid).str();

         // The final tables are created with CREATE TABLE AS ... SELECT * FROM ...
         // This means that they won't get this autovacuum setting, so it doesn't
```

PR #761, opened 2017-05-28
Extracting middle sources to the separate folder

gardster:
Part of the #738 PR.
Extracts all middle related files to the separate folder. I believe that this will help to simplify understanding the project structure, because all middle related sources will be in one separate directory. Maybe, as a next step, it will be possible to avoid a prefix "middle" in the filenames.


pnorman:
I'd rather look at all of the file organization at once. Do we want sub-folders? If so, what are they?

gardster:
You have already introduce implicit subfolders when using names with prefixes like `output- *` `middle- *` and others. And you will have more files of such type during refactoring or introducing additional functionality (like jsonb support). I suggest to use subfolders instead of prefixes.

PR #767, opened 2017-06-21
Add RPM spec for osm2pgsql

gongled:
Hello,

Thank you for your project. Would be nice to have RPM spec to let your users install `osm2pgsql` utility from prebuilt binaries for RHEL/CentOS systems. I hope someday it will will help somebody.

### DIFF
--- a/packaging/rpm/osm2pgsql.spec
+++ b/packaging/rpm/osm2pgsql.spec
@@ -1,0 +1,69 @@
+###############################################################################
+
+Summary:              OpenStreetMap data to PostgreSQL converter
+Name:                 osm2pgsql
+Version:              0.90.2
+Release:              0%{?dist}
+License:              GPL
+Group:                Development/Tools
+URL:                  http://wiki.openstreetmap.org/wiki/Osm2pgsql
+
+Source:               https://github.com/openstreetmap/%{name}/archive/%{version}.tar.gz
+
+BuildRoot:            %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
+
+BuildRequires:        cmake gcc-c++ boost-devel expat-devel zlib-devel bzip2-devel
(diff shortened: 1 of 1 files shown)


pnorman:
Distro-specific information should go to repos maintained by the distros as there is no way for the maintainers here to keep this information current.

It's also not clear how we'd maintain this with no test coverage, and with the maintainers on Debian, no way to test it.

See also #679, #491.

woodpeck:
I would like to mildly disagree. While the above ("can't ensure correctness, can't test" - and in one of the referenced tickets "why this distro and not others") is certainly correct from a puritanical viewpoint, I think that even an unsupported set of packaging instructions right here in this repo would make things easier for anyone who doesn't have the comfort of osm2pgsql already being included in their favourite distribution. Yes, it runs the risk of being untested and out of date, but the alternative is one or several one-off spec file rotting away somewhere completely out of view. Would it really hurt the maintainer pride to allow a directory "packaging/name_of_distribution" that clearly says it's unmaintained but at least serves as the go-to point for people looking for distribution specific build scripts/rules/recipes or whatever? I think it would be an improvement. (Says the author of https://github.com/openstreetmap/osm2pgsql/commit/d55e19114983df1c09eb455dbd3b11b43bb90466)

lonvia:
As far as I'm aware, there are well maintained RPMs of the latest osm2pgsql releases in Fedora. And we have a very active Debian package maintainer as well. Let's leave issues like packaging to people who actually understand something about it. I really wouldn't want to get in their way by adding half-baked, unmaintained scripts here.

pnorman:
In my view, if we accept something into this repo, we're committing to maintain it. Another factor in my consideration is that it's never people from the distros asking for this in this repo.

When a distribution (e.g. Homebrew) hasn't had people active on their side keeping their packaging up to date, we've helped them out with what they needed. I prefer to work that way.

PR #770, opened 2017-07-13
Edit English, regularize spacing, add comments

RichMorin:
In order to make the file a bit more readable, I edited some of the English text and regularized the spacing of the columns.  I then added comments indicating the relevant wiki page.

### DIFF
--- a/default.style
+++ b/default.style
@@ -1,46 +1,48 @@
+# default.style
+#
 # This is the default osm2pgsql .style file that comes with osm2pgsql.
 #
-# A .style file has 4 columns that define how OSM objects end up in tables in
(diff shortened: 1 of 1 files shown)


default.style
```diff
@@ -1,232 +1,240 @@
+# default.style
```
pnorman: I don't see a need for the filename here

RichMorin:
Most of the tags have wiki pages of the form ".../Key:<tag>", so it could be argued that the comments aren't all that useful.  And, for many cases where there is no such page, one could be created.  However, I find the comments useful as a quick reference and will retain them in my copy.

I like having a filename in the header because I find it handy when looking at a printed copy, etc.

PR #771, opened 2017-07-16
second attempt at file clean-up

RichMorin:
This version removes the inline comments for wiki pages, replacing them by a block comment and a small number of one-line comments.

### DIFF
--- a/default.style
+++ b/default.style
@@ -1,46 +1,46 @@
 # This is the default osm2pgsql .style file that comes with osm2pgsql.
 #
-# A .style file has 4 columns that define how OSM objects end up in tables in
-# the database and what columns are created. It interacts with the command-line
-# hstore options.
+# A .style file has four columns that define how OSM objects end up in tables
+# in the database and what columns are created.  The file interacts with the
+# command-line hstore options.
 #
 # Columns
 # =======
 #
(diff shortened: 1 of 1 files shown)


default.style
```diff
@@ -1,232 +1,279 @@
...
+#   access          http://wiki.openstreetmap.org/wiki/Key:access
+#   addr:housename  http://wiki.openstreetmap.org/wiki/Key:addr
+#   note:*          http://wiki.openstreetmap.org/wiki/Key:note
+#   tower:type      http://wiki.openstreetmap.org/wiki/Key:tower:type
+#
+# A few tags (e.g., abandoned, disused) are treated as lifecycle prefixes.
+# Their wiki pages contain a terminating colon (:).  For example:
+#
+#   abandoned       http://wiki.openstreetmap.org/wiki/Key:abandoned:
+#   disused         http://wiki.openstreetmap.org/wiki/Key:disused:
```
lonvia: I'm not convinced that the style file is the right place to explain the OSM tagging system. A one liner like "For an see the OSM wiki at https://wiki.openstreetmap.org." should be sufficient and would be less distracting.

pnorman:
Closing as abandoned.

PR #781, opened 2017-08-12
Fix counter initialisation in wkb_reader

### LINKED ISSUE #766: Duplicate rows with --expire-tiles option
I'm trying to find out which tiles are expired each time I apply osm day diff to my planet db.
I niticed that, when I use osm2pgsql with the _--expire-tiles_ option, the count of rows on my 4 planet tables is greater than the count of rows on the same tables obtained without using the _--expire-tiles_ option: It seems like it duplicates modified rows, leaving the old ones in the db.
Here is the command I run:

`osm2pgsql -d planetdb -s -C 4096 --number-processes 8 --hstore-all --append --expire-tiles 9-19 --expire-output /tmp/expired_tiles_list`

In addition to applying the changes, I would expect such an option only to return a output file, with respect to the "normal" execution. Instead, executing osm2pgsql in two different ways (with/without the --expire-tiles option), it leaves the db in two different states.
How could this be possible? Am I doing something wrong?

lonvia:
The counter in wkb_reader was always initialized with 0 (`result` being nullptr after having just been moved to `m_result`). So expiry from_db() has not been doing anything for a long time now.

As a test, import a single way and then append a change which modifies a node on this way.

@Nakaner Would you mind running this change through your setup? It is more extensive then my tests when it comes to expiry.

Fixes #766.

### DIFF
--- a/table.hpp
+++ b/table.hpp
@@ -58,7 +58,7 @@ class table_t
                 }
             private:
                 wkb_reader(pg_result_t &&result)
-                : m_result(std::move(result)), m_count(PQntuples(result.get())),
+                : m_result(std::move(result)), m_count(PQntuples(m_result.get())),
                   m_current(0)
                 {}
 


lonvia:
This is a candidate for backporting.

PR #784, opened 2017-08-30
Update Fedora build instructions

stevendlander:
* Adding additional libraries needed to build to Fedora install command
* Changing `yum` to `dnf`
* Adding specific Fedora versions using GEOS 3.5.1 by default in the package manager
* Adding a note as to why Fedora 26 has incompatible GEOS library, linking issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,12 +65,17 @@ sudo apt-get install make cmake g++ libboost-dev libboost-system-dev \
 On a Fedora system, use
 
 ```sh
-sudo yum install cmake gcc-c++ boost-devel expat-devel zlib-devel bzip2-devel \
-  postgresql-devel proj-devel proj-epsg lua-devel
+sudo dnf install cmake make gcc-c++ boost-devel expat-devel zlib-devel \
+  bzip2-devel postgresql-devel proj-devel proj-epsg lua-devel
 ```
 
 On RedHat / CentOS first run `sudo yum install epel-release` then install
-dependencies like on Fedora.
+dependencies with:
+
+```sh
+sudo yum install cmake make gcc-c++ boost-devel expat-devel zlib-devel \
+  bzip2-devel postgresql-devel proj-devel proj-epsg lua-devel
+```
 
 On a FreeBSD system, use
 


README.md
```diff
@@ -62,12 +62,13 @@ sudo apt-get install make cmake g++ libboost-dev libboost-system-dev \
...
-On a Fedora system, use
+On a Fedora 24/25 system, use
 
 ```sh
-sudo yum install cmake gcc-c++ boost-devel expat-devel zlib-devel bzip2-devel \
-  postgresql-devel proj-devel proj-epsg lua-devel
+sudo dnf install cmake make gcc-c++ boost-devel expat-devel zlib-devel bzip2-devel \
+  postgresql-devel proj-devel proj-epsg lua-devel geos-devel
 ```
+_Note: Fedora 26 uses GEOS 3.6.1 which is currently incompatible.  See [Issue 634](https://github.com/openstreetmap/osm2pgsql/issues/634)_
```
lonvia: GEOS is no longer a requirement, so this comment is not needed. Neither is the added geos-devel dependency.

README.md
```diff
@@ -62,12 +62,13 @@ sudo apt-get install make cmake g++ libboost-dev libboost-system-dev \
...
 ```
 
-On a Fedora system, use
+On a Fedora 24/25 system, use
 
 ```sh
-sudo yum install cmake gcc-c++ boost-devel expat-devel zlib-devel bzip2-devel \
-  postgresql-devel proj-devel proj-epsg lua-devel
+sudo dnf install cmake make gcc-c++ boost-devel expat-devel zlib-devel bzip2-devel \
+  postgresql-devel proj-devel proj-epsg lua-devel geos-devel
```
lonvia: Is dnf also used in the current stable CentOS release? If yes, you should be changing the EPFL installation instructions below, if not then the CentOS section below needs explicit installation instructions using yum.

README.md
stevendlander: I built today and can report that the current build instructions are not sufficient to get a working build on Fedora 24/25.   `cmake ..` will fail saying that GEOS is required.  Does this mean the make files need to be updated to remove the GEOS requirement?

I tested on Docker containers `fedora:24` and `fedora:25`.

lonvia: Did you try build master or a release?

README.md
stevendlander: Checked centos:7, dnf is *not* installed by default.

README.md
stevendlander: This was on `0.92.0` branch

stevendlander: Master works as expected, I will remove the GEOS note and devel library

stevendlander:
Do I need to squash the commits?

lonvia:
Squashing is not mandatory, it's up to you if you what you want to have in the git history.

Do you mind also adding the yum installation line to the CentOS section? Experience tells me there will be questions otherwise.

stevendlander:
Updated and squashed.

lonvia:
Thank you!

PR #794, opened 2017-10-23
Allow importing features whose name is missing.

edgarsi:
Sometimes features with no names are useful to import into the "places" table.


lonvia:
Thanks for your contribution and I agree that this would be useful to have but I'd like to see a more general solution to the problem where we have style file for gazetteer that makes all tags configurable, see also https://github.com/openstreetmap/Nominatim/issues/938. So, I'm sorry to have to reject this PR.

edgarsi:
It's fine, an imperfect solution works for my in-house needs. Hope you find the time for the better solution though!

PR #795, opened 2017-10-29
Extend mercator projection to 89.99 latitude

### LINKED ISSUE #793: Clipping in mercator conversion
The internal osm2pgsql function for conversion for mercator has a hard clipping at 85.07, see https://github.com/openstreetmap/osm2pgsql/blob/master/reprojection.cpp#L23. This is the reason why there are so many invalid polygons near the poles (see #787). We are currently using the exact algorithm for reprojection which degenerates only very close to the pole and PostGIS is quite happy with geometries that are outside the defined area for the projection. Therefore it should be safe to increase the clipping to something much closer to 90 degrees. This way we should be able to obtain valid geometries for most of the OSM objects close to the pole.

We could also switch to libosmium's projection function. It uses a much faster algorithm in the well-defined ranges and degenerates to the exact algorithm when closer to 90 degrees. However, this should be done on top of our choice of clipping. Libosmium itself does not do any clipping at all.

CC @imagico, @joto

lonvia:
Also switches to libosmium's projection function which uses a faster approximation for lower latitudes.

Fixes #793.

When importing Antarctica, the number of valid polygons increases from 64954 on master to 66949 in this change and the number of notices from postgresql wrt self-intersecting polygons decreases from many to 0. I have not checked the outcome visually.

### DIFF
--- a/reprojection.cpp
+++ b/reprojection.cpp
@@ -11,6 +11,8 @@
 #include <cstdlib>
 #include <cstring>
 
+#include <osmium/geom/mercator_projection.hpp>
+
 #include "reprojection.hpp"
 
 /** must match expire.tiles.c */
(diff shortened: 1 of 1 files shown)


lonvia:
On a full planet import there are now 14 self-intersecting polygons left. I still need to check visually if the results are okay.

lonvia:
Visual check out ok, too.

PR #800, opened 2017-11-10
Simplify Windows installation instructions

zstadler:


lonvia:
These instructions seem to globally install osm2pgsql in the Windows system directories. Are these steps really necessary? I was under the impression that the osm2pgsql binary can simply be started as long as the libraries in the same directory.

zstadler:
[Perfect is the enemy of good](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good)

The existing instructions did not work for me. I'm suggesting instructions that did work for me. If anyone has simpler instructions that work, let's have them in the README.

This patch is waiting for one of those with write permission for two months. In the meantime, a conflict was created that requires a resolution and the broken Windows instructions are still there.

pnorman:
We don't currently have Windows install instructions, just a link to the artifacts.

>  I was under the impression that the osm2pgsql binary can simply be started as long as the libraries in the same directory.

It can.

Installing is more complicated than just unzipping the artifact because there are assumptions about default locations for the style file.

I'm not sure what the answer here is, but don't know that the documentation changes here are a good way to do it.

Also, it removes the content about where to get the source if you want to install from it.

martin-s:
Installing into program files is not required, as it maybe forces you to upgrade to an administrator account. Just put the binaries and dependencies into one folder and execute it. This should be sufficient.

lonvia:
@zstadler Could you elaborate why just downloading und unpacking the zip archive and then starting the binary did not work for you? Maybe we can improve the build process to make it work.

@pnorman OSM2PGSQL_DATADIR is indeed looks to be completely broken for Windows. We should fix that. I would suggest to just use the current directory.

zstadler:
Apparently, I created `C:\libs\share` in order to avoid
```
Osm2pgsql failed due to ERROR: Couldn't open style file 'C:/libs/share/osm2pgsql/default.style': No such file or directory
```
It has nothing to do with DLLs or PATH.

Administrator permissions are needed only one chooses to install `osm2pgsql` in on of the `Program Files` directories.

README.md
```diff
@@ -18,10 +18,18 @@ Nominatim, or general analysis.
...
 
-Most Linux distributions include osm2pgsql. It is also available on macOS with [Homebrew](http://brew.sh/).
+**Linux**: Most distributions include osm2pgsql.
 
-Unoffical builds for Windows are available from [AppVeyor](https://ci.appveyor.com/project/openstreetmap/osm2pgsql/history) but you need to find the right build artifacts. The latest
-release is [0.92.0](https://ci.appveyor.com/api/projects/openstreetmap/osm2pgsql/artifacts/osm2pgsql_Release.zip?tag=0.92.0).
+**macOS**: osm2pgsql is available with [Homebrew](http://brew.sh/).
+
+**Windows**: Unoffical builds are available from [AppVeyor](https://ci.appveyor.com/project/openstreetmap/osm2pgsql/history). The latest
+release is [0.94.0](https://ci.appveyor.com/api/projects/openstreetmap/osm2pgsql/artifacts/osm2pgsql_Release.zip?tag=0.94.0). Just unzip the archive into a directory of your choice. You may need Admnistrator permissions in order to unzip into `C:\Program Files (x86)`.
```
Nakaner: The sentence "You may need Admnistrator […]" is not necessary. That's obvious. Users who don't know these very basics of permission management on Windows are often also unable to use the command line and therefore unable to use Osm2pgsql.

README.md
```diff
@@ -18,10 +18,18 @@ Nominatim, or general analysis.
...
+
+_Note:_ `osm2pgsql` expects the default style file at `C:/libs/share/osm2pgsql/default.style`. In order to avoid an error message use <code>--style <em>&lt;osm2pgsql directory&gt;</em>\default.style</code> or run the following commands once:
+* `mkdir C:\libs\share`
+* <code>mklink /D C:\libs\share\osm2pgsql "<em>&lt;osm2pgsql directory&gt;</em>"</code>
```
Nakaner: I think that this is too detailed. Users who know what they are doing, don't need this instruction. All other users should just stick to adding `--style FILE` to every invocation of Osm2pgsql.

README.md
zstadler: Windows is not the primary target of the osm2pgs distribution. For example, it does not have an installer as other Windows software packages have. It would be nice to help Windows users in overcoming that hurdle.

README.md
zstadler: The fact is that the use of `--style FILE` is not needed on Linux. As far as I could see, it isn't part of all the documented examples.
Even users that "know what they're doing" on Windows may be unaware of the assumptions taken by the authors of the README.

Nakaner: It is need on Linux if you don't install Osm2pgsql but just build it and execute the resulting binary.

lonvia: osm2pgsql now looks in the current directory for the default.style on Windows. This should solve the issue. At least in theory. Would be nice if somebody could test this.

lonvia:
The latest release has changed the handling of the style file. It is now expected in the current directory. This will hopefully improve the situation a little bit on Windows and addresses the main concern of this PR. Closing here as the documentation has now diverged too much. Thank you for rising this issue.

Issue #801, opened 2017-11-18
Using LuaJIT to speed up lua code


mmd-osm:
LuaJIT is a powerful drop in replacement to run Lua code in native mode with JIT optimizations: see http://luajit.org/luajit.html for details. To evaluate the potential impact on osm2pgsql runtimes, I did some very quick & dirty prototyping. Primary goal was to have a working osm2pgsql , and being able to do some initial run time comparison between the current and a LuaJIT enabled version.

Test was using the latest openstreetmap-carto style on a very small country extract (30MB PBF). Platform: Core i7 and Ubuntu 16.04, 64bit.

`osm2pgsql -G --hstore --style openstreetmap-carto.style --tag-transform-script openstreetmap-carto.lua -d gis  extract.osm.pbf`

Before (Lua)
`Processing: Node(3045k 1015.2k/s) Way(556k 46.37k/s) Relation(1790 1790.00/s)  parse time: 16s`

After (LuaJIT)
`Processing: Node(3045k 3045.5k/s) Way(556k 139.12k/s) Relation(3410 3410.00/s)  parse time: 6s`

First conclusion: LuaJIT allows for some significant savings in the osm2pgsql parsing phase. I'd recommend to further evaluate this option and see how it might be integrated into osm2pgsql.

## Code

see pull request

NB: I do realize that OSRM removed LuaJIT support last year (see: Rip out LuaJIT for good #2955), due to lack of usage in their (Mapbox) own  setup, unclear impact wrt speed improvement and users reporting crashes. At least to some extent this seems to be related to the their build system, and automatic fallback to LuaJIT. That's why I'd propose `WITH_LUAJIT` should be disabled by default.

lonvia:
This is definitely worth exploring further. Do you mind making a proper PR? Then I would do some performance tests. We do have the advantage over OSRM that we don't use Luabind which brings a whole lot of additional pain. So I suspect that having LuaJIT as a drop-in replacement is less complicated for us.

mmd-osm:
The quality of this patch is still not good enough for a PR, but since you asked for one for testing purposes... :smile: 

Disabling WITH_LUAJIT settings via ccmake doesn't really work once it has been enabled once. I'm pretty sure there's some way to fix these caching issues, but I didn't investigate this any further yet. So for the time being, you would have to use a new build directory for testing:

1. Start with new clean `build` directory
2. Run `cmake -D WITH_LUAJIT=ON ..`
[...]

PR #810, opened 2018-01-10
Added LuaJIT support

mmd-osm:
See #801 for discussion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,8 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 set(DATA_DIR \"${CMAKE_INSTALL_PREFIX}/share/osm2pgsql\")
 
 option(BUILD_TESTS "Build test suite" OFF)
-option(WITH_LUA    "Build with lua support" ON)
+option(WITH_LUA    "Build with Lua support" ON)
+option(WITH_LUAJIT "Build with LuaJIT support" OFF)
 
 if (NOT TESTING_TIMEOUT)
   set(TESTING_TIMEOUT 1200)
@@ -85,8 +86,14 @@ find_package(Osmium REQUIRED COMPONENTS io proj)
 include_directories(SYSTEM ${OSMIUM_INCLUDE_DIRS})
 
 if (WITH_LUA)
-  find_package(Lua REQUIRED)
-  include_directories(${LUA_INCLUDE_DIR})
+  if (WITH_LUAJIT)
+    find_package(LuaJIT REQUIRED)
+    include_directories(${LUAJIT_INCLUDE_DIR})
+    set(HAVE_LUAJIT 1)
+  else()
+    find_package(Lua REQUIRED)
+    include_directories(${LUA_INCLUDE_DIR})
+  endif()
   set(HAVE_LUA 1)
 endif()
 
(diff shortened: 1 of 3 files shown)


pnorman:
Are there really no changes to the C++ code needed?

mmd-osm:
Yes, that's correct. OSRM used something similar, here's their respective patch to remove luajit again: https://github.com/Project-OSRM/osrm-backend/commit/eafe3fa928ce5bdade7f124d4d96357d23e4f18e

CMakeLists.txt
```diff
@@ -11,7 +11,8 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 set(DATA_DIR \"${CMAKE_INSTALL_PREFIX}/share/osm2pgsql\")
 
 option(BUILD_TESTS "Build test suite" OFF)
-option(WITH_LUA    "Build with lua support" ON)
+option(WITH_LUA    "Build with Lua support" ON)
+option(WITH_LUAJIT "Build with LuaJIT support" OFF)
```
StyXman: How should these be used? `WITH_LUAJIT` implies `WITH_LUA` or not? Or should they be mutually exclusive?

mmd-osm: Yes, see line 88 + 89 below, `WITH_LUAJIT` implies `WITH_LUA`

lonvia: Well, my reading below is that you need to have both set WITH_LUA and WITH_LUAJIT. That's okay for me though.

cmake/FindLuaJIT.cmake
```diff
@@ -0,0 +1,63 @@
+# Locate LuaJIT library
+# This module defines
+#  LUAJIT_FOUND, if false, do not try to link to Lua
+#  LUA_LIBRARIES
+#  LUA_INCLUDE_DIR, where to find lua.h
+#  LUAJIT_VERSION_STRING, the version of Lua found (since CMake 2.8.8)
+
+## Copied from default CMake FindLua51.cmake
```
lonvia: Any chance that can be merged with FindLua.cmake? It looks partially duplicated.

mmd-osm: I'm not fully convinced that this is really a good idea.

Both lua and luajit are independent packages with their own lifecycle. Once a new lua version arrives, FindLua.cmake can be simply replaced by a new stock version (I think Lua projects provides their own FindLua.cmake, but I don't know if this is used here).

Although they partially look similar, they reference different include files and libraries in different places. I fell merging FindLuaJIT into FindLua might complicate the latter too much.

lonvia: That's a fair point. In this case the FindLuaJIT script should be at least consistent about the variables it exports. Either make them all LUAJIT_* or LUA_*. The latter would keep the rest of the cmake file simpler.

Other than that, the PR looks alright to me. I haven't tested on other OSses, though.

mmd-osm: Thank you for your feedback. I have changed all variable names to LUAJIT_* now, as I was a bit worried to introduce naming clashes otherwise. That might be an issue in particular for variable LUA(_JIT)_FOUND, which derived from the package name "Lua" or "LuaJIT" respectively.

mmd-osm:
Some Travis CI test results: As expected, the MacOS build doesn't work due to missing libraries. Otherwise, results look ok:

https://travis-ci.org/mmd-osm/osm2pgsql/builds/363811377?utm_source=github_status&utm_medium=notification

Not sure if we want to run this as part of the official .travis.yml file by adding some luajit specific entries in the build matrix?

https://github.com/mmd-osm/osm2pgsql/commit/5c10629f54fc8c50feccf98f2ff7f64e9a9b324f

lonvia:
Yes, please add the travis stuff. It's okay to ignore mac for now.

mmd-osm:
Travis stuff now added in https://github.com/openstreetmap/osm2pgsql/pull/810/commits/fa956df485a50901863ffb49654daee00f16a03c ...

mmd-osm:
LuaJIT improves parsing phase runtime by about 50%. Unfortunately, we haven't really added any guidance in the README file on how to enable it. As a consequence people still run full planet imports using the slower Lua 5.2. Should we do something about this?

PR #811, opened 2018-01-10
Use HTTPS for planet links in man page

pnorman:


pnorman:
cc @woodpeck is http or https the preferred form for geofabrik download links?

PR #815, opened 2018-02-03
Skip output for untagged objects

### LINKED ISSUE #706: Skip transforms for untagged objects
From ClearTables/ClearTables#109

A full featured multi-transform can easily contain 10+ point tables. Reasonably written transforms will exit in tens or hundreds of ns, but the cost of calling 10 transforms adds up for 4b nodes. 97% of nodes have no tags and will be filtered out by any transform I can think of.

Skipping these untagged nodes in C++ would be faster than calling to Lua and skipping them. Although the performance is mainly an issue with the multi-backend, I propose also doing it for the pgsql backend with lua transforms. The C backend already skips untagged nodes under all possible style files, but it's worth checking that it skips it early in the transform to avoid unnecessary work.

lonvia:
When adding objects call output backends only when there
are any tags. Only works for the initial import. When updating
data, all objects need to be forwarded in case they need to
be deleted. That also means that output backends still must
be able to handle untagged objects.

Fixes #706.

### DIFF
--- a/osmdata.cpp
+++ b/osmdata.cpp
@@ -41,9 +41,13 @@ int osmdata_t::node_add(osmium::Node const &node)
     mid->nodes_set(node);
 
     int status = 0;
-    for (auto &out : outs) {
-        status |= out->node_add(node);
+
+    if (!node.tags().empty()) {
+        for (auto &out : outs) {
+            status |= out->node_add(node);
+        }
     }
+
     return status;
 }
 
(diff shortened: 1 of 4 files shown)


pnorman:
> Import time for node is approx. half with this change and there is no significant difference for ways and relations (as excepted). Given that node processing makes up a rather small part of the processing time, the overall gain is in the order of 1%

This probably depends significantly on the overall speed of the machine, on faster machines nodes take up more time relative to ways and relations.

PR #819, opened 2018-02-11
add short option for flat-node file

lonvia:
This option is now so frequently used that it deserves a
short option and to be listed more prominently in the help.
Also adapted the size.

### DIFF
--- a/options.cpp
+++ b/options.cpp
@@ -16,7 +16,7 @@
 
 namespace
 {
-    const char * short_options = "ab:cd:KhlmMp:suvU:WH:P:i:IE:C:S:e:o:O:xkjGz:r:V";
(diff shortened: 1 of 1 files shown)


options.cpp
```diff
@@ -60,7 +60,7 @@ namespace
         {"number-processes", 1, 0, 205},
         {"drop", 0, 0, 206},
         {"unlogged", 0, 0, 207},
-        {"flat-nodes",1,0,209},
+        {"flat-nodes",1,0, 'F'},
```
pnorman: Is this the formatting clang-format wants?

lonvia: With regard to the space in front of the option: yes. However, if I run clang-format over that particular change it will reformat the entire array (the indention is wrong) and I'd rather leave that as is.

Issue #812, opened 2018-01-12
Issue in windows 7 64bit


alirdn:
I've tried to use the latest version binary built artifacts from appveyor (also mentioned 0.94.0 link in Readme), and run on windows 7 64bit. But I've got an error about CreateFile2 not located in KERNEL32.dll. But latest 0.92.0 binary works like a charm on my OS. So I done some searches to find out what's the cause of the problem. Here are whats I found:

1. Latest working binary: [0.0.352](https://ci.appveyor.com/project/openstreetmap/osm2pgsql/build/0.0.352)
2. First successful built binary that doesn't works: [0.0.397](https://ci.appveyor.com/project/openstreetmap/osm2pgsql/build/0.0.397)
3. The built binary that doesn't works with existing commit (previous commit was deleted so cannot compare those commits): [0.0.402](https://ci.appveyor.com/project/openstreetmap/osm2pgsql/build/0.0.402)
3. Compare of those two commits: [Compare 18880484c5f4e9e7f11a51b4c2a36f601af05e5a and 6c6dcd152c6c1fddfc197b94de137a44ffcf7c1f](https://github.com/openstreetmap/osm2pgsql/compare/18880484c5f4e9e7f11a51b4c2a36f601af05e5a...6c6dcd152c6c1fddfc197b94de137a44ffcf7c1f)

Well found nothing by comparing those two (not familiar with the osm2pgsql project very well), So do some more searches and found that there is a [regression in zlib](https://github.com/madler/zlib/issues/49) that osm2pgsql uses, with the same behavior. The regression is in zlib 1.2.8 and fixed in 1.2.9. By reading the log files of appveyor builds, found **Found ZLIB: C:/libs/lib/zlib.lib (found version "1.2.8")** line, so osm2pgsql uses the same version with regression. So I think the problem came from zlib. I really needs osm2pgsql latest version that works on my OS, and above results are the all things I can do to find a fix for the problem.

alirdn:
It's been fixed by merged PR #820. Also by the PR, issue #803 fixed two. I'm closing this issue and it's OK to close  #803 issue.

Issue #803, opened 2017-12-20
osmpgsql for windows 32 bit


souradh:
Hi,

I downloaded osm2pgsql using this link :
 "https://ci.appveyor.com/api/buildjobs/iisabxffssilv7gl/artifacts/osm2pgsql_Release.zip"

After that I unzipped and tried to execute osm2pgsql.exe.
Then i got a message  :
"This version of C:\osm2pgsql\osm2pgsql-bin\osm2pgsql.exe is not compatible with
the version of Windows you're running. Check your computer's system information
to see whether you need a x86 (32-bit) or x64 (64-bit) version of the program, a
nd then contact the software publisher."

Mine is windows 7 32 bit.

Where can I download Osm2pgsql for windows 32 bit??

pnorman:
> Where can I download Osm2pgsql for windows 32 bit??

Appveyor just builds one version. You'll have to compile it yourself and you might need to build your own dependencies.

lonvia:
Appveyor could build a 32 bit version but we don't have the libraries for it and it was quite a bit of work to create them. Before doing that I'd like to make sure that the 32bit on Linux is still running. We don't test that at all at the moment.

lonvia:
With thanks to @alirdn, we now have a 32bit pre-built binary as well. You can download the latest build for the master branch at https://ci.appveyor.com/api/projects/openstreetmap/osm2pgsql/artifacts/osm2pgsql_Release_x86.zip?job=Environment%3A%20arch%3Dx86&branch=master (release builds will only appear starting from the next release).

I do have to add a word of warning though. 32bit versions are not really actively maintained anymore. They still build but beyond that, they are not particularly well tested.

PR #820, opened 2018-02-11
Appveyor enhancement

alirdn:
The windows builds available as appveyor artifacts has some issue. So some enhancement added to fix these issues. The main reason for these changes, is issue #812. Because of old zlib (1.2.8) version used for builds, the created binary is not running on windows 7. This pull request fix issue #812 by changing the zlib from 1.2.8 to 1.2.11 (latest). Also other libraries are updated. Finally,  windows 32 bit version also will be built (fixes issue #803).

List of changes:
1. Update libraries:
    - expat: 2.2.0 -> 2.2.5
    - lua: 5.1.5 -> 5.3.4
    - zlib: 1.2.8 -> 1.2.11
    - wingetopt (Build both x86 and x64 by myself using [https://github.com/alex85k/wingetopt](https://github.com/alex85k/wingetopt))
2. Build matrix for both x86 and x64 builds.

PostgreSQL 9.4.7 lib used to build. But PostgreSQL 9.6 used to run tests.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,8 +14,8 @@ environment:
     BOOST_PATH: C:\Libraries\boost_1_63_0
     POSTGRESQL_VER: 9.6.6
     TESTS_POSTGRESQL_ROOT: C:\Progra~1\PostgreSQL\9.6
-    POSTGIS_VER: 2.4.3
-    POSTGIS_URL: "https://winnie.postgis.net/download/windows/pg96/buildbot/postgis-pg96-binaries-2.4.3w64gcc48.zip"
+    POSTGIS_VER: 2.4.2
+    POSTGIS_URL: "http://download.osgeo.org/postgis/windows/pg96/postgis-bundle-pg96-2.4.2x64.zip"
     PGUSER: postgres
     PGPASSWORD: Password12!
   matrix:
@@ -44,7 +44,7 @@ clone_depth: 1
 init:
   - git config --global core.autocrlf input
   - '"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall" %vcvarsall_arg%'
-
+  - ps: iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
 install:
   - cd c:\
   - set PATH=%TESTS_POSTGRESQL_ROOt%\bin;%PATH%;%conda_path%
@@ -56,7 +56,7 @@ install:
   - 7z x -y C:\lua_%platform%.zip -oC:\lua
   - ps: if (!(Test-Path "C:\postgis.zip")){(new-object net.webclient).DownloadFile($env:POSTGIS_URL, "C:\postgis.zip")}
   - 7z x C:\postgis.zip -oC:\postgis
-  - xcopy /e /y /q C:\postgis\postgis-pg96-binaries-2.4.3w64gcc48 %TESTS_POSTGRESQL_ROOT%
+  - xcopy /e /y /q C:\postgis\postgis-bundle-pg96-2.4.2x64 %TESTS_POSTGRESQL_ROOT%
   - git clone https://github.com/alex85k/wingetopt -b %WINGETOPT_VER% C:\wingetopt
   - python -V
   - pip install psycopg2
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,160 +1,115 @@
 environment:
   global:
-    DOWNLOADS_PATH: c:\downloads
-    LIBS_PATH: c:\libs
+    BZIP2_VER: 1.0.6
+    EXPAT_VER: 2.2.5
+    PROJ4_VER: 4.9.3
+    ZLIB_VER: 1.2.11
+    LUA_VER: 5.3.4
+    LUA_INCLUDE_DIR: C:\lua\include
+    LUA_LIBRARIES: C:\lua\lua53.lib
+    LUA_DLL: C:\lua\lua53.dll
+    WINGETOPT_VER: v0.95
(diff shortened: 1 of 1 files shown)


lonvia:
Very nice, thanks.

Could you elaborate about the odd Postgres setup? Is there a particular reason why not to build against Postgers 9.6 as well? Would we run into issues with backwards compatibility?

Do you think wingetopt is small enough that we could simply build it as part of the appveyor script as well? It would be nice to have all dependencies available via automatic builds.

alirdn:
> Could you elaborate about the odd Postgres setup? Is there a particular reason why not to build against Postgers 9.6 as well? Would we run into issues with backwards compatibility?

Well, during my tries to make a successful build, I've used Postgres 9.6 64bit (without running the unit tests), and when used the .exe to run an import, an error showed up (can't remember the error now, will try to build again to reproduce error), so decided to use 9.4.7 for building. I just want to update appveyor config for a more mature auto builds, as a starting point, so can make a pull request and discuss about other things here. About the backwards compatibility, as far as I know, there is no issue. But I'll search to get a 100% trust-able answer.

> Do you think wingetopt is small enough that we could simply build it as part of the appveyor script as well? 
Yes. It's tiny. Just 1 cpp file. And also cmake compatible.

> It would be nice to have all dependencies available via automatic builds.
You mean that we build every single one during the build process? Well need to consider more things I think. 
List of dependencies that needed:
- expat
- proj
- bzip2
- zlib
- Lua
- wingetopt
- PostgreSQL
- Boost libraries

Well Postgres and Boost are huge libraries. Also Boost is available in appveyor vms. So i think, we could create a dependencies zip file from dependencies except Postgres and Boost, for x86 and x64, host it somewhere, and download and use it during build process (All the files needed to build for both arch, except Postgres and Boost, are zipped and saved as an artifact). By using this approach, updating the dependencies become a trivial task (Just change the content of zip file without even touching the cmake or appveyor config files).

lonvia:
> You mean that we build every single one during the build process? 

No, just for the winopt library. The issue for the osm2pgsql maintainers is that none of us has a windows installation ready. So it would be good if we can get all dependencies either via some package manager (like Anaconda, NuGet or appveyor) or be able to build it directly via appveyor. Then there is a chance that the script remains up to date.

pnorman:
> No, just for the winopt library. The issue for the osm2pgsql maintainers is that none of us has a windows installation ready. So it would be good if we can get all dependencies either via some package manager (like Anaconda, NuGet or appveyor) or be able to build it directly via appveyor. Then there is a chance that the script remains up to date.

My desktop is Windows, but I don't to any Windows development. So although I can test stuff if necessary, I have no clue what to do about Windows' lack of a sane well-established package management system.

alirdn:
> No, just for the winopt library. The issue for the osm2pgsql maintainers is that none of us has a windows installation ready. So it would be good if we can get all dependencies either via some package manager (like Anaconda, NuGet or appveyor) or be able to build it directly via appveyor. Then there is a chance that the script remains up to date.

I've updated the script, so now most of the dependencies will be downloaded from conda using miniconda. But wingetopt, lua and postgis is not available on any windows package manager, so lua and postgis will be downloaded from other sources and wingetopt will be built using github repo.

Version of the dependencies is hard-coded. So we could have control over them.

Also PostgreSQL is updated to 9.6.6 both for build and test. That issue I've mentioned before, was because of wrong dll arch I've used. I've copied 64bit dll for 32bit version.

Also, the "libintl-8.dll" (x64) & "intl.dll" (x86) is omitted from final release. As far as I know, no need to include it.

pnorman:
> But wingetopt, lua and postgis is not available on any windows package manager

But you've got the PostgreSQL client headers from conda, right? Since PostGIS isn't used when building osm2pgsql, only when testing, I'm not too worried there.

Lua is from LuaBinaries, which is official, so I'm not worried there either. It's also probably more active than the NuGet Lua project.

alirdn:
> But you've got the PostgreSQL client headers from conda, right?

Yes.

> Since PostGIS isn't used when building osm2pgsql, only when testing, I'm not too worried there.

Yes, PostGIS is just for testing.

> Lua is from LuaBinaries, which is official, so I'm not worried there either. It's also probably more active than the NuGet Lua project.

Beside the fact that luaBinaries is more active and it's updated to the last lua version, the NuGet lua project is dll only so could not be used during build process.

lonvia:
I second @pnorman. The sources look good now. This will make it indefinitely easier to keep the windows binaries up-to-date in the future. Thanks a lot.

I'll merge this now and would appreciate if Windows users could try out the new binaries.

lonvia:
So this PR turns out to have an unexpected side effect: permalinks to the download artefacts have become rather ugly. The reason is that we now have two different jobs for 32 and 64 bits and to get to the artefact, the job name needs to be appended. Unfortunately the job name cannot be set to something reasonable. Instead appveyor uses a string that is concatenated from the entries in the matrix parameter. So for the 32bit zip file we get: https://ci.appveyor.com/api/projects/openstreetmap/osm2pgsql/artifacts/osm2pgsql_Release_x86.zip?job=Environment%3A%20platform%3Dx86%2C%20vcvarsall_arg%3Dx86%2C%20conda_path%3DC%3A\Miniconda36\Scripts%2C%20conda_library_path%3DC%3A\Miniconda36\envs\osm2pgsql\Library%2C%20lua_url%3Dhttps%3A%2F%2Flonvia.dev.openstreetmap.org%2Fosm2pgsql-winbuild%2Flua_x86.zip%2C%20build_type%3DRelease

I've tried to remove all variables except `platform` to get the link to a manageable size but failed miserably to make the windows shell do what I want. @alirdn would you mind having a look at that? It would be good if we end up with a simple:

    -matrix:
        - platform: x86
        - platform: x64

in the appveyor config. My failed attempts to do that can be found in the appveyor-test branch.

PR #832, opened 2018-03-21
Fix parsing of command line option "-e 12" (no expiry maxzoom given)

### LINKED ISSUE #830: --expire-tiles: Parameter parsing does not work
After upgrading from Debian Stretch (0.92) to Buster (0.94) our update script stopped working. I have tried many styles of passing the `--expire-tiles` parameter with no luck.
The code parsing this parameter was changed significantly in 24488bfdaec2c432e33144ecb2688387253db37c, breaking existing functionality. The command `osm2pgsql -d gis -H localhost -U gisuser -S ~/osm2pgsql/default.style -P 5432 -s -C 800 -a diffs.osc.gz -e15 -o expire.list`, suggested by the OSM Wiki, no longer works.    

What I have tried so far:
  - `-e19`
  - `-e0-19`
  - `-e 19`
  - `-e "19"`
  - `--expire-tiles 19`
  - `--expire-tiles 0-19`
  - `--expire-tiles "19"`
  - `--expire-tiles "0-19"`
  - `--expire-tiles "just work pls"`

When supplying only the max zoom level, I get "_ERROR: Minimum and maximum zoom level for tile expiry must be separated by '-'._" Supplying both min and max yields "_ERROR: Missing zoom level for tile expiry._" Negative zoom levels actually get caught and throw an error.

Is this a user error on my side or is this option actually broken?

Nakaner:
Changes:

* add support to parse `-e 12` (no maximum zoom level given)
* better error messages
* more unit tests for parsing the argument of `-e`

This fixes #830.

### DIFF
--- a/options.cpp
+++ b/options.cpp
@@ -380,14 +380,14 @@ options_t::options_t(int argc, char *argv[]): options_t()
             break;
         case 'e':
             if (!optarg || optarg[0] == '-') {
-                throw std::runtime_error("Missing argument for option -e. Zoom "
+                throw std::runtime_error("Missing argument for option --expire-tiles. Zoom "
                                          "levels must be positive.\n");
             }
             char *next_char;
             expire_tiles_zoom_min =
                 static_cast<uint32_t>(std::strtoul(optarg, &next_char, 10));
             if (expire_tiles_zoom_min == 0) {
-                throw std::runtime_error("Missing zoom level for tile expiry. "
+                throw std::runtime_error("Bad argument for option --expire-tiles. "
                                          "Minimum zoom level must be larger "
                                          "than 0.\n");
             }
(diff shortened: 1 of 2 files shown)


options.cpp
```diff
@@ -385,8 +387,9 @@ options_t::options_t(int argc, char *argv[]): options_t()
             expire_tiles_zoom_min =
                 static_cast<uint32_t>(std::strtoul(optarg, &next_char, 10));
             if (expire_tiles_zoom_min == 0) {
-                throw std::runtime_error(
-                    "Missing zoom level for tile expiry.\n");
+                throw std::runtime_error("Missing zoom level for tile expiry. "
+                                         "Minimum zoom level must be larger "
+                                         "than 0.\n");
```
lonvia: I'd remove the 'Missing zoom level' part entirely or rename it to 'Bad argument for...'. As is, it seems misleading given that we have checked a couple of lines above that there is an optarg.

PR #845, opened 2018-04-22
Lua function errors now trigger runtime exception

### LINKED ISSUE #664: Give a better error when a transform returns a non-string type
I had a multi backend transform which returned tags `{foo=true}`, and it reports the error

```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
```

This error doesn't give any clue that the problem is in the transform

mmd-osm:
I don't know exactly why errors in Lua processing are only sent to stderr and processing continues. There could be some nasty issues overflowing the stack and causing other types of corruption.

I have changed those messages to runtime errors now. Maybe there are some reasons why this doesn't make sense, in which case I'm happy to revert that part of this PR.

Also, I added some error handling for lua_tostring with details about incorrect data type. Instead of throwing some strange exception as in #664, processing now aborts with the following message.

_Osm2pgsql failed due to ERROR: Basic tag processing returned NULL value. Possibly this is due an incorrect data type 'boolean'._

Another example:

```lua
function filter_tags_generic(tags)
    local foo = 0

    local bar = {}

    bar[123] = function()
      return 123
    end

    if foo == 0 then
      return 0, bar
    end
```

Error message: _Osm2pgsql failed due to ERROR: Basic tag processing returned NULL value. Possibly this is due an incorrect data type 'function'._

Fixes #664

### DIFF
--- a/tagtransform-lua.cpp
+++ b/tagtransform-lua.cpp
@@ -83,11 +83,12 @@ bool lua_tagtransform_t::filter_tags(osmium::OSMObject const &o, int *polygon,
     lua_pushinteger(L, sz);
 
     if (lua_pcall(L, 2, (o.type() == osmium::item_type::way) ? 4 : 2, 0)) {
-        fprintf(stderr,
-                "Failed to execute lua function for basic tag processing: %s\n",
-                lua_tostring(L, -1));
         /* lua function failed */
-        return 1;
+        throw std::runtime_error(
+            (boost::format(
+                 "Failed to execute lua function for basic tag processing: %1%") %
+             lua_tostring(L, -1))
+                .str());
     }
 
     if (o.type() == osmium::item_type::way) {
@@ -105,6 +106,22 @@ bool lua_tagtransform_t::filter_tags(osmium::OSMObject const &o, int *polygon,
     while (lua_next(L, -2) != 0) {
         const char *key = lua_tostring(L, -2);
         const char *value = lua_tostring(L, -1);
+        if (key == NULL) {
+           int ltype = lua_type(L, -2);
+           throw std::runtime_error(
+               (boost::format(
+                    "Basic tag processing returned NULL key. Possibly this is due an incorrect data type '%1%'.") %
+                lua_typename(L, ltype))
+                   .str());
+        }
+        if (value == NULL) {
+           int ltype = lua_type(L, -1);
+           throw std::runtime_error(
+               (boost::format(
+                    "Basic tag processing returned NULL value. Possibly this is due an incorrect data type '%1%'.") %
+                lua_typename(L, ltype))
+                   .str());
+        }
         out_tags.emplace_back(key, value);
         lua_pop(L, 1);
     }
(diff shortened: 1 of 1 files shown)


mmd-osm:
> Any objections to changing to early fail here?

It not only changing to early fail, it really a question of failing or not failing at all. The current implementation silently discards _any_ errors raising in Lua processing. If you don't check your stderr job logs manually, you will never find out that a particular transformation might have an issue.

mmd-osm:
As seen on https://help.openstreetmap.org/questions/64259/osm2pgsql-failed-due-to-error-failed-to-execute-lua-function:

@SomeoneElseOSM seems to be the first one to hit this stricter check: the previous z_order calculation logic had some issue ("attempt to perform arithmetic on field 'layer' (a string value)")  and the error message on osm2pgsql was easy to ignore. With the stricter check, the error message is no longer ignored, and the stylesheet needs to be fixed to use proper Lua syntax.
 
Code in question: https://github.com/SomeoneElseOSM/SomeoneElse-style/blob/master/style.lua#L16

pnorman:
It's good to know that it's working - that code was always in error.

```
5*"4"
stdin:1: unexpected symbol near '5'
```

mmd-osm:
Well, the code didn't really check, if the field contains a number. In many cases that's not really an issue if the value can be implicitly converted to a number. It fails though, when people use multiple values, like in `layer = 4;3`, or put in some other non-numeric value in there.

```lua
> print(5*"4")
20
> print(tonumber("4"))
4

```

```lua
> print(5*"4;3")
stdin:1: attempt to perform arithmetic on a string value
stack traceback:
	stdin:1: in main chunk
	[C]: in ?
> print(tonumber("4;3"))
nil

```

SomeoneElseOSM:
@mmd-osm pull requests welcome :)

mmd-osm:
@SomeoneElseOSM : just copy & paste from upstream osm carto, it's already implemented properly there.

SomeoneElseOSM:
Oddly, I'm not able to reproduce https://github.com/SomeoneElseOSM/SomeoneElse-style/issues/5 locally with the latest version of osm2pgsql (with the "Failed to execute lua function for basic tag processing" text in it) and the dodgy style.lua with "z_order = 10*keyvalues["layer"]" in it.  I don't see the reported "Failed to execute" errors, even when I load data that has "layer=yes" in it (Oxfordshire, which contains https://www.openstreetmap.org/way/68346732).

Obviously I could correct the style.lua, but it'd be nice to know why the error doesn't occur for me first!  Any suggestions as to what to look at?  It's not some sort of alternate type of lua processing that I need to enable is it?  Currently I've just got a Kubuntu 18.04 laptop and followed https://wiki.openstreetmap.org/wiki/User:SomeoneElse/Ubuntu_1804_tileserver_load#Install_osm2pgsql on it.

mmd-osm:
I think there's plenty of other logic in your style, some of which would return before reaching the `add_z_order ` function. As the error has been reported on bangladesh-latest.osm.pbf, I'd give this 153MB download a try first (http://download.geofabrik.de/asia/bangladesh.html).

SomeoneElseOSM:
@mmd-osm thanks - that worked (Bangladesh showed the issue before fixing style.lua and didn't afterwards).

Issue #847, opened 2018-05-06
osm2pgsql fails to import if index names are already used


Komzpa:
osm2pgsql fails to import if index names are already used:
```
Osm2pgsql failed due to ERROR: CREATE INDEX highway_osm_line_index ON highway_osm_line USING GIST (way) WITH (FILLFACTOR=100)  failed: ERROR:  relation "highway_osm_line_index" already exists
```

Komzpa:
done in #848

PR #848, opened 2018-05-06
Let postgres resolve non conflicting index names itself

Komzpa:
Names are hardcoded now, but that leads to #847 in unfortunate event when tables were renamed but their respective indexes still have previous name, that you need to import once again.

### DIFF
--- a/table.cpp
+++ b/table.cpp
@@ -152,7 +152,7 @@ void table_t::start()
 
         //slim mode needs this to be able to delete from tables in pending
         if (slim && !drop_temp) {
-            sql = (fmt("CREATE INDEX %1%_pkey ON %1% USING BTREE (osm_id)") % name).str();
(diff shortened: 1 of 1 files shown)


mmd-osm:
I wonder a bit why you didn’t just do a `drop index if exists` first before creating the index to avoid creating multiple possibly redundant indices.

Komzpa:
@mmd-osm index is supposed to be dropped on `drop table` earlier. But it's not, since it's an index for another table. Dropping index for some other table uninvolved in import isn't a good idea.

mmd-osm:
Ah, ok. I see now. The other index was a left over from a table rename operation.

lonvia:
The only issue I could see with this is when people rely on the index name when they do post-processing or that they run into odd name clashes when adding additional indexes. If that is a real problem for somebody, now is the time to speak up, otherwise I'm inclined to merge this.

lonvia:
The indexes are called *_idx instead of *_index.

PR #851, opened 2018-06-07
[HOLD] : extending Travis with Docker + Database test 

ImreSamu:
this is a little experiment for a `docker-compose` based integration test.
*  local matrix test 
*  travis test

Based on official [gcc:x  docker images](https://hub.docker.com/r/library/gcc)  
and  `mdillon/postgis:10-alpine`  image.

work in progress


pnorman:
For DB testing we should use Travis' builtin PostgreSQL support, not docker.

ImreSamu:
>For DB testing we should use Travis' builtin PostgreSQL support, not docker.

thanks for the feedback.
I have separated the patch -  
And the first part is here:   #852

And we can discuss later,  that "full local test"  with docker -  is important or not ..

Sometimes it can be usefull  ..

PR #862, opened 2018-08-10
Remove support for old-style MPs

lonvia:
Also removes support for superseding ways from polygons. The lua bindings still expect that the supersede table is present but will simply drop it.

Still todo:
 - [x] recheck that really all old-style MP code is gone
 - [x] adapt example Lua style
 - [x] fix regression tests 
  
I've done some performance tests as this PR moves all way processing to the mono-threaded first computation stage. Test were done on a Hetzner EX41-SSD (i7-7700, 32GB RAM, SSD in Raid0) with command:

    ./osm2pgsql -d gis --drop -C 10 -s -S ../default.style --flat-nodes gis.flatnodes planet-180625.osm.pbf

The minimal cache is a bit of a worst case scenario as all node lookups go to the slightly slower flatnode file.

master:
```
Node stats: total(4581403065), max(5716309124) in 425s
Way stats: total(504094523), max(601431841) in 49995s
Relation stats: total(5899892), max(8406934) in 26409s

Going over pending ways...
        357968593 ways are pending

Using 8 helper-processes
Finished processing 357968593 ways in 28948 s

Osm2pgsql took 125689s overall
```

This PR:
```
Node stats: total(4581403065), max(5716309124) in 471s
Way stats: total(504094523), max(601431841) in 71760s
Relation stats: total(5899892), max(8406934) in 25839s

Going over pending ways...
        0 ways are pending

Using 8 helper-processes
Finished processing 0 ways in 0 s

Osm2pgsql took 116951s overall
```

So this looks fairly promising and it would be ok to go ahead with dropping old-style MPs without improving the first-stage processing first.


mmd-osm:
If way processing isn't totally IO bound, there's chances that the single threaded initial processing phase would benefit even more from luajit than it did in the past.

lonvia:
I had forgotten the part where tags are copied from ways to relations. That is gone now, too. I think that covers all old-style MP code.

@pnorman, it would help if you could look over the tagtransform code, if the handling is still right

lonvia:
There is no more pending step during import but it is still needed for updates, where changes on nodes/way need to be propagated upwards to ways/relations.

pnorman:
bumping this for myself - haven't gotten to look at it with holidays, forest fires, job hunting, and other stuff.

bagage:
@pnorman Did you have the time to look at this? You seem *really* busy but it'd be awesome to have this I think (when I see how much ways need too be processed for France only... it's killing my server!).

lonvia:
This has been lying around for a while now. I have some more refactoring changes pending and like to get this in to avoid merge conflicts. If there are no further objections by the end of the week, I'll just get it merged.

PR #865, opened 2018-08-28
Display progress while processing pending relations/ways

bagage:
This step is one of the biggest step with no logging, it looks like osm2pgsql is stuck but actually it is processing these pending items. Now it'll display the number of pending items left to process instead of being silent :).

### DIFF
--- a/osmdata.cpp
+++ b/osmdata.cpp
@@ -217,6 +217,20 @@ struct pending_threaded_processor : public middle_t::pending_processor {
         }
     }
 
+    static void print_stats(pending_queue_t &queue, std::mutex &mutex)
+    {
+        size_t queue_size;
+        do {
+            mutex.lock();
+            queue_size = queue.size();
+            mutex.unlock();
+
+            fprintf(stderr, "\rLeft to process: %zu...", queue_size);
+
+            std::this_thread::sleep_for(std::chrono::seconds(1));
+        } while (queue_size > 0);
+    }
+
     //starts up count threads and works on the queue
     pending_threaded_processor(std::shared_ptr<middle_query_t> mid,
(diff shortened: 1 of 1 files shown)


der-stefan:
Oh yes, this is a great feature - it can take up to days without showing progress...

bagage:
Yeah, it looks like it's related to #862 but in the mean time it seems that even if import won't need it anymore in the future, this still might be useful for updates, so I hope it'll be merged asap, wdyt @mmd-osm @lonvia @pnorman?

mmd-osm:
`print_stats` seems to be busy looping, maybe you want to use something like `std::this_thread::sleep_for(60s);` instead?

Also checking the queue status probably needs some additional locking via `mutex.lock();` + `mutex.unlock();` like done in `do_jobs`.

bagage:
Thanks for reviewing this @mmd-osm, I've updated the PR accordingly!

mmd-osm:
I didn't test your code, but this looks a bit strange. `queue.size()` is still not protected by a mutex, and I assume the queue is _not_ empty most of the time. Did you mean `while (!isEmpty)` perhaps?

bagage:
Indeed I refactored the do/while logic right before pushing, I should have tested it. This time I tested it and it should be OK (actually removed the `isEmpty()` code, to reduce code).

bagage:
Hey @lonvia, I did not run clang-format in the first place because it causes a lot of space diffs on the file, so I didn't want to commit these all. Maybe I applied it wrongly? 

Also, `git-clang-format` (`clang-format` actually because I could not find the former) is not fixing variables name. I don't know if it's supposed to, but I fixed it manually.

PR #905, opened 2019-01-30
Add support for --extra-attributes to Gazetteer output

Nakaner:
`--extra-attributes` is now supported by Gazetteer output. This pull requests writes all metadata fields to as to the `extratags` column.

If an OSM object lacks a metadata field (empty string or 0 as value) which is/will be common with usual OSM planet dumps and extracts, the field will not be written to the output.

Users can control whether they want all metadata fields or only specific ones. If the Nominatim style file (the new one, see #1277) contains a catch-all rule at the end, that catch-all rule will apply on all available metadata fields and they will be written to the `places` table. If one or many of the metadata fields are mentioned in an `extra` rule and no catch-all rule is present, only that field will be written.

**Example 1: write all tags and all metadata fields to the `extratags` column**

Append this to your style file:

```json
{
    "keys" : [""],
    "values" : {
        "" : "extra"
    }
}
```

**Example 2: write all metadata fields to the `extratags` column**

Append this to your style file:

```json
{
    "keys" : ["osm_version", "osm_timestamp", "osm_user", "osm_uid", "osm_changeset"],
    "values" : {
        "" : "extra"
    }
}
```

**Example 3: write timestamps to the `extratags` column**

Append this to your style file:

```json
{
    "keys" : ["osm_timestamp"],
    "values" : {
        "" : "extra"
    }
}
```

### DIFF
--- a/gazetteer-style.cpp
+++ b/gazetteer-style.cpp
@@ -128,8 +128,7 @@ gazetteer_style_t::flag_t gazetteer_style_t::parse_flags(std::string const &str)
     return out;
 }
 
-bool gazetteer_style_t::add_metadata_style_entry(std::string const &key,
-                                                 std::string const &value)
+bool gazetteer_style_t::add_metadata_style_entry(std::string const &key)
 {
     if (key == "osm_version") {
         m_metadata_fields.set_version(true);
@@ -190,10 +189,15 @@ void gazetteer_style_t::add_style_entry(std::string const &key,
         }
     }
 
-    if (add_metadata_style_entry(key, value)) {
+    if (add_metadata_style_entry(key)) {
         if (!value.empty()) {
             throw std::runtime_error("Style error. Rules for OSM metadata "
                                      "attributes must have an empty value.\n");
+        }
+        if (flags != SF_EXTRA) {
+            throw std::runtime_error("Style error. Rules for OSM metadata "
+                                     "attributes must have the style flag "
+                                     "\"extra\" and no other flag.\n");
         }
         return;
     }
--- a/gazetteer-style.cpp
+++ b/gazetteer-style.cpp
@@ -42,7 +42,6 @@ void gazetteer_style_t::clear()
     m_main.clear();
     m_names.clear();
     m_extra.clear();
-    m_metadata.clear();
     m_address.clear();
     m_operator = nullptr;
     m_admin_level = MAX_ADMINLEVEL;
@@ -71,11 +70,6 @@ void gazetteer_style_t::load_style(std::string const &filename)
             }
         }
     }
-}
-
-void gazetteer_style_t::set_metadata(const bool enabled)
-{
-    m_metadata_enabled = enabled;
 }
 
 gazetteer_style_t::flag_t gazetteer_style_t::parse_flags(std::string const &str)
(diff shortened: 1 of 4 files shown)


gazetteer-style.cpp
```diff
@@ -502,6 +547,13 @@ bool gazetteer_style_t::copy_out_maintag(pmaintag_t const &tag,
             escape_array_record(entry.second, buffer);
             buffer += "\",";
         }
+        for (auto const &entry : m_metadata) {
+            buffer += "\"";
+            escape_array_record(entry.first.c_str(), buffer);
```
lonvia: You can save the escaping here. All entry.first are well-known strings.

gazetteer-style.hpp
```diff
@@ -12,6 +12,7 @@ class gazetteer_style_t
 {
     using flag_t = uint16_t;
     using ptag_t = std::pair<char const *, char const *>;
+    using ptag_str_t = std::pair<const std::string, const std::string>;
```
lonvia: 'char const *' will do for the first parameter. It is only used with constant strings.

gazetteer-style.hpp
```diff
@@ -64,11 +65,17 @@ class gazetteer_style_t
...
+
 private:
     void add_style_entry(std::string const &key, std::string const &value,
                          flag_t flags);
     flag_t parse_flags(std::string const &str);
     flag_t find_flag(char const *k, char const *v) const;
+    void add_metadata_field(const std::string&& field, const std::string&& value);
+
+    template <typename T>
+    void add_metadata_field_num(const std::string&& field, const T value);
```
lonvia: Please use the const placement style as in the rest of the file.

gazetteer-style.cpp
```diff
@@ -384,6 +390,45 @@ void gazetteer_style_t::process_tags(osmium::OSMObject const &o)
...
+            add_metadata_field_num<osmium::changeset_id_type>("osm_changeset", o.changeset());
+        }
+
+        if (o.timestamp()) {
+            add_metadata_field("osm_timestamp", std::move(o.timestamp().to_iso()));
+        }
+    }
+}
+
+void gazetteer_style_t::add_metadata_field(const std::string&& field, const std::string&& value) {
```
lonvia: Please run clang-format over your change (see CONTRIBUTING.md).

gazetteer-style.cpp
```diff
@@ -384,6 +390,45 @@ void gazetteer_style_t::process_tags(osmium::OSMObject const &o)
...
+    // would require an allocation on heap and a cleanup at the end.
+    flag_t flag = find_flag(field.c_str(), value.c_str());
+    if (flag & SF_EXTRA) {
+        m_metadata.emplace_back(std::move(field), std::move(value));
+    }
+}
+
+template <typename T>
+void gazetteer_style_t::add_metadata_field_num(const std::string&& field, const T value) {
+    // This method is not linked to from outside this class. Therefore, it can stay in the source file.
```
lonvia: I'm not sure, I'm comfortable with that. Better move it to the header.

gazetteer-style.cpp
```diff
@@ -163,6 +163,25 @@ gazetteer_style_t::flag_t gazetteer_style_t::parse_flags(std::string const &str)
     return out;
 }
 
+bool gazetteer_style_t::add_metadata_style_entry(std::string const &key,
+                                                 std::string const &value)
```
lonvia: 'value' is unused here. You can just remove the second parameter. (I wonder why travis does not complain, we need to change that.)

Nakaner:
I added a check to ensure that only the "extra" flag is set.

I rebased on current master and adapted my code to your recent changes implemented in #914.

PR #906, opened 2019-02-09
Fix various unused parameter issues and enable -Wextra on travis

lonvia:
This fixes all -Wextra warnings. Mostly removes unused parameters and a couple of useless asserts. It also removes the randomized parameter test because it contained some bad parameter usage and it was impossible to figure out what the actual intention of the test was.


lonvia:
Appveyor is just confused that is has built the PR twice and errored out on artefact generation. That's safe to ignore.

PR #907, opened 2019-02-10
Some cleanup for output-pgsql members

lonvia:
This PR cleans up two things:

* Convert tables member to std::array. The size of the array is known at compile time and does not change, so a vector is not necessary here.
* export_list is moved to the transform classes. The style file is only used by output-pgsql to create the columns. During the actual import it is only handed to the transforms on every filter call. Moving the export_list to transforms makes the filter calls more more readable and also gives us the freedom to change the style description format independently for the transforms.

Have a look at the individual commit for each change.

### DIFF
--- a/output-multi.cpp
+++ b/output-multi.cpp
@@ -13,13 +13,12 @@
 
 output_multi_t::output_multi_t(std::string const &name,
                                std::shared_ptr<geometry_processor> processor_,
-                               export_list const &export_list_,
+                               export_list const &export_list,
                                std::shared_ptr<middle_query_t> const &mid,
                                options_t const &options)
 : output_t(mid, options),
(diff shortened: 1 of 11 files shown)


pnorman:
Failing travis with
```
/home/travis/build/openstreetmap/osm2pgsql/tagtransform-c.cpp:212:6: error: unused parameter ‘members’ [-Werror=unused-parameter]
 bool c_tagtransform_t::filter_rel_member_tags(
      ^
/home/travis/build/openstreetmap/osm2pgsql/tagtransform-c.cpp:212:6: error: unused parameter ‘member_roles’ [-Werror=unused-parameter]
```

lonvia:
> Failing travis with

Ah good, the new warning check is working. ;) Fixed.

PR #914, opened 2019-02-23
Introduce separate COPY threads

lonvia:
This PR reorganizes the way PostgreSQL tables are filled. So far we had a mixture of COPY and INSERT to add and update data. This PR introduces a common class for COPYing all data into the database and gets rid of INSERT entirely.

Data that should be inserted is collected in copy buffers which are COPYed into the database by a separate thread. To make this work with updates, the copy buffer also saves rows to be deleted. When such rows are available, the copy thread shortly stops the COPY, issues a single DELETE statement deleting all pending rows and then continues COPYing the buffered data. If the copy buffers are large enough, this should still be more efficient than the current DELETE/INSERT approach with prepared statements.

Copy threads can be used for multiple tables at the same time. When they detect that the next pending buffer is for a different table then it stops the ongoing copy and starts a new one for the new table.

The current implementation starts two copy threads, one for middle and one for all outputs. Each output table has its own db_copy_mgr, so that each table can queue data for copying independently. The processing step spawns one copy thread per processing thread serving the output tables. Middle is not written at this stage, so there is no copy thread needed for it.

This PR only provides a functional implementation. I haven't looked into optimising anything yet. In particular, the size of the copy buffers is currently fixed and we can now optimise the number of connections to the database for reading to reduce the number of DB connections to a maximum of 2 * parallel threads.


db-copy.hpp
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    /// Finish an array column.
+    void finish_array()
+    {
+        auto idx = m_current->buffer.size() - 1;
+        if (m_current->buffer[idx] == '{')
+            m_current->buffer += '}';
+        else
+            m_current->buffer[idx] = '}';
```
pnorman: A comment about how this is overwriting a `,` would be good here.

middle-pgsql.cpp
```diff
@@ -245,20 +156,34 @@ pg_result_t middle_query_pgsql_t::exec_prepared(char const *stmt,
     return exec_prepared(stmt, buffer);
 }
 
+pg_result_t middle_pgsql_t::exec_prepared(char const *stmt, osmid_t osm_id) const
+{
+    char buffer[64];
+    sprintf(buffer, "%" PRIdOSMID, osm_id);
```
pnorman: Opened #916 to cover the existing usage of `sprintf`

db-copy.hpp
lonvia: I've added documentation for all the functions in the copy manager. Hope that helps.

middle-pgsql.cpp
lonvia: Replaced all 4 occurrences I could find with snprintf.

pnorman:
With master and luajit
```
Processing: Node(5043406k 4137.3k/s) Way(563353k 40.39k/s) Relation(6601440 1005.55/s)  parse time: 21733s
Node stats: total(5043406477), max(6285198313) in 1219s
Way stats: total(563353404), max(671171896) in 13949s
Relation stats: total(6602164), max(9334518) in 6565s
Committing transaction for admin_area
Committing transaction for admin_line
Committing transaction for place_point
...
Mid: removing persistent node cache at nodes.bin
node cache: stored: 5043406477(100.00%), storage efficiency: 83.12% (dense blocks: 694886, sparse nodes: 187737901), hit rate: 100.00%
Osm2pgsql failed due to ERROR: Connection to database failed: FATAL:  sorry, too many clients already
```

with b89aea3 and no luajit
```
Processing: Node(5043406k 2688.4k/s) Way(563353k 25.76k/s) Relation(6601590 979.76/s)  parse time: 30484s
Node stats: total(5043406477), max(6285198313) in 1876s
Way stats: total(563353404), max(671171896) in 21870s
Relation stats: total(6602164), max(9334518) in 6738s
Mid: removing persistent node cache at nodes.bin
node cache: stored: 5043406477(100.00%), storage efficiency: 83.12% (dense blocks: 694886, sparse nodes: 187737901), hit rate: 100.00%
Osm2pgsql failed due to ERROR: Connection to database failed: FATAL:  sorry, too many clients already
```

I'm rerunning with luajit, I forgot to enable it when building the branch

pnorman:
It's 30 minutes faster under this PR
```
Processing: Node(5043406k 4292.3k/s) Way(563353k 42.88k/s) Relation(6601450 1143.90/s)  parse time: 20083s
Node stats: total(5043406477), max(6285198313) in 1175s
Way stats: total(563353404), max(671171896) in 13137s
Relation stats: total(6602164), max(9334518) in 5771s
```

PR #956, opened 2019-09-30
Avoid crashing on out-of-range inputs for int4 columns

### LINKED ISSUE #955: Large values cause import failure when using int4 columns
With the following style file:

```
# OsmType  Tag          DataType     Flags
node,way   population      int4      linear
```
and the following osm file:

```
<?xml version="1.0" encoding="UTF-8"?>
<osm version="0.6">
 <node id="1" visible="true" version="1" changeset="1" timestamp="2018-10-31T10:20:19Z" user="a" uid="1" lat="51.4779481" lon="-0.0014863">
   <tag k="population" v="4000000000" />
 </node>
</osm>
```
and the following command line
```
osm2pgsql -d osm2pgsql-test --style int4.style int4.xml
```
I get a failure of:
```
...snip...
Relation stats: total(0), max(0) in 0s
DB writer thread failed due to ERROR: result COPY_END for planet_osm_point failed: 
ERROR:  value "4000000000" is out of range for type integer
CONTEXT:  COPY planet_osm_point, line 1, column population: "4000000000"
```

I think it would be reasonable for osm2pgsql to just drop numbers that are outside of the valid int4 range, since it already drops anything that it can't interpret as a number from these columns.

gravitystorm:
Fixes #955

This approach also avoids undefined behaviour when using sscanf, and changes the mean calculation to avoid overflows.

As part of the development of this patch I made [a sample osm.xml file with a variety of numbers and ranges](https://gist.github.com/gravitystorm/409b5f651f013ac5a49ff71f84b82d37), to check that everything was working properly.

### DIFF
--- a/table.cpp
+++ b/table.cpp
@@ -457,12 +457,21 @@ void table_t::escape_type(const string &value, ColumnType flags)
     switch (flags) {
     case COLUMN_TYPE_INT: {
         // For integers we take the first number, or the average if it's a-b
-        long from, to;
-        int items = sscanf(value.c_str(), "%ld-%ld", &from, &to);
-        if (items == 1) {
+        int64_t from, to;
+        // limit number of digits parsed to avoid undefined behaviour in sscanf
+        int items = sscanf(value.c_str(), "%18ld-%18ld", &from, &to);
+        if (items == 1 && from <= std::numeric_limits<int32_t>::max() &&
+            from >= std::numeric_limits<int32_t>::min()) {
             m_copy.add_column(from);
         } else if (items == 2) {
-            m_copy.add_column((from + to) / 2);
+            // calculate mean while avoiding overflows
+            int64_t mean = (from / 2) + (to / 2) + ((from % 2 + to % 2) / 2);
+            if (mean <= std::numeric_limits<int32_t>::max() &&
+                mean >= std::numeric_limits<int32_t>::min()) {
+                m_copy.add_column(mean);
+            } else {
+                m_copy.add_null_column();
+            }
         } else {
             m_copy.add_null_column();
         }
--- a/table.cpp
+++ b/table.cpp
@@ -459,7 +459,7 @@ void table_t::escape_type(const string &value, ColumnType flags)
         // For integers we take the first number, or the average if it's a-b
         long long from, to;
         // limit number of digits parsed to avoid undefined behaviour in sscanf
-        int items = sscanf(value.c_str(), "%18Ld-%18Ld", &from, &to);
+        int items = sscanf(value.c_str(), "%18lld-%18lld", &from, &to);
         if (items == 1 && from <= std::numeric_limits<int32_t>::max() &&
             from >= std::numeric_limits<int32_t>::min()) {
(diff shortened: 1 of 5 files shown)


table.cpp
```diff
@@ -457,12 +457,21 @@ void table_t::escape_type(const string &value, ColumnType flags)
     switch (flags) {
     case COLUMN_TYPE_INT: {
         // For integers we take the first number, or the average if it's a-b
-        long from, to;
-        int items = sscanf(value.c_str(), "%ld-%ld", &from, &to);
-        if (items == 1) {
+        int64_t from, to;
+        // limit number of digits parsed to avoid undefined behaviour in sscanf
+        int items = sscanf(value.c_str(), "%18ld-%18ld", &from, &to);
```
lonvia: My reading of the sscanf man page is that you need to use '%18Ld' here to really get a 64bit value even on 32bit machines. You also need to use 'long long' then as data type for your variables to guarantee that everything matches. This should fix the Travis fail on OSX.

table.cpp
```diff
@@ -457,12 +457,21 @@ void table_t::escape_type(const string &value, ColumnType flags)
...
+        if (items == 1 && from <= std::numeric_limits<int32_t>::max() &&
+            from >= std::numeric_limits<int32_t>::min()) {
             m_copy.add_column(from);
         } else if (items == 2) {
-            m_copy.add_column((from + to) / 2);
+            // calculate mean while avoiding overflows
+            int64_t mean = (from / 2) + (to / 2) + ((from % 2 + to % 2) / 2);
```
mmd-osm: How about  `from + ( (to - from) / 2)`, assuming _to_  and _from_ are positive and _from <= to_?

table.cpp
gravitystorm: Great, I've taken that approach

gravitystorm:
> Nice tests. Do you mind adding them as an additional test to osm2pgsql directly with this PR? You can use `tests/test-output-pgsql-z_order.cpp` as an example.

I've added some tests too. I didn't see an existing way to test for null integer responses, so I used string comparison with an empty string instead.

Let's see how the CI gets on!

table.cpp
mmd-osm: Unfortunately, that's not supported in C++11:  `table.cpp:462:43: error: ISO C++11 does not support the ‘%Ld’ gnu_scanf format`

gravitystorm: OK, anyone got any suggestions of what integer type we can use and what format string to pair it with? It needs to parse *more than 32 bits* so that we can constrain it with a digit specifier and still get the full 32 bit output.

So far we've tried 
* `int64_t` with `%18ld` (macosx didn't like this)
* `long long` with `%18Ld` (no L in C++11)

lonvia: The recommendation (in the BUG section of scanf. Who reads that far down? ;) ) is '%18lld' (and `long long`). Can you try that? If it doesn't work either, then we have to go the `strtol` route. I can help out in that case.

gravitystorm: The CI is green with '%18lld' - thanks!

PR #960, opened 2019-10-14
limit the number of buffers that are in the copy queue

lonvia:
If PostgreSQL is much slower than the data processing, the queue can become very large and lead to out-of-memory errors.

### DIFF
--- a/db-copy.cpp
+++ b/db-copy.cpp
@@ -28,9 +28,18 @@ db_copy_thread_t::~db_copy_thread_t() { finish(); }
 void db_copy_thread_t::add_buffer(std::unique_ptr<db_cmd_t> &&buffer)
 {
     assert(m_worker.joinable()); // thread must not have been finished
-    std::unique_lock<std::mutex> lock(m_queue_mutex);
-    m_worker_queue.push_back(std::move(buffer));
-    m_queue_cond.notify_one();
+
+    for (;;) {
+        std::unique_lock<std::mutex> lock(m_queue_mutex);
+        if (m_worker_queue.size() >= db_cmd_copy_t::Max_buffers) {
+            m_queue_full_cond.wait(lock);
+            continue;
+        }
+
+        m_worker_queue.push_back(std::move(buffer));
+        m_queue_cond.notify_one();
+        break;
+    }
 }
 
 void db_copy_thread_t::sync_and_wait()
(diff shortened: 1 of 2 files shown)


lonvia:
Might have to do a new release after this one as it badly impacts imports without flatnode files.

mmd-osm:
> as it badly impacts imports without flatnode files.

could this be mitigated by increasing Max_buffers? Would it make sense to add a command line parameter for this setting?

lonvia:
I don't understand your comment. The issue here was that osm2pgsql was producing data for postgresql at a faster rate than postgresql can consume. The Max_buffers parameter has been explicitly introduced to limit the number of data buffers we queue for postgresql. There is not much point in making it larger as it only has to account for the jitter when producer and consumer are working at about the same speed.

Imports without flatnode files are hit especially hard because producing the data for `planet_osm_nodes` is blazingly fast. You pretty much always end up with the larger part of the node data in RAM or with an out-of-memory condition. There is a chance though to run into an OOM at other stages as well when postgresql is badly configured or the disks are very slow.

I'm against adding more obscure performance parameters. People are struggling with -C enough already.

mmd-osm:
Oh, I think I misinterpreted what you meant by your second comment. I was under the impression that introducing this limit somehow would have a negative impact on an import without flatnodes. 

As you said, it's quite the opposite, adding the limit is essential to avoid an OoM condition, and there's not much point increasing this value, let alone make it user configurable. Thank you for clarifying this.

db-copy.cpp
```diff
@@ -28,9 +28,18 @@ db_copy_thread_t::~db_copy_thread_t() { finish(); }
...
 {
     assert(m_worker.joinable()); // thread must not have been finished
-    std::unique_lock<std::mutex> lock(m_queue_mutex);
-    m_worker_queue.push_back(std::move(buffer));
-    m_queue_cond.notify_one();
+
+    for (;;) {
+        std::unique_lock<std::mutex> lock(m_queue_mutex);
+        if (m_worker_queue.size() >= db_cmd_copy_t::Max_buffers) {
+            m_queue_full_cond.wait(lock);
```
mmd-osm: just checked how this is done in _c++ concurrency in action_. They use a lambda function for the condition variable wait method: https://github.com/anthonywilliams/ccia_code_samples/blob/master/listings/listing_4.1.cpp#L47 (code example is not about a bounded queue, it's just to show how they use condition variables throughout chapter 4).

lonvia: Changed that for both condition variables.

PR #961, opened 2019-10-14
put different advise on cache size in sync

lonvia:
Recommend the size of the PBF file as this is agnostic to the database growth.


mmd-osm:
_size of the PBF file_ should maybe explicitly mention that the size in MB (or maybe MiB) is meant here, rather than e.g. bytes or GB.

PR #966, opened 2019-11-11
Use catch framework for testing

lonvia:
This PR adds [Catch2](https://github.com/catchorg/Catch2/) to the project and ports all tests to use it instead of hand-crafted asserts. Some of the tests are heavily refactored, in particular, the node store and middle tests. They should still cover the same ground as before. There are also new tests for the new db-copy classes and the gazetteer backend.


tests/test-middle.cpp
```diff
@@ -0,0 +1,235 @@
...
+        // other nodes are not retrievable
+        auto &n2 =
+            buffer.get<osmium::Way>(buffer.add_way(3, {1, 2, 1235})).nodes();
+        REQUIRE(mid_q->nodes_get_list(&(n2)) == 0);
+    }
+
+    SECTION("Set and retrieve a single way")
+    {
+        osmid_t way_id = 1;
+        double lat = 12.3456789;
```
joto: These can be `const`. Probably lots of other places in the code where `const` could be added.

tests/test-middle.cpp
```diff
@@ -0,0 +1,235 @@
...
+        REQUIRE(mid_q->ways_get(way_id, outbuf));
+
+        auto &way = outbuf.get<osmium::Way>(0);
+
+        CHECK(way.id() == way_id);
+        REQUIRE(way.nodes().size() == nds.size());
+
+        REQUIRE(mid_q->nodes_get_list(&(way.nodes())) == nds.size());
+        for (osmid_t i = 1; i <= 10; ++i) {
+            CHECK(way.nodes()[(size_t)i - 1].ref() == i);
```
joto: `static_cast<std::size_t>(i)`

lonvia:
Thanks for the review. I've addressed most of the comments except for a couple where we have to have general style decision first.

PR #970, opened 2019-11-13
Remove dependency on libgen.

joto:
Removes the dependency on libgen including some CMake stuff for the
price of writing a small function ourselves.


lonvia:
Please run through git-format.

joto:
Done.

PR #971, opened 2019-11-13
Use default member initialization in some classes.

joto:


options.hpp
```diff
@@ -45,56 +45,56 @@ struct options_t {
...
-    boost::optional<std::string> tblsmain_data; ///< Pg Tablespace to store main tables (no default TABLESPACE)
-    boost::optional<std::string> tblsslim_data; ///< Pg Tablespace to store slim tables (no default TABLESPACE)
-    std::string style; ///< style file to use
+    bool append = false; ///< Append to existing data
+    bool slim = false; ///< In slim mode
+    int cache = 800; ///< Memory usable for cache in MB
+    boost::optional<std::string> tblsmain_index{boost::none}; ///< Pg Tablespace to store indexes on main tables (no default TABLESPACE)
+    boost::optional<std::string> tblsslim_index{boost::none}; ///< Pg Tablespace to store indexes on slim tables (no default TABLESPACE)
+    boost::optional<std::string> tblsmain_data{boost::none}; ///< Pg Tablespace to store main tables (no default TABLESPACE)
+    boost::optional<std::string> tblsslim_data{boost::none}; ///< Pg Tablespace to store slim tables (no default TABLESPACE)
```
lonvia: Please also adhere to the 80 column limit while you are changing stuff here (i.e. move comments above the members).

options.cpp
```diff
@@ -278,30 +278,14 @@ std::string database_options_t::conninfo() const
...
-  expire_tiles_zoom(0), expire_tiles_zoom_min(0),
-  expire_tiles_max_bbox(20000.0), expire_tiles_filename("dirty_tiles"),
-  hstore_mode(HSTORE_NONE), enable_hstore_index(false), enable_multi(false),
-  hstore_columns(), keep_coastlines(false), parallel_indexing(true),
+: projection(reprojection::create_projection(PROJ_SPHERE_MERC)),
 #ifdef __amd64__
   alloc_chunkwise(ALLOC_SPARSE | ALLOC_DENSE),
 #else
   alloc_chunkwise(ALLOC_SPARSE),
 #endif
```
lonvia: Want to get rid of the #ifdef in the middle of the constructor as well?

options.hpp
joto: Fixed.

options.cpp
joto: Nope. There is another place in the code with the same 'ifdef' and I haven't fully understood what's going on there, so I'd rather leave this for now.

PR #973, opened 2019-11-14
Options cleanup

joto:
Cleanup of command line options list.
    
* Use nullptr where applicable
* Use no_argument and required_argument symbols instead of 0/1
* Sort by option name


lonvia:
Does this run cleanly through clang-format?

joto:
If you insist.

PR #981, opened 2019-11-18
Consistent naming and format for include guards in all hpp files

joto:


lonvia:
This one clashes with the other changes .Could you please rebase and force-push?

PR #982, opened 2019-11-19
Travis changes

joto:
* Added builds on Ubuntu bionic
* Added a build in release mode
* Updates various version outputs
* Some formatting cleanups


lonvia:
Can we remove the trusty version of gcc-8-pg96-luajit? There is no additional value against the new bionic version imo.

PR #983, opened 2019-11-19
Add travis, appveyor, and repology badges to README.md

joto:


mmd-osm:
I had issues with multiple branches, as Travis would report back the status of the latest run, irrespective of the branch. Better is to explicitly state the master branch name as parameter in the url: https://api.travis-ci.org/openstreetmap/osm2pgsql.svg?branch=master 

This link should be according to the following documentation: https://docs.travis-ci.com/user/status-images/

joto:
@mmd-osm You do realize you wrote this after the pull request was already merged? Most likely such a comment is going to be ignored now, but it is a good suggestion (also making the behaviour the same as with Appveyor). It would be much more helpful if you would open a new PR with that change. No PR is too small to be potentially helpful. And small changes like this are so easy to do in the Github UI, just edit the file and Github will lead you through the PR process.

mmd-osm:
Yes, I do realize I was a bit late to the party, yet I thought I'd better add a comment before I forget about it.

Issue #321, opened 2015-04-01
relative paths for multi-backend lua files


pnorman:
Currently if you set a tagtransform of `"tagtransform": "foo.lua"` it looks for foo.lua relative to your current working directory, not relative to the multi-backend style file or relative to osm2pgsql.

This means that it's impossible to have a style that works no matter where you call osm2pgsql from unless you use absolute paths.

I would prefer it to be relative to the style file.

lonvia:
This is the main blocker for making tests completely location-independent.

lonvia:
Fixed in #992.

PR #992, opened 2019-11-20
Look for lua script relative to style.json file.

joto:
Note that this is a breaking change. Users might have to change the file
names of their lua scripts in the style files.

See #321.


pnorman:
👍 in principle, haven't tested it.

PR #1004, opened 2019-11-25
Use fmt library instead of boost::format

joto:
The fmt library is included in contrib. Boost::format is not needed any more.

See #398

### DIFF
--- a/src/output-gazetteer.cpp
+++ b/src/output-gazetteer.cpp
@@ -1,6 +1,6 @@
-#include <boost/format.hpp>
 #include <libpq-fe.h>
 
+#include "format.hpp"
 #include "middle.hpp"
 #include "options.hpp"
 #include "osmtypes.hpp"
@@ -42,9 +42,9 @@ void output_gazetteer_t::delete_unused_classes(char const *osm_type,
     if (!clslist.empty()) {
         clslist[clslist.size() - 1] = '\0';
         // Delete places where classes have disappeared for this object.
-        m_conn->exec("DELETE FROM place WHERE osm_type = '%1%' "
-                     "AND osm_id = %2% and class = any(ARRAY[%3%])",
-                     osm_type, osm_id, clslist);
+        m_conn->exec("DELETE FROM place WHERE osm_type = '{}'"
+                     " AND osm_id = {} and class = any(ARRAY[{}])"_format(
+                         osm_type, osm_id, clslist));
     }
 }
 
@@ -85,9 +85,7 @@ int output_gazetteer_t::start()
             "  admin_level SMALLINT,"
             "  address HSTORE,"
             "  extratags HSTORE," +
-            (boost::format("  geometry Geometry(Geometry,%1%) NOT NULL") % srid)
-                .str() +
-            ")";
+            "  geometry Geometry(Geometry,{}) NOT NULL"_format(srid) + ")";
         if (m_options.tblsmain_data) {
             sql += " TABLESPACE " + m_options.tblsmain_data.get();
         }
--- a/src/table.cpp
+++ b/src/table.cpp
@@ -5,15 +5,12 @@
 #include <exception>
 #include <utility>
 
-#include <boost/format.hpp>
-
+#include "format.hpp"
 #include "options.hpp"
 #include "table.hpp"
 #include "taginfo.hpp"
 #include "util.hpp"
 #include "wkb.hpp"
-
-typedef boost::format fmt;
 
 #define BUFFER_SEND_SIZE 1024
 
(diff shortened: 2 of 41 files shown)


src/table.cpp
```diff
@@ -207,17 +202,17 @@ void table_t::stop(bool updateable, bool enable_hstore_index,
...
                 "    ORDER BY "
                 "ST_GeoHash(ST_Transform(ST_Envelope(way),4326),10)\n"
                 "    COLLATE \"C\"",
                 m_target->name, m_table_space);
             m_sql_conn->exec("RESET client_min_messages");
         }
-        m_sql_conn->exec("DROP TABLE %1%", m_target->name);
-        m_sql_conn->exec("ALTER TABLE %1%_tmp RENAME TO %1%", m_target->name);
+        m_sql_conn->exec("DROP TABLE {}", m_target->name);
+        m_sql_conn->exec("ALTER TABLE {}_tmp RENAME TO {}", m_target->name);
```
lonvia: second parameter missing

src/table.cpp
```diff
@@ -207,17 +202,17 @@ void table_t::stop(bool updateable, bool enable_hstore_index,
                in OSM. */
             m_sql_conn->exec("SET client_min_messages = WARNING");
             m_sql_conn->exec(
-                "CREATE TABLE %1%_tmp %2% AS\n"
-                "  SELECT * FROM %1%\n"
+                "CREATE TABLE {}_tmp {} AS\n"
+                "  SELECT * FROM {}\n"
```
lonvia: parameter missing

src/table.cpp
```diff
@@ -249,11 +244,11 @@ void table_t::stop(bool updateable, bool enable_hstore_index,
...
                     m_target->name);
 
-                m_sql_conn->exec("CREATE TRIGGER %1%_osm2pgsql_valid "
+                m_sql_conn->exec("CREATE TRIGGER {}_osm2pgsql_valid "
                                  "BEFORE INSERT OR UPDATE\n"
-                                 "  ON %1%\n"
+                                 "  ON {}\n"
                                  "    FOR EACH ROW EXECUTE PROCEDURE "
-                                 "%1%_osm2pgsql_valid();",
+                                 "{}_osm2pgsql_valid();",
```
lonvia: parameter missing

tests/test-db-copy-mgr.cpp
```diff
@@ -182,9 +182,9 @@ TEST_CASE("db_copy_mgr_t")
 
         auto c = db.connect();
 
-        auto sql = boost::format("SELECT h->'%1%' from test_copy_mgr");
+        std::string const sql{"SELECT h->'{}' FROM test_copy_mgr"};
```
lonvia: char const *?

joto:
I have re-done this PR using format strings parsed at compile time (works from C++14). Uses a slightly ugly syntax `"fmt string"_format(params)`, but the compile time check is worth it. Travis now tests the C++14 build on bionic.

src/table.cpp
```diff
@@ -263,21 +258,24 @@ void table_t::stop(bool updateable, bool enable_hstore_index,
...
             }
             for (size_t i = 0; i < hstore_columns.size(); ++i) {
-                m_sql_conn->exec("CREATE INDEX ON %1% USING GIN (\"%3%\") %4%",
-                                 m_target->name, i, hstore_columns[i],
-                                 (table_space_index
-                                      ? "TABLESPACE " + table_space_index.get()
-                                      : ""));
+                m_sql_conn->exec(
+                    "CREATE INDEX ON {} USING GIN (\"{}\") {}"_format(
+                        m_target->name, i, hstore_columns[i],
```
lonvia: 'i' needs to be removed. There is no %2% in the original anymore. I'm mildly surprised that the tests didn't catch that.

joto: With the `_format` this should have been detectable at compile time.

joto: It seems this is expected behaviour for fmt. too many parameters isn't considered an error. i have fixed this and did a force push.

src/output-gazetteer.cpp
```diff
@@ -42,9 +42,9 @@ void output_gazetteer_t::delete_unused_classes(char const *osm_type,
     if (!clslist.empty()) {
         clslist[clslist.size() - 1] = '\0';
         // Delete places where classes have disappeared for this object.
-        m_conn->exec("DELETE FROM place WHERE osm_type = '%1%' "
-                     "AND osm_id = %2% and class = any(ARRAY[%3%])",
-                     osm_type, osm_id, clslist);
+        m_conn->exec("DELETE FROM place WHERE osm_type = '{}'"
+                     " AND osm_id = {} and class = any(ARRAY[{}])"_format(
+                         osm_type, osm_id, clslist));
```
lonvia: This actually broke the gazetteer output because of the hacky line 43. Apparently boost:format ended the string at the first \0 while fmt now (correctly) prints everything in the std::string according to its size. Postgresql does not quite like the additional \0 in the copy stream.
This is already fixed in #1008, so no need to do anything. Just thought I mention it as an example why one should not mess with \0 in strings.

PR #1005, opened 2019-11-27
Add tests/__pycache__ to .gitignore

joto:
This is created by the new tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ build
 
 docs/html
 docs/latex
+
+tests/__pycache__
--- a/tests/regression-test.py.in
+++ b/tests/regression-test.py.in
@@ -1,4 +1,4 @@
-#! @PYTHON_EXECUTABLE@
+#! @PYTHON_EXECUTABLE@ -B
 
 import sys
 import unittest


lonvia:
According to the python man page, you can also run python with `-B` to suppress writing the cache files. Maybe we should always do that to keep the source tree clean?

joto:
Maybe that's better, yes. The extra overhead of always compiling the script is probably nothing compared to the time the test runs anyway.

I just tested this patch and it works well:
```
diff --git a/tests/CMakeLists.txt b/tests/CMakeLists.txt
index b8ce786..92f8243 100644
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -92,7 +92,7 @@ if (PYTHONINTERP_FOUND)
     configure_file(${CMAKE_CURRENT_SOURCE_DIR}/regression-test.py.in
                    ${CMAKE_CURRENT_BINARY_DIR}/regression-test.py)
     add_test(NAME regression-test-pbf
-             COMMAND ${CMAKE_CURRENT_BINARY_DIR}/regression-test.py)
+             COMMAND ${PYTHON_EXECUTABLE} -B ${CMAKE_CURRENT_BINARY_DIR}/regression-test.py)
     set_tests_properties(regression-test-pbf PROPERTIES TIMEOUT ${TESTING_TIMEOUT})
     set_tests_properties(regression-test-pbf
                          PROPERTIES FIXTURES_REQUIRED Tablespace)
```

PR #1006, opened 2019-11-27
README: 32bit systems are unsupported

joto:


mmd-osm:
Would it make sense to also add a compiler warning (or maybe even error)?

```C++
diff --git a/src/osm2pgsql.cpp b/src/osm2pgsql.cpp
index d048895..ad9cb49 100644
--- a/src/osm2pgsql.cpp
+++ b/src/osm2pgsql.cpp
@@ -35,6 +35,7 @@
 #include "reprojection.hpp"
 #include "util.hpp"
 
+#include <cstdint>
 #include <cstdio>
 #include <cstdlib>
 #include <ctime>
@@ -44,6 +45,10 @@
 
 #include <libpq-fe.h>
 
+#if INTPTR_MAX != INT64_MAX
+#warning osm2pgsql is only supported on 64bit systems
+#endif
+
 int main(int argc, char *argv[])
 {
     fprintf(stderr, "osm2pgsql version %s (%zu bit id space)\n\n", VERSION,
```

lonvia:
We can't formulate it this way because there is a windows 32bit binary linked in the README. We'd need to discuss removing this first.

PR #1008, opened 2019-11-30
Fix deletion on update in gazetteer

### LINKED ISSUE #964: osm2pgsql hangs in gazetteer mode on diff
Last night osm2pgsql hung on both Nominatim servers on the following change: http://lonvia.de/attic/hanging-change-191028.osc There was no activity from osm2pgsql anymore. Last output:

```
Parsing gazetteer style file '/srv/nominatim.openstreetmap.org/nominatim/settings/import-full.style'.
Using projection SRS 4326 (Latlong)

Reading in file: /srv/nominatim.openstreetmap.org/bin/osmosischange.osc
Using XML parser.
Processing: Node(4k 4.2k/s) Way(0k 0.79k/s) Relation(20 2.50/s))
```

Postgresql activity logs show that [relation 9718561 version 5](https://www.openstreetmap.org/relation/9718561) was likely last processed. Restarting the updates helped as a workaround.

The tile servers are fine, so I suppose it is a gazetteer-specific issue.

lonvia:
Objects so far have been deleted directly from the main thread in gazetteer. This may lead to deadlock as the updates in the copy thread trigger SQL triggers which may touch an object currently being deleted.

Gazetteer objects are now deleted in the copy thread in a batch just like we do for all other tables. For this we need a different kind of deleter as the primary key is different for gazetteer's place table. For this, delete lists become first class objects and the copy manager becomes responsible for adding the right deleter to the copy buffer object.

Fixes #964.

### DIFF
--- a/src/db-copy.cpp
+++ b/src/db-copy.cpp
@@ -104,15 +104,12 @@ void db_copy_thread_t::disconnect() { m_conn.reset(); }
 
 void db_copy_thread_t::write_to_db(db_cmd_copy_t *buffer)
 {
-    if (buffer->deleter.has_data() ||
+    if (buffer->has_deletables() ||
         (m_inflight && !buffer->target->same_copy_target(*m_inflight.get()))) {
         finish_copy();
     }
 
-    if (buffer->deleter.has_data()) {
-        buffer->deleter.delete_rows(buffer->target->name, buffer->target->id,
-                                    m_conn.get());
-    }
+    buffer->delete_data(m_conn.get());
 
     if (!m_inflight) {
         start_copy(buffer->target);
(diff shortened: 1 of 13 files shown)


src/db-copy.cpp
```diff
@@ -119,24 +118,19 @@ void db_copy_thread_t::write_to_db(db_cmd_copy_t *buffer)
...
-    sql.reserve(buffer->target->name.size() + buffer->deletables.size() * 15 +
-                30);
-    sql += buffer->target->name;
-    sql += " WHERE ";
-    sql += buffer->target->id;
-    sql += " IN (";
-    for (auto id : buffer->deletables) {
-        sql += std::to_string(id);
+    std::string sql = "DELETE FROM {} WHERE {} IN ("_format(table, column);
+    sql.reserve(m_deletables.size() * 15 + sql.size());
```
joto: Unexplained magic value 15 / unexplained calculation.

src/db-copy.cpp
```diff
@@ -119,24 +118,19 @@ void db_copy_thread_t::write_to_db(db_cmd_copy_t *buffer)
...
+
+    for (auto id : m_deletables) {
+        sql += fmt::to_string(id);
```
joto: This will create a string for each id and then append it. Consider using `fmt::basic_memory_buffer` or `fmt::format_to` with `back_inserter`.

src/gazetteer-style.cpp
```diff
@@ -48,20 +61,26 @@ void gazetteer_style_t::clear()
...
-            if (std::get<2>(e) & SF_MAIN_NAMED) {
-                return !m_names.empty();
-            }
-            // XXX should handle SF_MAIN_NAMED_KEY as well
+    std::string l;
 
-            return true;
+    for (auto const &m : m_main) {
+        bool do_include = true;
+        if (std::get<2>(m) & SF_MAIN_NAMED) {
```
joto: Why the `do_include`? What's up with the `XXX`?

src/gazetteer-style.cpp
```diff
@@ -48,20 +61,26 @@ void gazetteer_style_t::clear()
...
+        // XXX should handle SF_MAIN_NAMED_KEY as well
 
-        return false;
-    });
+        if (do_include) {
+            l += "'{}',"_format(std::get<0>(m));
+        }
+    }
+
+    if (!l.empty())
```
joto: {}

src/output-gazetteer.hpp
```diff
@@ -108,23 +107,19 @@ class output_gazetteer_t : public output_t
...
-    void delete_place(char const *osm_type, osmid_t osm_id);
     int process_node(osmium::Node const &node);
     int process_way(osmium::Way *way);
     int process_relation(osmium::Relation const &rel);
     void connect();
 
-    void prepare_query_conn() const;
-
     void delete_unused_full(char const *osm_type, osmid_t osm_id)
     {
```
joto: Why is the parameter here a `char const *`, but `delete_object()` uses a `char`?

src/db-copy.cpp
lonvia: Added a comment.

src/db-copy.cpp
lonvia: Aha, that's where the additional second in the regression tests went. Switched to `fmt::memory_buffer` (also below). The `back_inserter` doesn't seem to be necessary. Compile-time strings only work with the FMT_STRING() macro in this constellation. That's mildly annoying.

src/gazetteer-style.cpp
lonvia: Because I thought the if would be too complicated once the XXX is solved. On further thinking, the XXX needs to be solved in a totally different place. Something for another PR. (Note that the XXX was there before, so no regression here.)

src/gazetteer-style.cpp
lonvia: Done.

src/output-gazetteer.hpp
lonvia: It was historically easier to use. No longer necessary, made it a char parameter.

PR #1012, opened 2019-12-03
Move declaration of static `db` into tests

joto:
`pg::tempdb_t db` or `testing::db::import_t db` is defined as static in many test files. This is problematic, because the class may throw errors which we can't catch and the testing framework can't report correctly then.

This is caught by clang-tidy with error messages like this:
"warning: initialization of 'db' with static storage duration may throw an exception that cannot be caught [cert-err58-cpp]"

This commit fixes all these cases.

This commit also adds a bunch of `const`s.

### DIFF
--- a/tests/test-db-copy-mgr.cpp
+++ b/tests/test-db-copy-mgr.cpp
@@ -7,11 +7,10 @@
 #include "common-pg.hpp"
 #include "db-copy-mgr.hpp"
 
-static pg::tempdb_t db;
-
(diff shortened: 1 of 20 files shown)


lonvia:
The tempdb_t object is intentionally static so that the database is only created once per test run. Also note that the constructor for tempdb_t catches exceptions and exits with an appropriate error message. I admit though that there is still room for improvement in that constructor, e.g. adding noexcept and having an extra catch block for the extension creation that drops the database again.

joto:
Closed in favour of #1025.

PR #1017, opened 2019-12-05
Add compile_commands.json to gitignore

joto:
Having a compile_commands.json is useful for you-complete-me


lonvia:
Sorry, but the rule is that .gitignore should only contain project-specific files to ignore. You can add tool-specific files into the .gitignore file in your home directory.

PR #1021, opened 2019-12-08
Make some functions return void instead of int

joto:
This is the second batch of changes after #1020 switching the return type of some functions from int to void where the result is never used.

### DIFF
--- a/src/output-pgsql.cpp
+++ b/src/output-pgsql.cpp
@@ -111,7 +111,7 @@ void output_pgsql_t::enqueue_ways(pending_queue_t &job_queue, osmid_t id,
     }
 }
 
-int output_pgsql_t::pending_way(osmid_t id, int exists)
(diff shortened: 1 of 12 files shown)


src/output-pgsql.cpp
```diff
@@ -198,10 +196,9 @@ int output_pgsql_t::pending_relation(osmid_t id, int exists)
         }
 
         auto const &rel = rels_buffer.get<osmium::Relation>(0);
-        return pgsql_process_relation(rel);
+        pgsql_process_relation(rel);
+        return;
```
lonvia: Any particular reason?

joto: The new way makes clear that `pgsql_process_relation` doesn't return anything. If we ever change the function this is in to return something again, this new code will break regardless of whether `pgsql_process_relation` returns something or not, helping us to detect that there is a place here that needs looking at.

Issue #1036, opened 2019-12-19
New "flex" backend


joto:
In the past weeks I have started working on a new output backend. I am calling this the "flex" backend, because it is more flexible than any of the other backends and should eventually be able to support all use cases currently supported by the other backends and many more. This flexibility is mostly due to an increase in the use of Lua scripts which are used for configuration as well as more powerfull callback functions.

This work has mostly been triggered by #230, but it also touches many other issues like #901 .

Here are some design thoughts:

# Osm2pgsql design for a more flexible backend and Lua integration

We write a new backend according to the ideas presented in this document. The
new backend should be able to do everything the current backends can do and
would replace them in the long-term. (This also means that the C++-only
transforms without Lua will not be supported forever.)

# Table configuration

The new setup can work with any number of database tables. We provide Lua
functions define tables and their columns. Different tables can have different
column setups, for instance some tables might have lots of columns for tags
("old-style"), some have only an HSTORE or JSONB column for tags.

# Lua callbacks

For every OSM object (node, way, or relation) that is processed by osm2pgsql, a
Lua callback is called with the data from this object as parameter. The Lua
callback can process the data of the OSM object in any way it likes and emits
zero, one, or more table entries (for any table) that should be added by
osm2pgsql. This is done by calling an `add_row()` function on those tables.

Processing in a Lua callback includes but is not limited to:

* Decide which table or tables (if any) the data should go into.
* Map attributes and tags from OSM objects to tables and their columns.
* Convert data types (oneway=yes as boolean, width=10ft as integer in meters,
  put several tags into an HSTORE or JSON(B) column, etc.)
* Convert geometries (in limited ways), like turning around line strings or
  calculating centroids.

Part of this functionality would be supported by C++ functions callable from
Lua (something like geom = MakeCentroid(geom)).

# Getting data of related objects

For some use cases we need data from related objects that we want to add to the
currently processed object. We might, for instance, want the roles and tags
from all relations a way is a member of to be added to the way data in some
form.

This is solved by allowing Lua functions to
* tell osm2pgsql that it should process certain OSM objects (`mark` the object)
* store data in temporary storage in memory
* retrieve data from temporary storage

So in the mentioned use case the following would happen:
1. When processing a way with, say a `highway` tag, the Lua script tells
   osm2pgsql: Please find all relations having this way as member and call
   a Lua callback function later to re-process this way.
2. In a later step osm2pgsql processes relations (either because they are in
[...]

Issue #901, opened 2019-01-16
Re-do Lua API


pnorman:
Now that we have experience with the Lua API, the multi-backend, and have new feature requests which require API changes, we should redo the API.

To allow a transform to be compatible with current API we can use different function names, and then a transform can implement both entry points reasonably easily.

The current API is

```lua
function filter_tags_node(tags, num_tags)
return filter, tags

function filter_tags_way(tags, num_tags)
return filter, tags, polygon, roads

function filter_basic_tags_rel(tags, num_tags)
return filter, tags

function filter_tags_relation_member(tags, member_tags,
    roles, num_members)
return filter, tags, member_superseded, boundary,
    polygon, roads
```

Experience has shown num_tags is not needed because we're not in C. `next(tags) == nil` does the same check, is more idiomatic Lua, and is faster.

With old-style MP support gone, we don't need the superseded tests.

New features desired are
- Knowing if a way is closed or not to determine area formation (blocks recreation of C transform in multi backend)
- Allowing columns on objects from ways to be set based on relations they are a member of #230 

What other changes do we want to make?

cc @kevinkreiser @systemed

kevinkreiser:
Propogating tags to relations (the second bullet point) was the main feature I was hoping for! If anything else springs to mind I'll post back.

pnorman:
We should let the lua transforms set the distance for line splitting on a per-object area, as well as set if MULTIPOLYGONs are broken up

systemed:
In an ideal world I'd like filter_tags_way to have some very limited access to the way geometry (even if just the first node), for national rendering variations. Currently I cope with this by pre-/post-processing. This may or not be at all possible!

pnorman:
If it is possible - and I'm not sure on that - bounding box is probably the sensible arg to pass

lonvia:
It's possible, we have the information but it would make the processing more expensive. Currently we only retrieve geometries for ways that have not been filtered. If filtering needs the geometries then we need to retrieve it for all ways. So my favourite solution would be a callback function for Lua that can retrieve the geometry when needed. (e.g. in the particular case only for ways that are rendered in a country-specific style.).

systemed:
+1 - that's something I do in both tilemaker and my OSRM fork and it works very effectively.

mmd-osm:
This might be a new use case for the Lua API: https://twitter.com/realgiggls/status/1138066068117884929

lonvia:
Implemented in new 'flex' output.

Issue #768, opened 2017-06-22
Creating new Schema for OSM data


JakobMiksch:
**Feature Request:**
I would be nice to choose the target schema in the database with the command line. So far I only know this way to load the OSM data in an own schema: https://gis.stackexchange.com/a/242814
An option like `--schema` would be great.

lonvia:
There is #619 but that has gone stale again.

Stalfur:
Agree - ability to choose schema would be most useful.

mmd-osm:
I have a question about the search_path based approach used in #279: this seems to assume that we need to set the search_path to `SET search_path TO %1%,public;` everywhere, even though we need the public schema only in case of CREATE TABLE (e.g.due to the depency of hstore which presumably lives in the public schema).

I tried the patch below with a hardcoded schema name demo3. It seems to work here, even if there are no planet* tables in the public schema. 

<details>

```c++
diff --git a/db-copy.cpp b/db-copy.cpp
index a4f156e..47e27ab 100644
--- a/db-copy.cpp
+++ b/db-copy.cpp
@@ -104,6 +104,7 @@ void db_copy_thread_t::connect()
     // Let commits happen faster by delaying when they actually occur.
     pgsql_exec_simple(m_conn, PGRES_COMMAND_OK,
                       "SET synchronous_commit TO off;");
+    pgsql_exec_simple(m_conn, PGRES_COMMAND_OK, "SET search_path TO demo3;");
 }
 
 void db_copy_thread_t::disconnect()
diff --git a/middle-pgsql.cpp b/middle-pgsql.cpp
index b0f5ba1..0c76838 100644
--- a/middle-pgsql.cpp
+++ b/middle-pgsql.cpp
@@ -736,6 +736,8 @@ void middle_pgsql_t::start()
         util::exit_nicely();
     }
 
+    pgsql_exec_simple(m_query_conn, PGRES_COMMAND_OK, "SET search_path TO demo3,public;");
+
     if (append) {
         // Prepare queries for updating dependent objects
         for (auto &table : tables) {
@@ -861,6 +863,8 @@ middle_pgsql_t::get_query_instance(std::shared_ptr<middle_t> const &from) const
         src->out_options->database_options.conninfo().c_str(), src->cache,
         src->persistent_cache));
 
+    mid->exec_sql("SET search_path TO demo3;");
+
     // We use a connection per table to enable the use of COPY
     for (int i = 0; i < NUM_TABLES; i++) {
[...]

Issue #734, opened 2017-04-06
unused column "way_area" in the _line and _roads tables


thbeutin:
When i import an OSM extract using the [empty.style](https://raw.githubusercontent.com/openstreetmap/osm2pgsql/master/empty.style) i got the column "way_area" in the tables for lines and roads containing NULLs only. The values in the polygon table are fine.

Is this intentionally or a mistake?

pnorman:
I thought we didn't have a way_area column for lines, but checking shows the column is there too.

It's normal to not have an area for those, since they're not polygons.

thbeutin:
> It's normal to not have an area for those, since they're not polygons.

Yes, that's obvious, and this is the main reason for opening this issue ;)
But wouldn't [ST_Length](http://postgis.net/docs/ST_Length.html) fit here perfectly?

joto:
I expect we don't want to change this for backwards compatibility. The new "flex" backend allows you to set exactly the columns you need, so when people switch to it, this issue becomes moot.

PR #1037, opened 2019-12-19
Implementation of new flex backend

joto:
This PR introduces a new flex backend, see #1036 for the ideas behind this.

If you want to try this: Compile as usual, Lua support must be enabled. Then run osm2pgsql with `-O flex` and `-S some-config.lua`. Docs in `docs/flex.md`. Look into `flex-config` directory for some example config/style files with inline documentation. Start with `simple.lua`.

This PR more or less resolves the following issues: #109, #153, #230, #277, #328, #573, #677, #734, #901

In addition this PR impacts the following issues: #107, #328, #768


lonvia:
I've looked through the code and this looks good for a first experimental implementation. There might be room for consolidation with the existing code but I don't think it's worth doing that before the interface of the output is stable.

I would suggest to mark this as experimental and merge this in its current state. This makes it easier for people to test the new output and also means that we can have smaller PRs with further improvements. @pnorman what do you think?

@joto can you please add the flex output to the existing READMEs with a big fat warning that this is experimental and the API might change without notice.

appveyor.yml
```diff
@@ -72,6 +72,8 @@ install:
   - git clone https://github.com/alex85k/wingetopt -b %WINGETOPT_VER% C:\wingetopt
   - python -V
   - python -m pip install psycopg2
+  - ps: Invoke-Webrequest 'https://framagit.org/fperrad/lua-MessagePack/raw/master/src5.3/MessagePack.lua?inline=false' -OutFile c:\lua\MessagePack.lua
```
mmd-osm: Are there any files added by MessagePack that need to included in the zip file we build towards the end of the process?

joto: Well, the flex backend in osm2pgsql will not work without MessagePack available. Whether that means that we want to include it in a zip file, I don't know. osm2pgsql probably wouldn't find it anyway, because it depends on the Lua path.

mmd-osm: MessagePack seems to be a single lua file only, but the relevant version also depends on LuaJIT (the 5.3 version probably won’t run in that case). I think for the time being, having some description where to download MessagePack and how to set lua path would be a good.

joto: The newest version of this PR doesn't need MessagePack any more.

pnorman:
> I would suggest to mark this as experimental and merge this in its current state. This makes it easier for people to test the new output and also means that we can have smaller PRs with further improvements. @pnorman what do you think?

I'm okay with this.

docs/flex.md
```diff
@@ -0,0 +1,239 @@
...
+
+See also the example config files in the `flex-config` directory which contains
+lots of comments to get you started.
+
+## The Lua config file
+
+All configuration is done through the `osm2pgsql` object in Lua. It has the
+following fields:
+
+* `osm2pgsql.version`: The version of osm2pgsql as a string.
```
mmd-osm: I have a question in the context of https://github.com/openstreetmap/osm2pgsql/pull/1064 : 

`osm2pgsql.version` calls `get_osm2pgsql_version()` , which in turn returns a string like "1.2.0 (1.2.0-239-gcc9748f)". If the idea was to enable certain features in a lua script depending on the osm2pgsql version, it might be easier to have `version` return a simple string like "1.2.0" and maybe have an additional `fullversion` attribute returning the full version.

joto: I have changed this to return the released version number only.

jeisenbe:
Thank you, this looks very useful. Will this be part of an official release soon?

lonvia:
It will be part of the next release but I cannot give an ETA on that. We need some more testing (and more tests). So, the more people try out the master branch or volunteer tests, the sooner the release. ;)

PR #1041, opened 2019-12-24
Travis fix for osx

### LINKED ISSUE #1035: OSX builds on travis fail
The OSX builds on travis have been failing for a few days. This seems to be related to travis switching to OS X Mojave (10.14) which doesn't include the system header files in `/usr/include` any more. I have tried following some advice found "on the internet" to fix this (in [this branch](https://github.com/joto/osm2pgsql/tree/travis-test-osx)), but could not fix it.

Anybody with a Mac who can fix this?

mmd-osm:
This is pretty much the only configuration I got working on travis for OSX.

Unfortunately, I won't work with newer xcode releases due to some Apple bug (https://github.com/Homebrew/homebrew-core/issues/44579, see error message below). Also, it uses gcc-9 instead of clang. If someone happens to have lots of time and patience, we might get clang working at some point in the future again.

Changes were heavily inspired by https://github.com/google/crc32c/blob/master/.travis.yml

Fixes #1035

--- 

```
/Applications/Xcode-11.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/resource.h:443:34: error: expected initializer before '__OSX_AVAILABLE_STARTING'

  443 | int     getiopolicy_np(int, int) __OSX_AVAILABLE_STARTING(__MAC_10_5, __IPHONE_2_0);
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -88,16 +88,36 @@ matrix:
       addons: *bionic_clang7_pg96
 
   # ---- OSX + CLANG ---------------------------
-  #  - os: osx
-  #    compiler: clang
-  #    env: T="osx_clang_NoDB"
-  #         BUILD_TYPE="Debug" LUAJIT_OPTION="OFF" TEST_NODB=1
-  #         CXXFLAGS="-pedantic -Wextra -Werror" CPPVERSION=11
-  #    before_install:
-  #      - brew install lua; brew install lua
-  #    before_script:
-  #      - proj | head -n1
-  #      - lua -v
+    - os: osx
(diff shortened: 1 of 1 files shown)


lonvia:
Thanks for trying this. I'm just not convinced that it adds much to the other travis tests in this form. The point of the MacOS tests is mainly to ensure that it compiles with the native Xcode. gcc-9 is sufficiently tested under Linux.

mmd-osm:
Ok, I'll close it then.

PR #1050, opened 2020-01-04
Don't use ST_GeoHash for ordering on Postgis >= 2.4

lonvia:
Follow-up for #1016. Use native geometry order for clustering but only for Postgis versions >= 2.4.

### DIFF
--- a/src/table.cpp
+++ b/src/table.cpp
@@ -186,29 +186,44 @@ void table_t::stop(bool updateable, bool enable_hstore_index,
         fprintf(stderr, "Sorting data and creating indexes for %s\n",
                 m_target->name.c_str());
 
-        if (srid == "4326") {
-            /* libosmium assures validity of geometries in 4326, so the WHERE can be skipped.
-               Because we know the geom is already in 4326, no reprojection is needed for GeoHashing */
-            m_sql_conn->exec(
-                "CREATE TABLE {0}_tmp {1} AS\n"
-                "  SELECT * FROM {0}\n"
-                "    ORDER BY ST_GeoHash(way,10)\n"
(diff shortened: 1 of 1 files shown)


src/table.cpp
```diff
@@ -186,30 +186,45 @@ void table_t::stop(bool updateable, bool enable_hstore_index,
...
 
-        if (srid == "4326") {
-            /* libosmium assures validity of geometries in 4326, so the WHERE can be skipped.
-               Because we know the geom is already in 4326, no reprojection is needed for GeoHashing */
-            m_sql_conn->exec(
-                "CREATE TABLE {0}_tmp {1} AS\n"
-                "  SELECT * FROM {0}\n"
-                "    ORDER BY ST_GeoHash(way,10)\n"
-                "    COLLATE \"C\""_format(m_target->name, m_table_space));
+        // Notices about invlaid geometries are expected and can be ignored
```
joto: typo "invlaid"

src/table.cpp
```diff
@@ -186,30 +186,45 @@ void table_t::stop(bool updateable, bool enable_hstore_index,
...
+            // libosmium assures validity of geometries in 4326.
+            // Transformation to another projection could make the geometry
+            // invalid. Therefore add a filter to drop those.
+            sql += " WHERE ST_IsValid(way)";
+        }
+
+        auto res = m_sql_conn->query(
+            PGRES_TUPLES_OK,
+            "SELECT regexp_split_to_table(postgis_lib_version(), '\\.')");
+        auto postgis_major = std::stoi(res.get_value_as_string(0, 0));
```
joto: 2x const

PR #1051, opened 2020-01-07
Remove superfluous checks for append mode.

joto:
Inside the pgsql output (and only in that output) we checked whether we
are in append mode in (the node|way|relation)_(modify|delete) functions.
This is not needed because oms2pgsql checks early on that option -a
isn't used without -s.

This can also be seen because at the callsite of those functions we are
doing an unchecked dynamic_cast that would have failed and returned a
nullptr if we are not in slim mode. So we would have never reached those
checkes. I have added assert()s to make clear we are relying on those
dynamic_casts to work.


lonvia:
Looks good. I just added a tiny test for your assertion "oms2pgsql checks early on that option -a
isn't used without -s".

PR #1064, opened 2020-01-21
Put git commit into osm2pgsql version string

joto:
Better information about which version a bug reporter is using.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,27 @@ option(EXTERNAL_PROTOZERO "Do not use the bundled protozero" OFF)
 option(EXTERNAL_FMT       "Do not use the bundled fmt"       OFF)
 
 #############################################################
+# Version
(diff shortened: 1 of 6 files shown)


mmd-osm:
Just wanted to mention that the Appveyor binary shows the tagged version only, which does not include a commit hash. I believe the expected format should be more like 1.2.0-238-g3d172b6 here.

BTW: I think it would be helpful to include `osm2pgsql.exe -V`  in the build script for easier validation.

```
$ wine osm2pgsql.exe -V
osm2pgsql version 1.2.0 (64 bit id space)

Libosmium 2.15.4
Lua 5.3.4
```

joto:
@mmd-osm If the git command isn't found in the CMake config, it will fall back to the version set in the CMake config. Maybe this check fails on Appveyor for some reason. Maybe somebody with a Windows machine can debug this.

Issue #1079, opened 2020-02-22
Lua library for flex backend


joto:
The flex backend uses Lua much more than before. Many Lua scripts will need similar code to handle common cases like parsing/converting tags, checking for existence of certain tags etc. Some code for that is already [here](https://github.com/openstreetmap/osm2pgsql/blob/master/src/init.lua). This code is compiled into osm2pgsql and available as functions in the `osm2pgsql` namespace.

Please use this issue to propose functions for the library or discuss needs.

joto:
* [x] Remove common unused tags (by key or prefix) like `source`, or `note`
* [ ] Check for specific tags (for instance to decide whether something should go into a specific table)
* [ ] Check for way tags that should be handled as polygons
* [ ] Convert special tag formats (with semicolons, numbers with units, etc.) What do we need here?

mmd-osm:
A few ideas I could find in the Overpass docs:


* Number conversions
   - Check if string is a number
   - Convert string to a number

* Semicolons
   - Split semicolon delimited string into table
   - Check if string exists in this table

* Units
   -  Prefix / Suffix -> convert "34 mph" into "34" and "mph".

pnorman:
Why do these need to go directly into osm2pgsql?

joto:
@pnorman They don't *need* to go, this would just be something helpful so that not everybody has to reinvent the wheel. Of course it could be maintained outside osm2pgsql, but I this way it is easier to use for everybody.

gravitystorm:
I'd like to see the current integer handling available as a library function, namely the (currently default in pgsql backend) support for ranges in integer tags. For example, `<tag k="population" v="10000-20000" />` with an `int4` column type is outputted as 15000.

pnorman:
I have a few functions from cleartables that would probably be useful, I'll dig them up

pnorman:
Where should the test code for init.lua go?

joto:
@pnorman I have started writing tests for init.lua in a local branch somewhere under `tests` as simple Lua scripts using the Lua-builtin `assert`, but I don't know yet how to best integrate this with CTest and the rest of the tests. So far I have been running them "manually" by calling the Lua interpreter directly.

I see several issues here:
* Some tests need the whole osm2pgsql framework around the Lua code. Those should be run as part of the other osm2pgsql tests.
* All unit tests that only test Lua functions and don't need osm2pgsql around it should probably tested without it, just to keep things easy and simple.
[...]

PR #1084, opened 2020-02-26
Add some Lua helper functions with docs, and tests for them

joto:
See #1079

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,7 +98,7 @@ before_install:
       export POSTGIS_PKG="$POSTGIS_PKG $PPG $PPG-scripts";
     done
   - echo $POSTGIS_PKG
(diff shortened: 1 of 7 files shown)


mmd-osm:
I'd say this might be for performance reasons to avoid copying fields and setting up tables for each loop iteration.

joto:
@pnorman Yes, @mmd-osm is right. The idea here is the `create` function gets the data in a format that's easy for the user to write and converts it into something that is as fast as possible to execute for millions of objects.

For instance in the case of the "clean" function, it allows you to set up equality and prefix matches using the intuitive "foo" and "foo*" syntax, but we don't want to check for the `*` every time etc. So the create function sets up two different data structures for the different types of matches.

PR #1140, opened 2020-04-26
Quick fix for problem with updating data in flex backend

joto:
Under some circumstances when using the two-stage processing of the flex
backend data was not updated correctly. This happens when a way changes
that is in a relation and that relation has another way in common with
a different relation.

The solution for this is introducing a stage1c processing, which this
commit does in a really hacky way. To do this cleanly we need to
refactor a lot of the middle and id_tracker code.

Processing happens in these stages:

* stage 1a: Read input file and process all objects in it.
* stage 1b: Process dependent objects of objects from 1a (ie changed
  nodes trigger changes in ways with those nodes, changes in all
  objects potentially trigger changes in relations).
* stage 1c: Process dependent objects (currently only relations) of
  objects marked during stage 1a/1b (this one is new). No more marking
  is done in stage 1c.
* stage 2: Process objects marked in stage 1a or 1b.

### DIFF
--- a/src/osmdata.cpp
+++ b/src/osmdata.cpp
@@ -13,6 +13,7 @@
 #include "db-copy.hpp"
 #include "format.hpp"
 #include "middle.hpp"
+#include "middle-pgsql.hpp"
 #include "osmdata.hpp"
 #include "output.hpp"
 #include "util.hpp"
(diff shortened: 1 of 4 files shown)


joto:
Superseeded by #1215 which was merged. Closing here.

PR #1142, opened 2020-04-28
Provide a stack trace on a lua error with flex backend

### LINKED ISSUE #1136: Provide a stack trace on a lua error with flex backend
If you have an error in your lua script you get an error message like
```
Osm2pgsql failed due to ERROR: Failed to execute lua processing function: openstreetmap-carto.lua:533: Error in 'add_row': Trying to add relation to table 'planet_osm_line'
```

This doesn't provide as much information as a stack trace.

mmd-osm:
Closes #1136

Almost exactly the same code that is in use in the Lua interpreter, except for the signal handling: https://github.com/lua/lua/blob/master/lua.c#L109-L142

### DIFF
--- a/src/output-flex.cpp
+++ b/src/output-flex.cpp
@@ -983,6 +983,31 @@ void output_flex_t::add_row(table_connection_t *table_connection,
     }
 }
 
+static int msghandler (lua_State *L)
+{
+    const char *msg = lua_tostring(L, 1);
+    if (msg == NULL) {                           /* is error object not a string? */
+        if (luaL_callmeta(L, 1, "__tostring") && /* does it have a metamethod */
+            lua_type(L, -1) == LUA_TSTRING)      /* that produces a string? */
+            return 1;                            /* that is the message */
+        else
+            msg = lua_pushfstring(L, "(error object is a %s value)",
+                                     luaL_typename(L, 1));
+    }
+    luaL_traceback(L, L, msg, 1);                /* append a standard traceback */
+    return 1;                                    /* return the traceback */
+}
+
+static int docall (lua_State *L, int narg, int nres) {
+    int status;
+    int base = lua_gettop(L) - narg;       /* function index */
+    lua_pushcfunction(L, msghandler);      /* push message handler */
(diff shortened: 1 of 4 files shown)


joto:
This code breaks about every convention we have in our source code.

mmd-osm:
Ok, so what do you suggest moving forward?

src/output-flex.cpp
```diff
@@ -983,6 +983,35 @@ void output_flex_t::add_row(table_connection_t *table_connection,
...
+    luaL_traceback(lua_state, lua_state, msg, 1); // append a standard traceback
+    return 1;                                     // return the traceback
+}
+
+int output_flex_t::do_pcall(int narg, int nres)
+{
+    int const base = lua_gettop(lua_state()) - narg; // function index
+    lua_pushcfunction(lua_state(), msghandler);      // push message handler
+    lua_insert(lua_state(), base);                   // put it under function and args
+    int status = lua_pcall(lua_state(), narg, nres, base);
```
joto, line 1010: Can be `const`

src/output-flex.cpp
```diff
@@ -983,6 +983,35 @@ void output_flex_t::add_row(table_connection_t *table_connection,
...
+{
+    assert(lua_state);
+
+    char const *msg = lua_tostring(lua_state, 1);
+    if (msg == nullptr) {                                // is error object not a string?
+        if (luaL_callmeta(lua_state, 1, "__tostring") && // does it have a metamethod
+            lua_type(lua_state, -1) == LUA_TSTRING) {    // that produces a string?
+            return 1;                                    // that is the message
+        }
+        else {
```
joto, line 996: Formatting looks unusual. Did you run `clang-format`?

src/output-flex.cpp
```diff
@@ -983,6 +983,35 @@ void output_flex_t::add_row(table_connection_t *table_connection,
     }
 }
 
+static int msghandler(lua_State *lua_state)
```
joto, line 986: Maybe a bit more descriptive function name?

PR #1145, opened 2020-05-02
Remove superfluous breaks

joto:


mmd-osm:
By the way, this issue was originally found by DeepCode.ai

PR #1168, opened 2020-05-12
New dependency manager

joto:
The new dependency manager takes over some of the responsibilities of the middle. This way we decouple the dependency tracking from the object storage. It makes the code easier to understand and to change.

There are more refactorings coming after this.

### DIFF
--- a/src/dependency-manager.cpp
+++ b/src/dependency-manager.cpp
@@ -1,0 +1,60 @@
+
+#include "dependency-manager.hpp"
+
+void full_dependency_manager_t::node_changed(osmid_t id) {
+    for (auto const way_id : m_object_store->get_ways_by_node(id)) {
+        way_changed(way_id);
+        m_ways_pending_tracker.mark(way_id);
+    }
+
+    for (auto const rel_id : m_object_store->get_rels_by_node(id)) {
+        m_rels_pending_tracker.mark(rel_id);
+    }
+}
+
+void full_dependency_manager_t::way_changed(osmid_t id) {
+    if (m_ways_pending_tracker.is_marked(id)) {
+        return;
+    }
+
+    for (auto const rel_id : m_object_store->get_rels_by_way(id)) {
+        m_rels_pending_tracker.mark(rel_id);
+    }
+}
+
+void full_dependency_manager_t::relation_changed(osmid_t id) {
+    if (m_rels_pending_tracker.is_marked(id)) {
+        return;
+    }
+
+    for (auto const rel_id : m_object_store->get_rels_by_rel(id)) {
+        m_rels_pending_tracker.mark(rel_id);
+    }
+}
+
+void full_dependency_manager_t::relation_deleted(osmid_t id) {
+    for (auto const rel_id : m_object_store->get_ways_by_rel(id)) {
+        m_ways_pending_tracker.mark(rel_id);
+    }
+}
+
+bool full_dependency_manager_t::has_pending() const noexcept
+{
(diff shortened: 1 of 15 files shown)


src/dependency-manager.hpp
```diff
@@ -0,0 +1,161 @@
...
+public:
+    virtual ~dependency_manager_t() = default;
+
+    /**
+     * Mark a node as changed to trigger the propagation of this change to
+     * ways and relations.
+     *
+     * This has to be called *after* the object was stored in the object store.
+     */
+    virtual void node_changed(osmid_t) {};
```
lonvia, line 43: remove semicolon, also below

src/dependency-manager.cpp
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    for (auto const rel_id : m_object_store->get_rels_by_way(id)) {
+        m_rels_pending_tracker.mark(rel_id);
+    }
+}
+
+void full_dependency_manager_t::relation_changed(osmid_t id) {
+    if (m_rels_pending_tracker.is_marked(id)) {
+        return;
```
lonvia, line 27: This is not true. A marked relation has not necessarily marked its dependencies. That would only be true if we called relation_changed in lines 11, 21 and 31.

src/osm2pgsql.cpp
```diff
@@ -63,7 +63,17 @@ int main(int argc, char *argv[])
...
+            std::any_of(outputs.cbegin(), outputs.cend(),
+                        [](std::shared_ptr<output_t> const &output) {
+                            return output->need_forward_dependencies();
+                        });
+
+        auto dependency_manager = std::unique_ptr<dependency_manager_t>(
+            need_dependencies ? new full_dependency_manager_t{middle.get()}
+                              : new dependency_manager_t{});
+
+        osmdata_t osmdata{dependency_manager.get(), middle, outputs};
```
lonvia, line 76: Any particular reason not to hand ownership over to osmdata? Or rather move the code into the osmdata constructor?

src/dependency-manager.hpp
```diff
@@ -0,0 +1,161 @@
...
+    template <typename TOutputIterator>
+    void get_pending_relation_ids(TOutputIterator &&it) {
+        osmid_t id;
+        while (id_tracker::is_valid(id = m_rels_pending_tracker.pop_mark())) {
+            *it++ = id;
+        }
+    }
+
+private:
+    middle_t* m_object_store;
```
lonvia, line 155: make that a shared ptr?

joto: Bare pointer is all we need here and we know it outlives the dependency manager.

src/osm2pgsql.cpp
joto: Currently the dependency manager doesn't need any configuration. But if it does in the future, so I think it is better created here where the options are available.

Moving into osmdata would be possible, just as with middle and outputs. That's certainly something to consider, but I'd rather do all of them or none of them.

src/dependency-manager.hpp
joto: fixed

src/dependency-manager.cpp
joto: fixed

src/dependency-manager.hpp
joto: I added a commit that uses a share_ptr here.

src/osm2pgsql.cpp
joto: I added a commit that moves the dependency manager into osmdata.

PR #1169, opened 2020-05-13
Remove verbose comments and else after return

joto:


mmd-osm:
I wouldn't say the code is easier to read and understand now, quite the contrary.

src/lua-utils.cpp
```diff
@@ -153,29 +153,26 @@ static int pcall_error_traceback_handler(lua_State *lua_state)
...
-                          "__tostring") && // does it have a metamethod
-            lua_type(lua_state, -1) == LUA_TSTRING) { // that produces a string?
-            return 1;                                 // that is the message
-        } else {
-            msg = lua_pushfstring(lua_state, "(error object is a %s value)",
-                                  luaL_typename(lua_state, 1));
+    if (msg == nullptr) {
+        if (luaL_callmeta(lua_state, 1, "__tostring") &&
+            lua_type(lua_state, -1) == LUA_TSTRING) {
+            return 1;
```
mmd-osm, line 159: This part of the code is fairly cryptic, not sure if it's really a good idea to entirely remove all comments here. You can do me a favor and add a link to https://github.com/lua/lua/blob/master/lua.c so people who want to understand this code know where to look for more information.

joto: Comments have a tendency to get out of date. Osm2pgsql is full of comments that are wrong, sometimes commenting something that isn't there any more.

SomeoneElseOSM: Whilst I certainly wished for more explanatory comments that last time I was looking at osm2pgsql's lua interface, I'm not convinced that the ones being removed here are a great loss.  What would have been really useful though would be an actual document describing how it works (and how to expand it).

PR #1181, opened 2020-05-21
README.md translates for Korean

suuubbb:
I translated README.md to Korean because I wanted Korean contributors to contribute the project easily.
 The README.kor.md has been added to the master branch, so please check it!

### DIFF
--- a/README.kor.md
+++ b/README.kor.md
@@ -1,0 +1,200 @@
+# osm2pgsql #
+
+Osm2pgsql은 openstreetmap 데이터를 지도에 렌더링, Nominatim 또는 일반적인 분석을 이용한 geocoding을 하면서 
+지도에 rendering하는 것과 같은 애플리케이션에 적합한 PostgreSQL과 PostGIS database로 로딩하기 위한 도구이다.
+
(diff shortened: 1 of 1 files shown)


lonvia:
Thank you this work. Unfortunately, I have to decline this because there is no way for us to keep this up-to-date. There is a osm2pgsql page [in the OSM wiki](https://wiki.openstreetmap.org/wiki/Osm2pgsql). How about adding the Korean translation there?

kennykb:
If @suubbb is willing to maintain the translation, keeping it up to date wouldn't be an insurmountable problem. I've set up this sort of thing on other open-source projects. It can be handled manually, or if you want to get fancy, by a commit script. The idea is that when you update the files in question, you open tickets for the translators, and the translators create PR's for the changes. It's not usually a horrible workload once the procedures are in place. It's not a catastrophe if a translation is somewhat out of sync as long as someone updates it eventually, ideally before tagging-and-bagging a release.

suuubbb:
> Thank you for this work. Unfortunately, I have to decline this because there is no way for us to keep this up-to-date. There is a osm2pgsql page [in the OSM wiki](https://wiki.openstreetmap.org/wiki/Osm2pgsql). How about adding the Korean translation there?


I think the up-to-date system can be helped by many people besides me.

Of course, I can up-to-date translate.

if this is not good to keep master branch, make new branch to merge is okay.

PR #1184, opened 2020-05-23
Clarify docs regarding tablespace options ignored in flex backend

### LINKED ISSUE #1182: Flex backend question: does flex backend also ignore tablespace for SLIM data?
According to the flex backend documentation (https://github.com/openstreetmap/osm2pgsql/blob/master/docs/flex.md), flex backend ignores the table space command line options for osm2pgsql. However, there are two types of tablespaces that can be defined: SLIM (--tablepace-slim-data/index) and the output data (--tablespace-main-data/index).

Does the new flex backend also ignore the **--tablepace-slim-data/index** command line options, and if so, am I right SLIM data will end up in PostgreSQL default tablespace always with flex backend?

joto:
Fixes #1182.

### DIFF
--- a/docs/flex.md
+++ b/docs/flex.md
@@ -271,9 +271,10 @@ flex backend, because they don't make sense in that context:
   option.)
 * `-G|--multi-geometry` (Use the `multi` option on the geometry transformation
   instead.)
-* The command line options to set the tablespace are ignored by the flex
-  backend, instead use the `data_tablespace` or `index_tablespace` options
-  when defining your table.
+* The command line options to set the tablespace (`-i|--tablespace-index`,
+  `--tablespace-main-data`, `--tablespace-main-index`) are ignored by the flex
+  backend, instead use the `data_tablespace` or `index_tablespace` options when
+  defining your table.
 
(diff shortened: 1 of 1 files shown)


mboeringa:
@joto, you don't mention the actual SLIM table space options in this latest commit:  

From the https://www.volkerschatz.com/net/osm/osm2pgsql-usage.html webpage:
> --tablespace-slim-data tablespacename
>     Store the slim mode tables in the given tablespace. 
> --tablespace-slim-index tablespacename
>     Store the indices of the slim mode tables in the given tablespace. 

Adding this in as well is probably best to truly clarify it.

mboeringa:
I realize, considering **--slim** is still doing what it is supposed to do, that my question I posed in https://github.com/openstreetmap/osm2pgsql/issues/1182 is likely already answered.

That said, a few words about SLIM tables in the flex config, and explicitely stating the **--tablespace-slim-data** tablespacename and **--tablespace-slim-index** tablespacename options of osm2pgsql are still valid option combined with **--slim** when using the flex backend, might help reducing possible confusion for new users.

PR #1196, opened 2020-05-30
Rename file format option and simplify default (auto) case

joto:
* Rename option "input_reader" to "input_format".
* Also rename a parameter from "fmt" to "format" because "fmt"
  is used as namespace for the format library we are using.
* Change default setting for autodetecting the file format from
  "auto" to simply the empty string. This matches the way the
  underlying osmium library does it.

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -488,7 +488,7 @@ options_t::options_t(int argc, char *argv[]) : options_t()
             enable_multi = true;
             break;
         case 'r':
-            input_reader = optarg;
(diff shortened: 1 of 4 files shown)


lonvia:
`auto` is an explicit name that can be used for the option. This is also stated in [the help string](https://github.com/openstreetmap/osm2pgsql/blob/master/src/options.cpp#L198).

joto:
Just pushed a rebase onto current master.

PR #1197, opened 2020-05-30
Fix in flex output: Updating tables without geometry should work

joto:


joto:
Just pushed a rebase onto current master.

src/output-flex.cpp
```diff
@@ -1115,10 +1115,19 @@ void output_flex_t::delete_from_table(table_connection_t *table_connection,
...
     assert(table_connection);
     auto const id = table_connection->table().map_id(type, osm_id);
-    auto const result = table_connection->get_geom_by_id(type, id);
-    if (m_expire.from_result(result, id) != 0) {
-        table_connection->delete_rows_with(type, id);
+
+    if (table_connection->table().has_geom_column()) {
+        auto const result = table_connection->get_geom_by_id(type, id);
+
+        if (m_expire.from_result(result, id) == 0) {
```
lonvia, line 1122: The short-cut here should now apply when result.num_tuples == 0. If you test the result of `from_result` then you don't catch the case where the expiry is disabled.

joto:
I have re-worked the PR and rebased it.

PR #1198, opened 2020-05-30
Split up processing into functions for each stage

joto:
Split up the processing in osmdata_t into functions for each stage. This
better documents what is happening.


joto:
I have started to write some documentation that described the multi-stage processing. It will be in a later PR.

PR #1209, opened 2020-06-12
Delete migrations.md

mmd-osm:
I'm proposing to remove this file as it hasn't been updated since quite some time and is probably only of historic interest.

All relevant information on incompatible changes between versions can be found in the release notes: https://github.com/openstreetmap/osm2pgsql/releases


lonvia:
The release notes are hidden away in commit message of git tags. That's the last place somebody would look for migrations.

The migration file is up to date. There haven't been any breaking database changes since 0.93. If we do such changes, the file should be updated.

mmd-osm:
Ok, I thought this information was somehow outdated based on the 0.94.0 release notes, which mentioned a need for a reimport for the slim table changes:

> Release 0.94.0, a new stable branch
> This is the last release which will support old-style multipolygons.
> Upgrading to this version requires a reimport for the slim table changes.

lonvia:
The slim table change is mentioned for version 0.93. This has something to do with the odd/even release numbering schema we used until recently. Might be a good idea to simply change that to 0.94 in migrations.md as there never was an actual 0.93 release.

PR #1210, opened 2020-06-12
Remove the old SVN Changelog

mmd-osm:
An older commit to remove the Changelog has been reverted in https://github.com/openstreetmap/osm2pgsql/commit/22a874e08d505934a3d29050266f71ee4c6199be with the comment "Makefile.am appears to require a ChangeLog file to work correctly"

As we're no longer using autotools, I think it should be ok to finally remove this file now.

Also, `git log` shows the same information, and is up-to-date.


mmd-osm:
By the way, @pnorman created the original commit to delete this file: https://github.com/openstreetmap/osm2pgsql/commit/89e67cf468918f6dd53de0c20f0831de599efbcb

PR #1215, opened 2020-06-16
Fix multistage processing (take 2)

joto:
(This is a replacement for #1208.)

The multi-stage processing in the flex output has some problems when
updating data, for instance when relations are deleted after their
member ways use their tags.

This commit reorganizes how the multi-stage processing is done to
address these problems.

1. The way "marking" of objects is done is now different. Instead of
   doing this in the process_* Lua functions a new Lua function
   "mark_relation_members()" is introduced. This will be called not only
   for all new relations but also for deleted relations, or when a relation
   changed, for the old relation. This is needed so that we also mark and
   then re-create way entries in the database that used to depend on a
   parent relations tags, but don't do that any more. This function
   must return the way ids that need to be re-processed in stage 2.
2. We remove the Lua mark() function. Instead the return value of
   "mark_relation_members()" is used to mark way ids. Marking of nodes
   and relations isn't supported. It was never possible to mark nodes
   anyway. And there is no use case I am aware of currently that needs
   marking relations. Both can be reintroduced later when we have a
   better idea how to handle them. For the time being we concentrate on
   the important use case where member ways of relations are handled
   specially.
3. This introduces a new processing stage 1c:
    * stage 1a: Read input file and process all objects in it.
    * stage 1b: Process dependent objects of objects from 1a (ie
      changed nodes trigger changes in ways with those nodes, changes
      in all objects potentially trigger changes in parent relations).
    * stage 1c: Process dependent relations of objects marked during
      stage 1a/1b (this one is new).
    * stage 2: Process objects marked in stage 1a or 1b.
4. New Lua helper function osm2pgsql.way_member_ids() that returns
   the ids of all way members of the specified relation. This is often
   needed in mark_relation_members(). If you need more complex processing,
   for instance use the member roles to decide which ways need stage 2
   processing, you can still write your own loop.

### DIFF
--- a/docs/flex.md
+++ b/docs/flex.md
@@ -36,14 +36,14 @@ The following functions are defined:
 * `osm2pgsql.define_table(options)`: Define a table. This is the more flexible
   function behind all the other `define_*_table()` functions. It gives you
   more control than the more convenient other functions.
-* `osm2pgsql.mark_way(id)`: Mark the OSM way with the specified id. This way
-  will be processed (again) in stage 2.
 
 You are expected to define one or more of the following functions:
 
-* `osm2pgsql.process_node()`: Called for each node.
-* `osm2pgsql.process_way()`: Called for each way.
-* `osm2pgsql.process_relation()`: Called for each relation.
+* `osm2pgsql.process_node()`: Called for each new or changed node.
+* `osm2pgsql.process_way()`: Called for each new or changed way.
+* `osm2pgsql.process_relation()`: Called for each new or changed relation.
+* `osm2pgsql.select_relation_members()`: Called for each deleted or added
+  relation. See below for more details.
 
 Osm2pgsql also provides some additional functions in the
 [lua-lib.md](Lua helper library).
@@ -76,7 +76,7 @@ stored as is, relation ids will be stored as negative numbers.
 With the `osm2pgsql.define_table()` function you can also define tables that
 * don't have any ids, but those tables will never be updated by osm2pgsql
 * take *any OSM object*, in this case the type of object is stored in an
-  additional column.
+  additional `char(1)` column.
 * are in a specific PostgresSQL tablespace (set option `data_tablespace`) or
   that get their indexes created in a specific tablespace (set option
   `index_tablespace`).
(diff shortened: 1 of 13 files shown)


src/osmdata.cpp
```diff
@@ -389,16 +393,39 @@ progress_display_t osmdata_t::process_file(osmium::io::File const &file,
...
 
-void osmdata_t::process_stage1b() const
+void osmdata_t::process_stage1bc() const
 {
-    if (m_dependency_manager->has_pending()) {
-        multithreaded_processor proc{m_conninfo, m_mid, m_outs,
-                                     (std::size_t)m_num_procs};
+    if (!m_dependency_manager->has_pending()) {
+        return;
+    }
```
lonvia, line 400: This only excludes stage 1b, not stage 1c.

src/osmdata.cpp
```diff
@@ -389,16 +393,39 @@ progress_display_t osmdata_t::process_file(osmium::io::File const &file,
...
+    proc.merge_expire_trees();
+
+    // If none of the outputs wants to do stage 2 processing, we can skip
+    // stage 1c.
+    if (std::none_of(m_outs.cbegin(), m_outs.cend(),
+                     [](std::shared_ptr<output_t> const &out) {
+                         return out->has_stage2_processing();
+                     })) {
+        return;
+    }
```
lonvia, line 417: If there is no stage2 processing, get_marked_way_ids()  simply is empty. So the loop below seems enough.

src/output-flex.cpp
```diff
@@ -1055,15 +1039,104 @@ void output_flex_t::pending_way(osmid_t id)
...
+            "field, but it isn't an array table"};
+    }
+
+    // Iterate over the 'ways' table to get all ids...
+    lua_pushnil(lua_state());
+    while (lua_next(lua_state(), -2) != 0) {
+        if (!lua_isnumber(lua_state(), -2)) {
+            throw std::runtime_error{
+                "Table returned from mark_relation_members() contains 'ways' "
+                "field, but it isn't an array table"};
```
lonvia, line 1093: ...array table of numbers

src/output-flex.cpp
```diff
@@ -1185,6 +1264,7 @@ void output_flex_t::way_delete(osmid_t osm_id)
 
 void output_flex_t::relation_delete(osmid_t osm_id)
 {
+    mark_relation_members(osm_id);
```
lonvia, line 1267: Do we call the output_delete before deleting from middle? I suspect not.

docs/flex.md
```diff
@@ -35,14 +35,14 @@ The following functions are defined:
...
 
 You are expected to define one or more of the following functions:
 
-* `osm2pgsql.process_node()`: Called for each node.
-* `osm2pgsql.process_way()`: Called for each way.
-* `osm2pgsql.process_relation()`: Called for each relation.
+* `osm2pgsql.process_node()`: Called for each new or changed node.
+* `osm2pgsql.process_way()`: Called for each new or changed way.
+* `osm2pgsql.process_relation()`: Called for each new or changed relation.
+* `osm2pgsql.mark_relation_members()`: Called for each deleted or added
```
lonvia, line 44: I'm still not quite happy with the name. The function doesn't mark anything, it returns members to be reprocessed.

src/osmdata.cpp
joto: Okay. Fixed.

src/output-flex.cpp
joto: The check is for the table key not the value. The table key is a number in "array tables", so we are checking the table is an array, not a hash.

src/output-flex.cpp
joto: I believe we do.

docs/flex.md
joto: Now called `select_relation_members()`. Not perfect but maybe a bit better.

src/osmdata.cpp
joto: Not applicable any more.

joto:
I think I have addressed all concerns in the new rebased version of this commit.

src/output-flex.cpp
```diff
@@ -1055,33 +1043,129 @@ void output_flex_t::pending_way(osmid_t id)
...
+    if (!m_process_relation) {
+        return;
+    }
+
+    if (!m_mid->relation_get(id, m_rels_buffer)) {
+        return;
+    }
     auto const &relation = m_rels_buffer.get<osmium::Relation>(0);
 
+    m_disable_add_row = true;
```
lonvia, line 1164: Shouldn' t that be set to false again after the function returns? I' d expect that add_row() now does nothing in stage2 as well (although that is obviously wrong because the tests pass).

joto: Threading confusion strikes again. This is only set in the sub-threads, while the stage2 runs in the main thread. I have changed the code anyway and disabled it after the call again, to avoid this confusion and make it more robust if we decide to change in what thread things run.

joto:
One change, rebased to current master and force pushed.

PR #1227, opened 2020-06-20
Make clustering of tables by geometry optional

joto:


src/flex-table.cpp
```diff
@@ -265,26 +297,7 @@ void table_connection_t::stop(bool updateable, bool append)
...
-                                              table().geom_column().name()));
-
-            m_db_connection->exec("CREATE TRIGGER \"{0}_osm2pgsql_valid\""
-                                  " BEFORE INSERT OR UPDATE"
-                                  " ON {1}"
-                                  " FOR EACH ROW EXECUTE PROCEDURE"
-                                  " {0}_osm2pgsql_valid();"_format(
-                                      table().name(), table().full_name()));
-        }
+        create_trigger_for_geom_check();
```
lonvia, line 300: Are you sure that isn't called twice when clustering is disabled?

joto: You are right. This call is in the wrong position. New version of PR coming once #1229 is through.

lonvia:
I would prefer to leave this after the next release. We might want to have a discussion how to handle the post-processing of the table in general first.

PR #1232, opened 2020-06-25
The "compatible.lua" config is now much more similar to c transform

joto:
There are still some minor differences, but this should be usable as a
replacement. Also supports several of the old command line options
functionality through settings in the config file. This should give
users a good migration path from original pgsql output with C transform
to the flex output.

### DIFF
--- a/flex-config/compatible.lua
+++ b/flex-config/compatible.lua
@@ -1,8 +1,51 @@
 
 -- This configuration for the flex backend tries to be compatible with the
--- original pgsql backend.
+-- original pgsql c-transform backend. There might be some corner cases but
+-- it should mostly do exactly the same.
 
--- Objects with any of the following keys will be treated as polygon
+-- Set this to true if you were using option -K|--keep-coastlines.
+local keep_coastlines = false
+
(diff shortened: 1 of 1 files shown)


mboeringa:
@joto , nice to see this taking shape! This should indeed help with a smooth transition for many, while still allowing the nice new features of the flex backend in development. Good work.

Just one question: could you elaborate a bit on the differences with the work Paul Norman did on the flex backend LUA file for openstreetmap-carto:

https://github.com/gravitystorm/openstreetmap-carto/blob/flex/master/openstreetmap-carto.lua

Of course, that one includes a new "route" table, which is a nice new feature that helps in displaying or working with OSM routes, while this "compatible.lua" does not have that, but adds some of the familiar osm2pgsql command line options in a LUA implementation? Any other things to be a aware off and how these to LUA configs relate, and the respective purposes?

joto:
Pauls configuration is trying to be compatible with the old openstreetmap-carto style, the one in this PR is trying to be compatible with the standard style of osm2pgsql that you get when you don't specify a style at all.

lonvia:
I don't know if it's worth keeping the routes special handling. It's buggy in the C version already.

PR #1233, opened 2020-06-25
Add docs about using `create_only` columns for postprocessed data

joto:


pnorman:
This particular example would probably be best done with a `GENERATED` column. Do we have a better example use?

joto:
a) `GENERATED` columns are only supported since PostgreSQL 12.
b) There is currently no way to create `GENERATED` columns with the flex backend. I thought about adding it, but I am unsure about how to best do this given a). And adding more and more column options might be a rabbit hole I don't want to go down at the moment.

So basically this documents a usable work-around for something that we might support in the future with `GENERATED`, but for the time being this should work well enough and once people do these kinds of things, we'll have a better idea whether and how we want to add proper support in osm2pgsql for this kind of thing.

PR #1234, opened 2020-06-28
Update protozero and libosmium in contrib to newest versions

joto:


ldeffenb:
Is it true that this pull request will fix the issue with the recent complex relation that stalls updates?

joto:
@ldeffenb Yes, it will. But you should use the released version [1.2.2](https://github.com/openstreetmap/osm2pgsql/releases/tag/1.2.2)

PR #1238, opened 2020-06-30
update version number 1.2.0 => 1.2.2

mtmail:
to match the latest release https://github.com/openstreetmap/osm2pgsql/releases/tag/1.2.2


lonvia:
We don't usually do that because 1.2.2 and master are two completely different branches.

PR #1241, opened 2020-07-02
Extended schema support to pgsql backend.

zerebubuth:
There was already support in progress for schemas in the flex backend, which is likely to replace the pgsql backend at some point. Until then, I've found it useful to be able to import data directly into a schema.


lonvia:
Not so quick. While I value @joto's opinions, he is not the one with the power over the merge button around here. ;)

It is true that we need to reconsider the approach to configuration options. But that is not going to happen today or next week and we need to be practical here. This is a feature that has been requested quite often and I'm on record stating that I take a pull request that implements this as an extra option. This is still true. On a quick look into it, this PR meets the requirements.

So I'll reopen the PR in the hope that we might get it in before the next release (which is rather imminent). I still need to do a detailed review later. For the moment, two of the points that @joto mentions, I'd definitely liked to see addressed, too: the documentation needs to be adapted and we should have at least one test. The test could consist of an additional regression test or do as @joto says. @zerebubuth I'd really appreciate if you can look into this.

lonvia:
Closing in favour of #1288.

PR #1249, opened 2020-07-08
Add more way tests for flex output

joto:


tests/test-output-flex-way-change.cpp
```diff
@@ -0,0 +1,274 @@
...
+    CHECK(1 == conn.get_count("osm2pgsql_test_t1"));
+    CHECK(1 == conn.get_count("osm2pgsql_test_t2"));
+    CHECK(3 == conn.get_count("osm2pgsql_test_tboth"));
+    CHECK(1 == conn.get_count("osm2pgsql_test_tboth",
+                              "way_id = 10 AND rel_ids = '{30}'"));
+
+    options.append = true;
+
+    REQUIRE_NOTHROW(db.run_import(
+        options, "r30 v2 dV Tt=ag Mw10@,w11@,w12@mark,w13@,w14@mark"));
```
lonvia, line 169: That boils to a relation change. Should already be covered by the relation change tests and can be removed here.

tests/test-output-flex-way-change.cpp
```diff
@@ -0,0 +1,274 @@
...
+    CHECK(1 == conn.get_count("osm2pgsql_test_t1"));
+    CHECK(1 == conn.get_count("osm2pgsql_test_t2"));
+    CHECK(3 == conn.get_count("osm2pgsql_test_tboth"));
+    CHECK(1 == conn.get_count("osm2pgsql_test_tboth",
+                              "way_id = 10 AND rel_ids IS NULL"));
+
+    options.append = true;
+
+    REQUIRE_NOTHROW(db.run_import(
+        options, "r30 v2 dV Tt=ag Mw10@mark,w11@,w12@mark,w13@,w14@mark"));
```
lonvia, line 199: dto

tests/test-output-flex-way-change.cpp
```diff
@@ -0,0 +1,274 @@
...
+    SECTION("change node to make way invalid (n11 same location as n10)")
+    {
+        REQUIRE_NOTHROW(db.run_import(options, "n11 v2 dV x10.0 y10.0"));
+    }
+
+    CHECK(1 == conn.get_count("osm2pgsql_test_t1"));
+    CHECK(1 == conn.get_count("osm2pgsql_test_t2"));
+    CHECK(2 == conn.get_count("osm2pgsql_test_tboth"));
+    CHECK(0 == conn.get_count("osm2pgsql_test_tboth",
+                              "way_id = 10 AND rel_ids = '{30}'"));
```
lonvia, line 243: Test for "way_id = 10" only, it's broader.

tests/test-output-flex-way-change.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    REQUIRE_NOTHROW(db.run_import(options, data()));
+
+    auto conn = db.db().connect();
+
+    CHECK(1 == conn.get_count("osm2pgsql_test_t1"));
+    CHECK(1 == conn.get_count("osm2pgsql_test_t2"));
+    CHECK(2 == conn.get_count("osm2pgsql_test_tboth"));
+    CHECK(0 == conn.get_count("osm2pgsql_test_tboth",
+                              "way_id = 10 AND rel_ids = '{30}'"));
```
lonvia, line 262: dto

tests/test-output-flex-way-change.cpp
```diff
@@ -0,0 +1,274 @@
...
+    testing::data_t data{tdata};
+    data.add({"w10 v1 dV Tt1=yes,t2=yes,tboth=yes Nn10",
+              "r30 v1 dV Tt=ag Mw10@mark,w11@,w12@mark,w13@,w14@mark"});
+
+    REQUIRE_NOTHROW(db.run_import(options, data()));
+
+    auto conn = db.db().connect();
+
+    CHECK(1 == conn.get_count("osm2pgsql_test_t1"));
+    CHECK(1 == conn.get_count("osm2pgsql_test_t2"));
```
lonvia, line 259: General comment on T2 checks: I'm not convinced the test is sufficient. We should test that for all ways we have the right relations.

tests/test-output-flex-way-relation-add.cpp
```diff
@@ -0,0 +1,379 @@
...
+    "w11 v1 dV Tt1=yes Nn12,n13",
+    "w12 v1 dV Tt2=yes Nn14,n15",
+    "w13 v1 dV Ttboth=yes Nn16,n17",
+    "w14 v1 dV Ttboth=yes Nn18,n19",
+    "w15 v1 dV Tt=ag Nn17,n19",
+    "r30 v1 dV Tt=ag Mw10@,w11@,w12@mark,w13@,w14@mark"};
+
+// adding relation (marked)
+
+TEST_CASE("test way: add relation with way in t1 (marked)")
```
lonvia, line 30: Comment should reflect that we are testing relation change as well.

tests/test-output-flex-way-change.cpp
joto: fixed

tests/test-output-flex-way-change.cpp
joto: fixed

tests/test-output-flex-way-relation-add.cpp
joto: fixed

tests/test-output-flex-way-change.cpp
joto: okay, have remove it

tests/test-output-flex-way-change.cpp
joto: also removed

tests/test-output-flex-way-change.cpp
joto: i have added some missing tests for t2 relation.

lonvia:
Travis now complains that the log is getting too long. I would suggest to disable verbose logging for the multi-version test.

tests/test-output-flex-way-relation-add.cpp
```diff
@@ -0,0 +1,387 @@
...
+{
+    options_t options = testing::opt_t().slim().flex(conf_file);
+
+    testing::data_t data{tdata};
+
+    data.add("w10 v1 dV Tt1=yes,t2=yes Nn10,n11");
+
+    SECTION("no relation") {}
+    SECTION("relation without w10") { data.add("r32 v1 dV Tt=ag Mw15@"); }
```
lonvia, line 40: We are missing the indirect dependencies through members here. r32 should have the other ways as members, when it gets changed. This should then cover the last error case we have seen (the one that resulted in #1244). Unfortunately that means that we need separate functions. :(

joto:
We can just use `ctest --output-on-failure` instead of `ctest -VV`.

Issue #1097, opened 2020-03-31
Deprecate use of negative IDs


joto:
Looking at the code I can see that osm2pgsql has provisions for the use of negative node/way/relation IDs in some places. But I am unsure whether this works correctly. Maybe it does in some circumstances and not in others. One problem is the use of negative IDs in the database for areas created from multipolygon relations. This will clash with negative way IDs.

Catering for negative IDs makes the code more complex and at least some of the code looks like it uses extra memory to store those IDs that almost no user will have. Do we actually need this and want to support it? What are the use cases? Are there uses that can't be replaced by using `osmium renumber` to first create a file without negative IDs?

migurski:
Negative IDs are commonly used when handling external data that doesn't come from OSM.org but needs to be stored alongside public data with positive IDs for rendering or conflation purposes. It’s helpful to have an integer range that’s known to be free of legitimate OSM.org entities.

pnorman:
> But I am unsure whether this works correctly

I don't think they do.

-----



> Negative IDs are commonly used when handling external data that doesn't come from OSM.org but needs to be stored alongside public data with positive IDs for rendering or conflation purposes. It’s helpful to have an integer range that’s known to be free of legitimate OSM.org entities.

Are you doing this successfully with osm2pgsql? Have you evaluated if it works when the absolute value of an id conflicts? Does it work with updates?

migurski:
I’ve not done this with osm2pgsql, but I have used negative ranges in pipelines with other toolchains. I’ve seen the pattern at just about every OSM-using engineering org I’ve been a part of, making me think it’s either common or obvious enough to support.

mmd-osm:
I've seen people using some ridiculously large offsets to handle custom data, which sometimes triggers unexpected issues for some tools (e.g. see [this post](https://gis.stackexchange.com/questions/273204/what-is-the-max-value-for-node-ids-in-overpass)). Working with carefully chosen positive offsets should work in general, but requires careful end-to-end testing.

Nakaner:
@joto This issue about deprecating support of negative IDs in the input files, not in the "osm_id" column of the tables?

joto:
@Nakaner Yes, its about the input files.

Nakaner:
If it is about input files only, I do not have any objections against the deprecation.

mmd-osm:
> Do we actually need this and want to support it? What are the use cases?

[...]

PR #1251, opened 2020-07-15
Check for negative object ids in input files and output a warning

joto:
Negative ids for OSM objects never worked correctly in all cases and
will not be supported in future versions of osm2pgsql.

See #1097

@pnorman Can you please review?

### DIFF
--- a/src/input-handler.cpp
+++ b/src/input-handler.cpp
@@ -9,8 +9,24 @@ input_handler_t::input_handler_t(osmium::Box const &bbox, bool append,
 : m_data(osmdata), m_bbox(bbox), m_append(append)
 {}
 
+void input_handler_t::negative_id_warning()
+{
+    fmt::print(
+        stderr,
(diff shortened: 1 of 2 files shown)


src/input-handler.cpp
```diff
@@ -9,8 +9,23 @@ input_handler_t::input_handler_t(osmium::Box const &bbox, bool append,
...
 {}
 
+void input_handler_t::negative_id_warning()
+{
+    fmt::print(stderr,
+               "\nWARNING: The input file contains at least one object with "
+               "a negative id.\n         Negative ids are not properly "
+               "supported in osm2pgsql (and never\n         were). They will "
+               "not work in future versions at all. You can use\n"
+               "         the osmium tool to 'renumber' your file.\n\n");
```
pnorman, line 19: Can we reformat this text? If necessary, the max line length can be overridden for this part of the file.

joto: I have forced pushed a new version with better formatting.

Issue #1113, opened 2020-04-10
Dead links (404) in documentation


mboeringa:
Hi,

Minor issue, but I noticed there is a couple of dead links to a no longer accessible blog post:

http://paulnorman.ca/blog/2014/03/osm2pgsql-and-hstore/

Documentation pages affected:
https://github.com/openstreetmap/osm2pgsql/blob/master/docs/usage.md#output-data-options
https://github.com/openstreetmap/osm2pgsql/blob/master/docs/analysis.md

PR #1252, opened 2020-07-16
Fix link for performance analysis for adding --multi-geometry and --hstore

hiddewie:
The www is needed somehow.


joto:
There is another link in docs/analysis.md .

This fixes #1113

hiddewie:
@joto Done, I missed that one :+1:

PR #1263, opened 2020-08-04
Cleanup reprojection

joto:
Some steps cleaning up the reprojection code. See individual commits.


src/expire-tiles.cpp
```diff
@@ -122,6 +122,21 @@ int expire_tiles::normalise_tile_x_coord(int x)
...
 }
 
+/**
+ * Converts from target coordinates to tile coordinates.
+ *
+ * The zoom level for the coordinates is explicitly given in the
+ * parameter map_width.
+ */
+static void coords_to_tile(reprojection const &projection, double *tilex,
+                           double *tiley, double lon, double lat, int map_width)
```
lonvia, line 132: If you make that a member of expire_tiles, you can save the projection and map_width parameter. And while you are at it, I would suggest to move output parameters to the end of the parameter list.

src/reprojection.cpp
```diff
@@ -1,74 +1,59 @@
...
 
 #include "reprojection.hpp"
 
-/** must match expire.tiles.c */
-#define EARTH_CIRCUMFERENCE 40075016.68
-
 namespace {
 
-void latlon2merc(double *lat, double *lon)
+Coordinates lonlat2merc(Coordinates coords)
```
lonvia, line 7: Shouldn't it rather be coord2merc now?

src/reprojection.hpp
```diff
@@ -35,28 +45,22 @@ class reprojection
...
 
     /**
-     * Converts coordinates from target projection to tile projection (EPSG:3857)
+     * Converts coordinates from target projection to tile projection
+     * (EPSG:3857)
      *
-     * Do not confuse with coords_to_tile which does *not* calculate coordinates in the
-     * tile projection, but tile coordinates.
+     * Do not confuse with coords_to_tile which does *not* calculate
+     * coordinates in the tile projection, but tile coordinates.
```
lonvia, line 55: Remove this comment now that coords_to_tile has gone away?

src/reprojection.cpp
joto: It still converts coordinates from lon/lat format to Mercator. The `Coordinates` just holds two doubles and can be used by many different coordinate systems.

src/expire-tiles.cpp
joto: I added a commit with this.

src/reprojection.hpp
joto: Adressed in one of the commits.

PR #1270, opened 2020-09-02
Middle cleanup queries

joto:
Cleaning up some unused queries and such in middle. See individual commits.


lonvia:
Turns out that Nominatim needs the array indexes on planet_osm_rels after all. It does a lookup by member. So this causes a bit of a regression in Nominatim, see https://github.com/osm-search/Nominatim/issues/1978#issuecomment-701743883. However, I think this is better fixed on the Nominatim side where we can just create the indexes we need.

PR #1271, opened 2020-09-03
Draft bucket index

joto:
This is a draft PR of an improved version of PR #1058. This now allows both the old and the new index to be created/used. Also adds some documentation.


joto:
Close in favour of finished PR #1275.

PR #1277, opened 2020-10-01
Set projection in flex output in column config

joto:
The projection set on the command line with -l|--latlong, -m|--merc, or
-E|--proj is now ignored in the flex output. Instead you can set it
individually on all geometry columns using the `projection` option which
either takes an EPSG code or a few commonly understood string values
such as "merc", "latlong", or "WGS84".

This also makes the flex output ignore the --reproject-area option.
The area is now calculated in web mercator by default and you can
change this by setting the `projection` option on the area column.
(This currently only allows setting 4326 and 3857, other projections
will not work. We have to fix this later. This is due to the way
the area is calculated from the WKB geometry.)

The osm2pgsql.srid variable is not set any more in the Lua config.

So there are some breaking changes here for the flex output, which
we can still do as it is marked experimental.

With this change the command line options -l, -m, -E, and
--reproject-area, that can be regarded as belonging to the style are not
used any more in the flex output. This completes the move away from
command line option to config file for the flex output.

### DIFF
--- a/flex-config/geometries.lua
+++ b/flex-config/geometries.lua
@@ -4,27 +4,20 @@
 -- at this file. This file will show some options around geometry processing.
 -- After you have understood this file, go on to "data-types.lua".
 
--- The projection configured on the command line is available in Lua:
-print('osm2pgsql srid: ' .. osm2pgsql.srid)
-
--- You can use it, for instance, to figure out the maximum lengths for line
--- geometries, see below for how this is used.
-if osm2pgsql.srid == 3857 then
-    max_length = 100000
-else
-    max_length = 1
-end
-
 local tables = {}
 
 tables.pois = osm2pgsql.define_node_table('pois', {
     { column = 'tags', type = 'hstore' },
+    -- Create a geometry column for point geometries. The geometry will be
+    -- in web mercator, EPSG 3857.
     { column = 'geom', type = 'point' },
 })
 
 tables.ways = osm2pgsql.define_way_table('ways', {
     { column = 'tags', type = 'hstore' },
-    { column = 'geom', type = 'linestring' },
+    -- Create a geometry column for linestring geometries. The geometry will
+    -- be in latlong (WGS84), EPSG 4326.
+    { column = 'geom', type = 'linestring', projection = 4326 },
 })
 
 tables.polygons = osm2pgsql.define_area_table('polygons', {
@@ -32,7 +25,9 @@ tables.polygons = osm2pgsql.define_area_table('polygons', {
     { column = 'geom', type = 'geometry' },
     -- The 'area' type is used to store the calculated area of a polygon
     -- feature. This can be used in style sheets to only render larger polygons
-    -- in small zoom levels.
+    -- in small zoom levels. This will use the area in web mercator projection,
+    -- you can set 'projection = 4326' to calculate the area in WGS84. Other
+    -- projections are currently not supported.
     { column = 'area', type = 'area' },
 })
 
@@ -151,7 +146,7 @@ function osm2pgsql.process_way(object)
         -- multiple rows that are identical except for the geometry.
         tables.ways:add_row({
             tags = object.tags,
-            geom = { create = 'line', split_at = max_length }
+            geom = { create = 'line', split_at = 1 }
         })
     end
 end
--- a/src/flex-table-column.hpp
+++ b/src/flex-table-column.hpp
@@ -2,6 +2,7 @@
 #define OSM2PGSQL_FLEX_TABLE_COLUMN_HPP
 
 #include "format.hpp"
+#include "reprojection.hpp"
 
 #include <cstdint>
 #include <string>
@@ -85,9 +86,13 @@ public:
 
     void set_create_only(bool value = true) noexcept { m_create_only = value; }
 
-    std::string sql_type_name(int srid) const;
+    void set_projection(char const *projection);
+
+    std::string sql_type_name() const;
     std::string sql_modifiers() const;
-    std::string sql_create(int srid) const;
+    std::string sql_create() const;
+
+    int srid() const noexcept { return m_srid; }
 
 private:
     /// The name of the database table column.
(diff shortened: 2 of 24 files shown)


mboeringa:
@joto,

Nice to see work continuing on osm2pgsql. However, I am slightly confused by this particular pull request, and have a couple of questions and remarks:

- Personally, i think there is little sense in setting an individual projection on each geometry column of every table created. I doubt if there are many people who would desire to have a mixed projection database, unless you foresee demand from people running a dual projection database e.g. latlong 4326 + Web Mercator 3857. Technically, with this pull request, one would be able with flex output to output the same (or different) content in two different tables with different projection in during import session / run of osm2pgsql, something which is clearly impossible in the current osm2pgsql. However, I think this is an edge case, especially considering many users are already complaining about lack of disk space with the ever growing OSM data, and most users will simply choose between either 4326 or 3857 (of course the option output any projection is still highly desirable for country extracts).

- As to area calculation, I would suggest to calculate geodesic area instead of Web Mercator. We all know the severe limitations of Web Mercator in terms of accurate areas and lengths. I actually implemented that for my own renderer in custom Python code, and it works well with PostGIS. I probably don't have to tell you, but it requires something like this:

If the data is in 4326:
`ST_Area(way::geography,<spheroid_option: True/False>)`

If the data is in any other projection:
`ST_Area(ST_Transform(way,4326)::geography,<spheroid_option: True/False>)`

By the way, _I ran into an issue with the spheroid option_, that I think may have been, or is a current bug, in PostGIS: if setting the spheroid option to **False**, thus using a faster **sphere** based calculation, I experienced crashes with data near the international date line. When I switched to **True** to use a **spheroid**, sacrificing some speed but getting more accurate results, the issue went away. I haven't been able to delve deeper into this, but suspect it to be a bug and not a data issue considering the spheroid option works.

- Extending on this, and looking at some the code comments, you write:

" in web mercator, or you can set the `projection` parameter of the column to `4326` to calculate it with WGS84 coordinates."

What exactly do you mean "with WGS84 coordinates". I guess you mean just using the lat/long values "as-is", calculating some quite meaningless value? Or something else, like the geodesic area I suggested?

pnorman:
I haven't done a line by line review, but 👍 to the idea.

joto:
@mboeringa Yes, multiple different projections in the same database are an edge case. Most people will just use web mercator, that's why it is still the default. But some people have asked for it and it is easy to support, so why not?

I am aware that the area calculation in both web mercator and latlong isn't really that great and that there are better options. In this regard, this pull request aims only at supporting in the flex output what was already supported in the old outputs so that people can move to the new flex output without loosing any functionality. We are planning to put a more generic "geometry handling" in place in the future that would allow more options here, but that will take a while.

src/osmium-builder.hpp
```diff
@@ -24,6 +24,8 @@ class osmium_builder_t
       m_writer(m_proj->target_srs())
     {}
 
+    int srid() const noexcept { return m_proj->target_srs(); }
```
lonvia, line 27: Function unused?

src/output-flex.cpp
```diff
@@ -494,12 +494,13 @@ void output_flex_t::write_row(table_connection_t *table_connection,
...
-                    get_options()->reproject_area
-                        ? ewkb::parser_t(geom).get_area<reprojection>(
-                              get_options()->projection.get())
-                        : ewkb::parser_t(geom)
-                              .get_area<osmium::geom::IdentityProjection>();
+                    column.srid() == srid
+                        ? ewkb::parser_t(geom)
+                              .get_area<osmium::geom::IdentityProjection>()
+                        : ewkb::parser_t(geom).get_area<reprojection>(
+                              reprojection::create_projection(srid).get());
```
lonvia, line 503: This looks rather expensive. You create (as in allocate) a reprojection object every time a row is written.

flex-config/geometries.lua
```diff
@@ -4,35 +4,30 @@
...
+    -- Create a geometry column for point geometries. The geometry will be
+    -- in web mercator, EPSG 3857.
     { column = 'geom', type = 'point' },
 })
 
 tables.ways = osm2pgsql.define_way_table('ways', {
     { column = 'tags', type = 'hstore' },
-    { column = 'geom', type = 'linestring' },
+    -- Create a geometry column for linestring geometries. The geometry will
+    -- be in latlong (WGS84), EPSG 3857.
```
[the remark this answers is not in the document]
joto: fixed

src/flex-table-column.hpp
```diff
@@ -85,9 +86,18 @@ class flex_table_column_t
...
+    void set_projection(char const *projection);
+
+    std::string sql_type_name() const;
     std::string sql_modifiers() const;
-    std::string sql_create(int srid) const;
+    std::string sql_create() const;
+
+    int srid() const noexcept { return m_srid; }
+
+    std::shared_ptr<reprojection> projection() const
```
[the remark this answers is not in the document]
joto: fixed

src/osmium-builder.hpp
joto: removed.

src/output-flex.cpp
joto: Yes, this is expensive. But this code is hopefully soon going to be revised anyway when we clean up the geometry generation with a cache holding already created geometries. I opted to leave this in for the time being. In most cases where this is used the projection for the `area` column will be the same as for the geometry column anyway, in which case the first option with the hard-coded `IdentityProjection` is used. So the expensive path here will only be taken if you use a non-mercator projection and use an area column.

lonvia: I disagree that geometry generation will automatically fix that. The problem at hand here is that this PR switches from having one global projection object to instantiating them locally with potentially different projections. This is orthogonal to having multiple geometries (which would still all use the same global projection if it wasn't for this PR). Fixing this requires to reconsider how the projection code is called, possibly getting rid of the shared object and having something lightweight that can be instantiated on the fly.

PR #1278, opened 2020-10-01
Keep the man page in markdown format and compile it with pandoc

joto:
This makes the man page easier to update and it can be included more
easily into the documentation on the osm2pgsql website.

It uses pandoc which is readily available as packages on all major Linux
distributions to compile the man page into the special format for man
pages.

This will probably not work on Windows, but man pages aren't used there
anyway. Making the man page available on the web should make it easier
for Windows people to access it anyway.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
 
       - name: Install prerequisits
         run: |
-          brew install lua boost postgis
+          brew install lua boost postgis pandoc
           pip3 install psycopg2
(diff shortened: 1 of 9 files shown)


SomeoneElseOSM:
(re Windows) the nearest thing Windows has to a package manager does include it: https://chocolatey.org/packages/pandoc .

joto:
I force-pushed a new version that only builds the man page when calling "make man". The man page is now included in its compiled form in the repository for easy access for users.

PR #1280, opened 2020-10-05
Print database version and do version check after startup

joto:
Right after startup, connect to the database and print and check the
database version.

This introduced a new function check_db() which does this check and can
later be extended to do other checks on the database, for instance the
PostGIS version etc.


src/db-check.hpp
```diff
@@ -0,0 +1,8 @@
+#ifndef OSM2PGSQL_DB_CHECK_HPP
+#define OSM2PGSQL_DB_CHECK_HPP
+
+#include "options.hpp"
+
+void check_db(options_t const &options);
```
lonvia, line 6: Can you please document what the function does. The name is rather unspecific.

joto: Just pushed a new version that has some documentation.

PR #1281, opened 2020-10-05
Allow conninfo string in -d/--database option

### LINKED ISSUE #1242: Add ability to provide conninfo string (database URL) in -d field
Dear maintainers,

Many software using libpq allow to provide a conninfo string with all required information in one single field instead of 3 different arguments.
https://www.postgresql.org/docs/11/libpq-connect.html#LIBPQ-CONNSTRING

For instance :

```sh
psql -d postgresql://mypgserver:5432/database
```

instead of 
```sh
psql -d database -h mypgserver -p 5432
```

It actually eases config writing and make scripts a little lighter
Would you accept to add this possibility to osm2pgsql -d argument please?
```sh
osm2pgsql -d postgresql://mypgserver:5432/database
```

Currently, it can't be used this way in osm2pgsql 0.96 (Debian)

All the best

joto:
Optionally allow a PostgreSQL conninfo string in the -d/--database
option instead of just a database name. Similar to what the psql comand
does.

Fixes #1242

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -124,7 +124,8 @@ void long_usage(char const *arg0, bool verbose)
                         for full planet imports. Default is disabled.\n\
     \n\
     Database options:\n\
-       -d|--database    The name of the PostgreSQL database to connect to.\n\
+       -d|--database    The name of the PostgreSQL database to connect to or\n\
+                        a PostgreSQL conninfo string.\n\
        -U|--username    PostgreSQL user name (specify passsword in PGPASSWORD\n\
                         environment variable or use -W).\n\
        -W|--password    Force password prompt.\n\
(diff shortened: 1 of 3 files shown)


lonvia:
Could we adapt some of the regression tests to test this?

joto:
As far as I can see, all regression tests just use the same line `cmdline.extend(('-d', CONFIG['test_database']))` in `def run_import`. I am not sure how to best bring this in there.

src/options.cpp
```diff
@@ -256,10 +257,27 @@ void long_usage(char const *arg0, bool verbose)
...
+}
+
 std::string database_options_t::conninfo() const
 {
+    if (compare_prefix(db, "postgresql://") ||
+        compare_prefix(db, "postgres://")) {
+        return db;
+    }
+
     std::string out{"fallback_application_name='osm2pgsql'"};
```
pnorman, line 273: Should this be above the block for `postgresql://` handling? We still want to set the fallback application name

src/options.cpp
```diff
@@ -256,10 +257,27 @@ void long_usage(char const *arg0, bool verbose)
...
 
+    if (std::strchr(db.c_str(), '=') != nullptr) {
```
pnorman, line 275: Would this cause problems with properly escaped DB names containing a =?

src/options.cpp
joto: No. In that case we can't set the application name if I understand the postgresql docs correctly. These are two totally different ways of specifying the connection string and they can't be mixed.

src/options.cpp
joto: Yes, it probably would. But how likely is that? I haven't looked at the code in `psql`, but I implemented this from their man page and that says: `If this parameter contains an = sign...`

lonvia:
> As far as I can see, all regression tests just use the same line cmdline.extend(('-d', CONFIG['test_database'])) in def run_import. I am not sure how to best bring this in there.

I'd suggest to call the `extend()` only if `params` does not contain a `-d` or `--database` already. The you can add a special `-d` to some of the tests (or have a separate test for it).

joto:
I have pushed a new version that contains tests along those lines.

PR #1288, opened 2020-10-11
Extended schema support to middle and pgsql/multi output

joto:
There was already support in progress for schemas in the flex backend,
which is likely to replace the pgsql backend at some point. Until then,
it's useful to be able to import data directly into a schema.

This is the original PR from Matt Amos with some changes from me: There
are two different options now for setting the schema for the middle and
the pgsql output, respectively. There are some docs and tests now.

This PR is an improved version of #1241


tests/regression.py
```diff
@@ -121,6 +124,10 @@ def setUpClass(cls):
 
     @classmethod
     def tearDownClass(cls):
+        with psycopg2.connect("dbname='{}'".format(CONFIG['test_database'])) as conn:
+            with conn.cursor() as cur:
+                if cls.schema:
```
lonvia, line 129: You'd want to pull out this 'if' to avoid creating a connection when there is no schema.

tests/regression.py
```diff
@@ -751,14 +762,20 @@ class TestDBAccessConninfo(BaseUpdateRunner, unittest.TestCase,
...
 class TestDBAccessURIPostgres(BaseUpdateRunner, unittest.TestCase,
-                           PgsqlBaseTests):
+                              PgsqlBaseTests):
     extra_params = ['--slim', '-d', 'postgres:///' + CONFIG['test_database']]
 
+# Schema tests
+
+class TestDBOutputSchema(BaseUpdateRunnerWithOutputSchema, unittest.TestCase,
+                         PgsqlBaseTests):
+    extra_params = ['--slim', '--output-pgsql-schema=osm']
```
lonvia, line 780: I would expect that the `PgsqlBaseTests` do not work here because the tests use table names without a schema. Am I missing something?

tests/regression.py
joto: Fixed.

tests/regression.py
joto: The test was accessing tables in the public namespace left over from earlier tests. Fixed now.

joto:
I added some additional checks in a new commit and rebased everything on master.

PR #1292, opened 2020-10-16
Do not create planet_osm_nodes table if flat nodes are used

### LINKED ISSUE #39: planet_osm_nodes table is unnecessary when --flat-nodes is used
Currently osm2pgsql will create a planet_osm_nodes table even if --flat-nodes is specified. It would probably be best to not create this table.

It is also probably possible to screw up the DB by omitting --flat-nodes on diff processing and this could eliminate that possibility,

joto:
Also checks in append mode whether the table is there. If not and there
is no --flat-nodes option specified, osm2pgsql will stop with an error.

Fixes #39

### DIFF
--- a/src/middle-pgsql.cpp
+++ b/src/middle-pgsql.cpp
@@ -578,14 +578,12 @@ void middle_pgsql_t::start()
         // (Re)create tables.
         m_db_connection.exec("SET client_min_messages = WARNING");
         for (auto const &table : m_tables) {
-            if (!table.m_create.empty()) {
-                fmt::print(stderr, "Setting up table: {}\n", table.name());
-                auto const qual_name = qualified_name(
-                    table.m_copy_target->schema, table.m_copy_target->name);
-                m_db_connection.exec(
-                    "DROP TABLE IF EXISTS {} CASCADE"_format(qual_name));
-                m_db_connection.exec(table.m_create);
-            }
+            fmt::print(stderr, "Setting up table: {}\n", table.name());
+            auto const qual_name = qualified_name(
+                table.m_copy_target->schema, table.m_copy_target->name);
+            m_db_connection.exec(
+                "DROP TABLE IF EXISTS {} CASCADE"_format(qual_name));
+            m_db_connection.exec(table.m_create);
         }
 
         // The extra query connection is only needed in append mode, so close.
--- a/src/middle-pgsql.cpp
+++ b/src/middle-pgsql.cpp
@@ -577,13 +577,15 @@ void middle_pgsql_t::start()
     } else {
         // (Re)create tables.
         m_db_connection.exec("SET client_min_messages = WARNING");
-        for (auto &table : m_tables) {
-            fmt::print(stderr, "Setting up table: {}\n", table.name());
-            auto const qual_name = qualified_name(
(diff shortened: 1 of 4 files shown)


src/middle-pgsql.cpp
```diff
@@ -577,13 +577,15 @@ void middle_pgsql_t::start()
...
-            m_db_connection.exec(
-                "DROP TABLE IF EXISTS {} CASCADE"_format(qual_name));
-            m_db_connection.exec(table.m_create);
+        for (auto const &table : m_tables) {
+            if (!table.m_create.empty()) {
+                fmt::print(stderr, "Setting up table: {}\n", table.name());
+                auto const qual_name = qualified_name(
+                    table.m_copy_target->schema, table.m_copy_target->name);
+                m_db_connection.exec(
+                    "DROP TABLE IF EXISTS {} CASCADE"_format(qual_name));
```
lonvia, line 586: I think we should still try to drop the table here, just in case that it was left over from a previous run without flatnode store.

tests/regression.py
```diff
@@ -749,6 +752,14 @@ class TestMPUpdateSlimLuaHstore(MultipolygonUpdateRunner, unittest.TestCase,
     extra_params = ['--slim', '-k']
     use_lua_tagtransform = True
 
+class TestMPUpdateSlimWithFlatNodes(MultipolygonUpdateRunner, unittest.TestCase,
+                                    MultipolygonTests):
+    extra_params = ['--slim', '-F', 'flat.nodes']
```
lonvia, line 757: You'd need to add a cleanup for that file.

src/pgsql.cpp
```diff
@@ -55,11 +55,18 @@ void pg_conn_t::set_config(char const *setting, char const *value) const
...
+        query(PGRES_COMMAND_OK, sql);
+    }
+}
 
 void pg_conn_t::exec(std::string const &sql) const
 {
-    query(PGRES_COMMAND_OK, sql.c_str());
+    if (!sql.empty()) {
+        query(PGRES_COMMAND_OK, sql.c_str());
+    }
```
lonvia, line 69: These two changes don't see really needed for this PR?

src/middle-pgsql.cpp
joto: Should we check whether there is data in there before dropping it? Or maybe just warn the user?

lonvia: We've never done this so far.

tests/regression.py
joto: Fixed in 3947e44

src/middle-pgsql.cpp
joto: Fixed in a5a2642

src/pgsql.cpp
joto: Because of [this `if`](https://github.com/openstreetmap/osm2pgsql/blob/cd9770b2d724f4e7602b50659acc6684d86d3a51/src/middle-pgsql.cpp#L634), `sql.create_table` and `sql.prepare_query` can be empty, but they are still run. It was easiest to just not run empty queries instead of putting checks in more places.

For consistency I decided to add the emptyness check in both `exec` functions.

PR #1293, opened 2020-10-16
Fix order of parameters in set_config()

lonvia:
When using the UPDATE notation then the value of the setting comes before the name.

### DIFF
--- a/src/pgsql.cpp
+++ b/src/pgsql.cpp
@@ -51,7 +51,7 @@ void pg_conn_t::set_config(char const *setting, char const *value) const
     // implemented.
     auto const sql =
         "UPDATE pg_settings SET setting = '{}' WHERE name = '{}'"_format(
-            setting, value);
+            value, setting);
     query(PGRES_TUPLES_OK, sql);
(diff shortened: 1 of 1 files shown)


lonvia:
This is worth backporting to 1.3.

mboeringa:
> This is worth backporting to 1.3.

I am a little bit surprised this even ever worked? Shouldn't this have thrown an exception in osm2pgsql, or is this from unreleased code only?

lonvia:
> I am a little bit surprised this even ever worked? Shouldn't this have thrown an exception in osm2pgsql, or is this from unreleased code only?

The whole point of the code was that it does not throw an error, which is also why it is hard to catch with CI. See comment above the SQL statement.

mboeringa:
> The whole point of the code was that it does not throw an error, which is also why it is hard to catch with CI. See comment above the SQL statement.

Yes, missed that, because I was just looking at the change view in GitHub. Just curious, any examples of the types of settings that are changed by osm2pgsql on run time?

PR #1294, opened 2020-10-19
clean up cmake code for PostgreSQL detection

lonvia:
Remove the parts that set explicit directories for the Postgresql repository and rely on PostgreSQL_ADDITIONAL_VERSIONS instead. This is now set to all version we support.


lonvia:
I could not quite figure out why Travis needs the PostgreSQL include path set but it seems to be the only one that needs it. Works fine installing from the Postgres repos without it.

PR #1310, opened 2020-11-05
Add --middle-propagate-changes option

### LINKED ISSUE #1304: Do not propagate flag
Nominatim doesn't need the propagation of changes from nodes to ways to relations. Currently there is a hack in the code that when the Nominatim output is used, the middle disables the propagation. This should be a command line option instead, so that

a) we can switch Nominatim to the flex output
b) other software which might not need propagation can use the same switch

joto:
Add new command line option --middle-propagate-changes. Set this to
"false" to suppress the normal propagation of changes in append mode
from nodes to ways to relations. This is intended mainly for Nominatim
which doesn't need this propagation.

I have chosen the flag name as --middle-propagate-changes=VALUE to leave
the possibility open to later allow values like "nodes-only" or
something like that. But currently only "true" and "false" are
supported.

This replaces the hardcoded check for the gazeteer output in the middle.
As the (now deleted) comment in the code shows an earlier intention was
to let the output tell the middle whether it needs the propagation. But
this commit is not going that route: Some outputs can not know whether
the propagation is needed (i.e. the flex output) without having an extra
flag for that. But if we need a flag anyway, it makes more sense to have
it on the middle where it belongs thematically.

Fixes #1304


src/middle-pgsql.cpp
```diff
@@ -72,13 +72,7 @@ middle_pgsql_t::table_desc::table_desc(options_t const &options,
...
-    // Gazetteer doesn't use mark-pending processing and consequently
-    // needs no way-node index.
-    // TODO Currently, set here to keep the impact on the code small.
-    // We actually should have the output plugins report their needs
-    // and pass that via the constructor to middle_t, so that middle_t
-    // itself doesn't need to know about details of the output.
-    if (options.output_backend != "gazetteer") {
+    if (options.propagate_changes) {
         m_prepare_intarray = build_sql(options, ts.prepare_mark);
         m_array_indexes = build_sql(options, ts.create_index);
```
lonvia, line 77: May I suggest to rename `prepare_mark` and `create_index` to reflect that this is the SQL needed for propagate_changes? The names were never particularly good to start with.

lonvia (commented):
The change looks okay. Just some nitpicking wrt naming: internally, we have called the process 'dependency management'. So the previous terminology of forward-dependency might fit better in general than propagate-changes.

src/middle-pgsql.cpp
joto: I added a commit for that.

src/osmdata.cpp
```diff
@@ -193,11 +194,7 @@ class multithreaded_processor
             auto copy_thread = std::make_shared<db_copy_thread_t>(conninfo);
 
             for (auto const &out : m_outputs) {
-                if (out->need_forward_dependencies()) {
-                    m_clones[i].push_back(out->clone(midq, copy_thread));
-                } else {
-                    m_clones[i].emplace_back(nullptr);
-                }
+                m_clones[i].push_back(out->clone(midq, copy_thread));
```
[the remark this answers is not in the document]
joto: Fixed.

joto:
New version of this PR rebased on top of master with new naming.

lonvia (approved):

PR #1318, opened 2020-11-10
Add trim(), split_unit() and split_string() Lua helper functions

joto:


src/init.lua
```diff
@@ -116,6 +116,36 @@ function osm2pgsql.make_clean_tags_func(keys)
...
 end
 
+-- from http://lua-users.org/wiki/StringTrim
+function osm2pgsql.trim(str)
+    local from = str:match("^%s*()")
+    return from > #str and "" or str:match(".*%S", from)
+end
+
+function osm2pgsql.split_unit(str, default_unit)
+    local val, unit = string.match(str, "^([0-9.]+) ?(%a*)$")
```
lonvia, line 126: Do we want to allow negative values?

joto: I had thought about that and don't see where this would be useful. `width`, `maxspeed` can only be positive. Maybe for `ele`?

joto: Its a low-level function that shoulde decide what the correct range of values is, so I have changed it to allow negative values.

joto:
PR has now an additional commit adding checks for nil on several Lua functions.

pnorman:
What is the value to adding these as helper functions?

joto:
See discussion in #1079 . These are just some pretty generic and useful functions and you got to start somewhere.

Issue #1316, opened 2020-11-10
Handling of "Multi" geometries in flex output


joto:
The flex output is still experimental. This gives us the chance to re-think the non-multi vs. multi-geometry issues. I see two distinct issues here:

1. Should multi-geometries be split into single geometries?
2. How does the generation of geometries relate to the type of the target geometry column?

Lets look at (multi)polygons first.

ad 1)

* Pro splitting: Faster index lookup, faster rendering, its what osm2pgsql has been doing as default up to now
* Con splitting: Ids not unique any more, rendering problems with labels, not expected by new users

ad 2) Closed ways and some multipolygon relations will result in Polygon geometries, other multipolygon relations in MultiPolygon geometries. If the target column has type "Geometry" we can store both Polygons and Multipolygons in it. If the target column has type "MultiPolygon" and we want to store a Polygon in it, we can either error out or "upgrade" all Polygons to MultiPolygons. If the target column has type "Polygon" and we want to store a "Multipolygon" we can either error out split it into several rows.

My current thinking here is:

ad 1) The default should be not to split. It is the "safe" choice, the one with the least surprises. Splitting is a performance tweak that has its drawbacks, so people should have to conciously choose it. This is also easier to document/explain.
ad 2) Setting a column type is a concious choice by the user. So if they choose, "Geometry", "Polygon", or "MultiPolygon" type, this is what they should get. But splitting of MultiPolygons in rows (see 1) should not be done automatically if the column has type row, instead this should be the one option, people have to set.

See also:

* http://www.paulnorman.ca/blog/2014/03/osm2pgsql-multipolygons/
* #1313 

Now to LineStrings: Here the situation is somewhat different. We are not splitting MultiLinestrings, we are splitting long lines. The flex output already doesn't do this any more unless the `split_at` parameter is used. This allows the user to set the maximum length of the segments and has the obvious default of "do not split". We could introduce a special value for `split_at` that says: Only split MultiLineStrings into their members, if we want that. But otherwise this looks okay to me.

For the column type issue (2) I think this should have the same logic as for (Multi)Polygons.

And now that I am thinking about all this: Maybe the behaviour for LineStrings and Polygons should be more similar? What if the splits of Multipolygons can do the equivalent of the LineString split and split a (Multi)Polygon along a grid? We don't have to introduce this now, but it would be the logical extension of what we have now.

@pnorman ? @lonvia ?

mboeringa:
@joto 

[...]

PR #1327, opened 2020-11-18
Optionally wrap polygon geometries in multipolygons

joto:
When you have an area table in the flex output, you always had to use the "geometry" column type for the geometry column, because ways are turned into polygons and multipolygon relations are turned into polygons or multipolygons. The common type for these is the generic "geometry".
    
With this change, declaring a geometry column as type "multipolygon" now works. All polygon geometries are automatically wrapped in a multipolygon before adding them to the database. Depending on what you are doing this might be want you want or not. It will make the geometries larger, but it also makes the data more consistent and so easier to work with and some software can only handle this consistent data.
    
See #1316


src/geom-transform.cpp
```diff
@@ -86,44 +89,52 @@ bool geom_transform_area_t::set_param(char const *name, lua_State *lua_state)
...
 {
     assert(builder);
 
-    return builder->get_wkb_multipolygon(relation, buffer, m_multi);
+    auto create_multigeoms = geom::create_multi_t::never;
+    if (target_geom_type == table_column_type::multipolygon) {
+        create_multigeoms = geom::create_multi_t::always;
+    } else if (m_multi) {
+        create_multigeoms = geom::create_multi_t::optional;
+    }
```
lonvia, line 135: If I read this right, then the `multi` flag is ignored, when the type is MULTIPOLYGON. Makes perfect sense but it might be worth documenting (and possibly testing, to document that this was the intended behaviour ;).

lonvia (commented):
Looks good to me. Can we have nice error messages now instead of COPY failing on bad geometry types?

joto:
I have restructured the code to better match intentions and added tests. Also added a new commit that switches the default for the multi setting from false to true as outlined in #1316. The config option is now called `split_at` to better match the somewhat similar option for lines, although the only allowed value is `multi`.

src/geom-transform.cpp
joto: After some more thought, this wasn't the right choice. The `multi` setting should be independent of the target geometry type. The new code implements this that way.

joto:
> Looks good to me. Can we have nice error messages now instead of COPY failing on bad geometry types?

Unfortunately not. The code creates WKB geometries and you can't tell later what type of geometry they are (without hacks), so we can't do the check. We'll rework the geometry code anyway in the near future using actual geometry types and then we can do this easily.

PR #1328, opened 2020-11-18
Flex output: Remove code that alters table in append mode

joto:
This code was basically copied over from the pgsql output. It is a
half-measure that can detect some needed changes in your database tables
when you change the config after import. But this was never tested
properly and what is implemented is not enough to make this a useful
feature. So lets get rid of it.


lonvia:
@pnorman Any idea is openstreetmap-carto has ever used this feature to add additional columns for the style?

lonvia:
Ah, never mind. This PR is only for the flex backend, that's fine.

Issue #922, opened 2019-03-14
Support PROJ >= 6.0.0


tburschka:
Currently the build failed if proj >= 6.0.0 is used. The Flag ACCEPT_USE_OF_DEPRECATED_PROJ_API_H has to be defined. 

see: https://github.com/postgis/postgis/pull/354/files#r242068366

jmckenna:
I agree that this flag is required now (as PROJ 6 was released on 2019-03-01).

Note that PROJ 7 release next year will drop proj_api.h completely.

pnorman:
Do we use proj directly or only via libosmium?

jmckenna:
Ah, it seems through libosmium only.

jmckenna:
libosmium's associated ticket: https://github.com/osmcode/libosmium/issues/277

jmckenna:
You can close this ticket, but I bet this will be an FAQ.

jmckenna:
related PROG 7.0 discussion happening at osmcode/libosmium#295

lonvia:
Reopening this. We are going to need osmium-independent proj >= 6 support.

kennykb:
Thanks for reopening. I was going to grumble that I produce some paper maps projected to UTM zone 18 and therefore maintain a small subset of OSM in PostGIS in that projection, so losing support for everything but plate carree and spherical Mercator would be a problem.

lonvia:
#1066 reorders the code so that it becomes fairly easy to implement a generic projection using the proj6 variant and then have cmake detect which one to use. Just add a `reprojection-generic-proj6.cpp` similar to the existing `reprojection-generic-proj4.cpp` implementation.

PRs would be most welcome.

jmckenna:
support for PROJ >6.1 added by @joto in https://github.com/openstreetmap/osm2pgsql/pull/1331   

I think this ticket can be closed.

PR #1331, opened 2020-11-20
Work-in-progress: Add support for projection using PROJ library >6.1

joto:
This is a stab at adding projection support using the PROJ library from
6.1 onwards. There might be an issue with multithreading, but the old
code most likely already had that issue. This needs to be checked.

The CMake config will prefer using the old API, simply because I don't
know how to check for the new API without it trying to use the new API
with version 5-6.0 which don't have everything we need. The code will
only work from 6.1 onwards.


joto:
See #922

jmckenna:
thank-you @joto for this

PR #1336, opened 2020-11-23
Check that you have any output tables defined in the flex config

### LINKED ISSUE #1332: osm2pgsql should error out when "-O flex" is chosen, but old style pgsql LUA script is fed
In the process of building a new virtual machine, I got a little bit confused and accidentally fed osm2pgsql the old style LUA script "style.lua" that is in the main folder when you create an osm2pgsql install using a Git clone, while having the "-O flex" option set on the command line.

Of course, this is wrong input, but nonetheless osm2pgsql did not error out. It actually went through the first stages of loading data, visibly showing progress loading nodes, ways and relations, then stopped without throwing an error, yet the output showed it did not generate e.g. indexes. When I inspected the resulting database, no table was there, the database was empty, so the changes did not get committed.

I think it would be good, with now two backends using different type LUA scripts, that osm2pgsql somehow can distinguish old style LUA scripts and throw a user friendly error to warn users of the wrong input, to prevent this type of issue of mixing up "old" **pgsql** and "new" **flex** type LUA scripts.

joto:
Without tables defined there isn't anything to do. So this means
something is wrong with your configuration. Possibly you are using a Lua
transform config for the pgsql output instead of a flex config.

Fixes #1332

### DIFF
--- a/src/output-flex.cpp
+++ b/src/output-flex.cpp
@@ -1388,6 +1388,11 @@ output_flex_t::output_flex_t(
         init_lua(m_options.style);
     }
 
+    if (m_tables->empty()) {
+        throw std::runtime_error{
+            "No tables defined in Lua config. Nothing to do!"};
+    }
+
     assert(m_table_connections.empty());
     for (auto &table : *m_tables) {
         m_table_connections.emplace_back(&table, m_copy_thread);


mboeringa:
@joto, just one minor remark: I am OK with the "generic" warning as first stop, but maybe adding at least a hint about a possible mix-up of old **pgsql** LUA and new **flex** LUA configuration, would be good to add. 

The generic message, although correct, could still have people with little experience guess why their **pgsql** LUA scripts they downloaded didn't work, as people may not be aware of the difference between these outputs, and the drastic differences in the LUA scripts.

This could especially happen for people with little to no programming experience, that are not familiar with reading of code...

joto:
@mboeringa: I don't want to add too much wording to the error messages. Instead I am adding the error message to the FAQ with a longer text that can explain what's going on and what possible reasons for this error are. This way chances are that confused people will find it using a Google search. And I can have a longer text is also easier to change later.

PR #1349, opened 2020-12-01
Various cleanups on help texts etc

joto:
See individual commits

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -98,180 +98,152 @@ void long_usage(char const *arg0, bool verbose)
 {
     char const *const name = program_name(arg0);
 
-    printf("Usage:\n");
-    printf("\t%s [options] planet.osm\n", name);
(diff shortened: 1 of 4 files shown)


src/options.cpp
```diff
@@ -98,180 +98,152 @@ void long_usage(char const *arg0, bool verbose)
...
+                    gazetteer - Output to a PostGIS database for Nominatim\n\
+                    null - No output. Used for testing.\n\
+    -S|--style=FILE  Location of the style file. Defaults to\n\
+                    '" DEFAULT_STYLE "'.\n\
+    -k|--hstore     Add tags without column to an additional hstore column.\n",
+               stdout);
+#ifdef HAVE_LUA
+    std::fputs("\
+       --tag-transform-script=SCRIPT  Specify a Lua script to handle tag\n\
+                    filtering and normalisation.\n",
```
lonvia, line 126: I suggest to add a `(pgsql output only)`

joto: I have added this to the commit.

PR #1350, opened 2020-12-01
Various Cleanups

joto:
See individual commits.

### DIFF
--- a/src/node-ram-cache.cpp
+++ b/src/node-ram-cache.cpp
@@ -240,14 +240,9 @@ void node_ram_cache::set_dense(osmid_t id, osmium::Location location)
         }
 
         if (queue[expectedpos] != &blocks[block]) {
-            if (m_warn_node_order) {
-                log_warn("Found out of order node {}"
-                         " ({},{}) - this will impact the cache efficiency",
-                         id, block, offset);
-
-                // Only warn once
-                m_warn_node_order = false;
-            }
+            // This can only happen if the node is out of order which it
+            // should never be because of earlier checks for that.
+            // Ignore the node.
             return;
         }
     }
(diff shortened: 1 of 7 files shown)


src/node-ram-cache.cpp
```diff
@@ -240,14 +240,6 @@ void node_ram_cache::set_dense(osmid_t id, osmium::Location location)
...
 
         if (queue[expectedpos] != &blocks[block]) {
-            if (m_warn_node_order) {
-                log_warn("Found out of order node {}"
-                         " ({},{}) - this will impact the cache efficiency",
-                         id, block, offset);
-
-                // Only warn once
-                m_warn_node_order = false;
-            }
```
lonvia, line 250: This leaves us with an odd 'if' check that is not quite understandable. If you don't want to touch the code any further than removing the check, I would suggest to add a comment about what used to be here aka 'This can only happen if the node is out of order. Ignore the node.'

src/options.cpp
```diff
@@ -726,24 +738,14 @@ void options_t::check_options()
...
-    }
-
-    if (sizeof(int *) == 4 && !slim) {
-        log_warn(
-            "This is a 32bit system with not a lot of RAM. Try using --slim.");
-    }
-
     // zoom level 31 is the technical limit because we use 32-bit integers for the x and y index of a tile ID
-    if (expire_tiles_zoom_min >= 32) {
+    if (expire_tiles_zoom_min > 32) {
```
lonvia, line 742: Did you mean to change the 32 to 31?

src/node-ram-cache.cpp
joto: I added a comment.

src/options.cpp
joto: Fixed.

PR #1371, opened 2020-12-14
Refactor middle

joto:
Various refactorings of the middle, removing the slim_middle_t class etc.

### DIFF
--- a/src/middle-pgsql.cpp
+++ b/src/middle-pgsql.cpp
@@ -26,6 +26,7 @@
 #include "node-ram-cache.hpp"
 #include "options.hpp"
 #include "osmtypes.hpp"
+#include "pgsql-helper.hpp"
 #include "util.hpp"
 
 /**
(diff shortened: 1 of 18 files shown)


src/osmdata.cpp
```diff
@@ -34,21 +34,12 @@ osmdata_t::osmdata_t(std::unique_ptr<dependency_manager_t> dependency_manager,
...
-slim_middle_t &osmdata_t::slim_middle() const noexcept
-{
-    auto *slim = dynamic_cast<slim_middle_t *>(m_mid.get());
-    assert(slim);
-    return *slim;
-}
-
 void osmdata_t::node(osmium::Node const &node)
 {
+    if (!m_bbox.valid() || node.deleted() || m_bbox.contains(node.location())) {
```
lonvia, line 39: contains() is now called twice. Not the cheapest of functions.

More importantly: you either need to move the node.location().valid() check to the front of the function or make sure that mid can handle invalid locations. I'd be in favour of the first solution. (Bonus points for adding a test for input with invalid locations.)

src/middle-pgsql.cpp
```diff
@@ -314,6 +314,46 @@ std::size_t middle_query_pgsql_t::get_way_node_locations_db(
...
+void middle_pgsql_t::node(osmium::Node const &node)
+{
+    if (node.deleted()) {
+        node_delete(node.id());
+    } else {
+        if (m_options->append) {
+            node_delete(node.id());
+            node_set(node);
+        } else {
+            node_set(node);
```
lonvia, line 327: Any particular reason why you don't move the node_set() outside the if? Dto for the other two functions.

joto: Just not gotten around to that. I have added a commit now that does that.

src/osmdata.cpp
joto: In an additional commit I have refactored this and added a test.

PR #1384, opened 2020-12-21
WIP: Experimental middle implementations

joto:
This PR adds two new middle implementations:
- `db` is similar to `pgsql`. All data is stored in the database. A new table structure with JSONB for tags and relation members
  is used.
- `mem` is similar to `ram`. It only stores node locations and way node lists in RAM. Used for import-only workflows that do
  not need the flex-output two-stage processing.

These are **EXPERIMENTAL**! They are not well tested, not complete and performance might be horrible. They are intended to start a discussion about what we need and how to best implement it. This PR will evolve over time.

To enable these, set the environment variable `OSM2PGSQL_MIDDLE` to either `db` or `mem`. This setting is independent of the `--slim` setting! This way of setting the middle is temporary and gives us flexibility while experimenting, it will not be available in this form in the final implementation.

More about this on [in my blog](https://blog.jochentopf.com/2020-12-21-osm2pgsql-middle-improvements.html).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -9,6 +9,7 @@ set(osm2pgsql_lib_SOURCES
   input.cpp
   logging.cpp
   middle.cpp
+  middle-db.cpp
   middle-pgsql.cpp
   middle-ram.cpp
   node-locations.cpp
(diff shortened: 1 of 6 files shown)


mmd-osm:
Are there any plans to support `-r pbf,locations_on_ways=true` as well? I think it would be a good memory efficient option for single-stage initial load scenarios.

joto:
New version of this PR. The 'mem' middle has been removed, because it has been merged into master.

joto:
osm2pgsql has changed quite a lot since this PR was made and it isn't really useful any more. Closing here. We are still planning to reintroduce something like this later.

PR #1387, opened 2021-01-07
Remove deprecated multi backend

### LINKED ISSUE #1062: Retire multi output
I'd like to propose to deprecate and eventually remove the multi output.

The multi output was an interesting experiment but I don't think that it was ever really used in production by anybody. The only style I'm aware of is @pnorman experimental style. The code for the multi output is rather complicated and hard to maintain. There are also some performance issues we've never tackled. The new 'flex' output can do the same as the multi output but hopefully in a more flexible and performant way. 

For the moment, I'd just add a deprecation warning when the output is used. Once the 'flex' output is production-ready, we could add a Lua script that processes a multi output configuration and then remove the output completely.

If you are a user of the multi output, please comment here.

mmd-osm:
Hopefully this covers all bits to remove the deprecated “multi” backend and doesn’t delete anything that would still be needed. 

Man page re-generation isn’t working here for some reason: it was creating a huge amount of changes which I didn’t want to check in as is.

The following four variables in options.hpp are no longer being used. I removed them in a separate commit and use hard-coded defaults in tagtransform-lua.cpp instead, if that's ok.

```
    boost::optional<std::string> tag_transform_node_func{boost::none};
    boost::optional<std::string> tag_transform_way_func{boost::none};
    boost::optional<std::string> tag_transform_rel_func{boost::none};
    boost::optional<std::string> tag_transform_rel_mem_func{boost::none};
```

Closes #1062


lonvia:
Thanks for all the work but I'm afraid that this particular topic is not the best for an external contribution. We haven't decided yet when exactly to remove the code. It might be tomorrow, or in a month or in three. Even a couple of weeks make a huge difference because the change is so extensive that it is bound to cause merge conflicts with any other changes we do. It is one of these changes where you stop everything else, get it done and then recommence work.

So I'm going to close this for now. Sorry about that but it seems unfair to have you invest a lot of work without being aware about the above.

mmd-osm:
Thanks for your feedback, that's really unfortunate. My impression was that the repo was going into some kind of hibernation mode now that the OSMF funding ended  osm2pgsql is maybe not an ideal repo for an occasional external contributor, as there's simply too much going on behind the scenes, and you have little chance to find out what's next.

lonvia:
There are opportunities for external contribution, removing a feature is just not one of them. You are kind of right about hibernation mode, though, and it is one of the reasons why I don't know when exactly multi will go away.  There is no reason to pull it until it is in the way of developing new code. As long as we are in bugfix-only mode, it might as well stay.

mmd-osm:
Ok, this makes sense. I just leave the branch as is now. It's probably not that difficult to re-apply those changes at some point in the future (whenever it seems reasonable), even it would mean to fix some merge conflicts.

PR #1392, opened 2021-01-10
Simplify build on Alpine

hoxovic:
Hey,

I was building osm2pgsql in an Alpine container and due to Alpine's popularity I figured it couldn't hurt to add Alpine build instructions while I'm at it.

I may have included a bit too much detail, but I was hoping it could save someone some time of troubleshooting why CMake is only finding part of the Lua "dependencies" (no unversioned symlink to `/usr/lib/lua5.3/liblua-5.3.{so|a}` under `/usr/lib`).

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,22 @@ pkg install devel/cmake devel/boost-libs textproc/expat2 \
   databases/postgresql94-client graphics/proj lang/lua52
 ```
 
+On Alpine, use
+
+```sh
+apk --update-cache add cmake make g++ boost-dev expat-dev \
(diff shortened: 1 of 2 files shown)


joto:
It would probably be better to amend the `cmake/FindLua.cmake` to find the library on Alpine. Can you try with https://github.com/Kitware/CMake/blob/master/Modules/FindLua.cmake ?

hoxovic:
Good point, I agree!
I've modified the module to "reuse" the library name variable as search path suffixes.

lonvia:
Sorry, this one got lost in the mail. Thanks!

PR #1411, opened 2021-02-04
Add a script for simple updating of a database

lonvia:
As people have been asking for this, here is a simple python script for handling updates of an osm2pgsql database. It takes care of finding the initial state and then downloading updates and running osm2pgsql against them. The state is kept in a table in the same database as osm2pgsql.

Quick use:
1. Import your database: `osm2pgsql -d gis -s -S default.style liechtenstein.osm.pbf`
2. a) Initialise update from the import file: `osm2pgsql-replication init -d gis --osm-file iechtenstein.osm.pbf`
    or b) Initialise updates from database (uses minutelies per default): `osm2pgsql-replication init -d gis`
3. Run updates: `osm2pgsql-replication update -d gis -- -S default.style`

The script does not check that it is already running, so be careful when using with cron.

Needs python3, pyosmium and psycopg2. On a Ubuntu/Debian install with: `sudo apt install python3-osmium python3-psycopg2`

Try it out and let me know if that would be useful or not.


osm2pgsql-replication
```diff
@@ -0,0 +1,338 @@
...
+            return None
+
+    LOG.info("Using way id %d for timestamp lookup", osmid)
+    # Get the node from the API to find the timestamp when it was created.
+    node_url = 'https://www.openstreetmap.org/api/0.6/way/{}/1'.format(osmid)
+    headers = {"User-Agent" : "osm2pgsql-update"}
+    with urlrequest.urlopen(urlrequest.Request(node_url, headers=headers)) as response:
+        data = response.read().decode('utf-8')
+
+    match = re.search(r'timestamp="((\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2}))Z"', data)
```
pnorman, line 69: Let's not parse XML with regular expressions.

We could get this with a HEAD request and the last-modified header

osm2pgsql-replication
```diff
@@ -0,0 +1,338 @@
...
+
+
+def update_replication_state(conn, table, seq, date):
+    """ Update sequence and date in the replication state table.
+        The table is assumed to exist.
+    """
+    with conn.cursor() as cur:
+        if date is not None:
+            cur.execute("""UPDATE {} SET sequence=%s, importdate=%s""".format(table),
+                        (seq, date))
```
pnorman, line 104: Quote {} with `"` for safety

osm2pgsql-replication
```diff
@@ -0,0 +1,338 @@
...
+
+    return dt.datetime.fromisoformat(match[1]).replace(tzinfo=dt.timezone.utc)
+
+
+def setup_replication_state(conn, table, base_url, seq, date):
+    """ (Re)create the table for the replication state and fill it with
+        the given state.
+    """
+    with conn.cursor() as cur:
+        cur.execute('DROP TABLE IF EXISTS {}'.format(table))
```
pnorman, line 86: Quote {} with `"` for safety

osm2pgsql-replication
```diff
@@ -0,0 +1,338 @@
...
+        cur.execute("""CREATE TABLE {}
```
pnorman, line 87: Quote {} with `"` for safety

osm2pgsql-replication
```diff
@@ -0,0 +1,338 @@
...
+                       (url TEXT,
+                        sequence INTEGER,
+                        importdate TIMESTAMP WITH TIME ZONE)
+                    """.format(table))
+        cur.execute("INSERT INTO {} VALUES(%s, %s, %s)".format(table),
```
pnorman, line 92: Quote {} with `"` for safety

osm2pgsql-replication
```diff
@@ -0,0 +1,338 @@
...
+def update_replication_state(conn, table, seq, date):
+    """ Update sequence and date in the replication state table.
+        The table is assumed to exist.
+    """
+    with conn.cursor() as cur:
+        if date is not None:
+            cur.execute("""UPDATE {} SET sequence=%s, importdate=%s""".format(table),
+                        (seq, date))
+        else:
+            cur.execute('UPDATE {} SET sequence=%s'.format(table),
```
pnorman, line 106: Quote {} with `"` for safety

pnorman (dismissed):
A few minor things - I want to do a bit more of a review still.

osm2pgsql-replication
mmd-osm: Since you're requesting this from the API 0.6: CGImap doesn't set a last-modified header at this time.

osm2pgsql-replication
```diff
@@ -0,0 +1,338 @@
...
+        osmid = cur.fetchone()[0] if cur.rowcount == 1 else None
+
+        if osmid is None:
+            LOG.fatal("No data found in the database.")
+            return None
+
+    LOG.info("Using way id %d for timestamp lookup", osmid)
+    # Get the node from the API to find the timestamp when it was created.
+    node_url = 'https://www.openstreetmap.org/api/0.6/way/{}/1'.format(osmid)
+    headers = {"User-Agent" : "osm2pgsql-update"}
```
mmd-osm, line 65: "Get the node" and node_url still seem to refer to a previous version that was based on nodes?

osm2pgsql-replication
```diff
@@ -0,0 +1,338 @@
...
+
+
+def init(conn, args):
+    """\
+    There are two ways to initialise the replication process: if you have imported
+    from a file that contains replication source information, then the
+    initialisation process can use this and set up replication from there.
+    Use the command '%(prog)s --osm-file <filename>' for this.
+
+    If the file has no replication informaton or you don't have the intial
```
joto, line 119: initial

osm2pgsql-replication
```diff
@@ -0,0 +1,338 @@
...
+def connect(args):
+    """ Create a connection from the given command line arguments.
+    """
+    return psycopg2.connect(dbname=args.database, user=args.username,
+                            host=args.host, port=args.port)
+
+
+def compute_database_date(conn, prefix):
+    """ Determine the date of the database from the newest object in the
+        data base.
```
joto, line 50: "database"

joto:
Maybe put it in a new `scripts` directory? We should also add a man page.

osm2pgsql-replication
lonvia: Used the json version of the API and parsed that 'properly'.

lonvia:
I can use argparse-manpage to transform the output of `--help` into a man page. The output looks half way usable. Would that be sufficient?

osm2pgsql-replication
```diff
@@ -0,0 +1,350 @@
...
+    # Get the node from the API to find the timestamp when it was created.
+    url = 'https://www.openstreetmap.org/api/0.6/way/{}/1'.format(osmid)
+    headers = {"User-Agent" : "osm2pgsql-update",
+               "Accept" : "application/json"}
+    with urlrequest.urlopen(urlrequest.Request(url, headers=headers)) as response:
+        data = json.loads(response.read().decode('utf-8'))
+
+    if not data.get('elements') or not 'timestamp' in data['elements'][0]:
+        LOG.fatal("The way data downloaded from the API does not contain valid data.\n"
+                  "URL used: %s", node_url)
```
joto, line 72: `node_url`?

mmd-osm:
Would updating a Liechtenstein extract via http://download.geofabrik.de/europe/liechtenstein-updates/ be in scope for this tool?

stephankn:
Quite interesting. With osmosis being end-of-life and osmupdate also last update three years ago, it is nice to see some alternatives around.

It is nice that it detects from a Geofabrik pbf file that is has there a replication service for extracts.
@mmd-osm that would cover your use-case. run init with the osm-file parameter.

If I use a Geofabrik extract as a base for my own extract then it does still insist to load updates from there.
I would like to reduce dependency to 3rd party services. When running without a parameter it tries to guess the starting sequence. Why not allowing to specify the state file URL where to start?

I also noticed that if I update my extract that there is no way to have a polygon filtering applied before feeding in the updates to the database.

Could you include a polygon filtering like the one used here:
https://github.com/Zverik/regional/blob/master/trim_osc.py

lonvia:
@stephankn You do have slightly specific requirements. I would suggest that you simply use [pyosmium-get-update](https://docs.osmcode.org/pyosmium/latest/updating_osm_data.html). With that you can set up a sequence id manually if you want (although, I'm not quite sure why you would. There are not third party services involved except the OSM API where it looks for the date of the newest object in the database. After that it just does a binary search on the replication service to find the right state file for the date.) And you can preprocess the downloaded diff before applying to osm2pgsql.

> If I use a Geofabrik extract as a base for my own extract then it does still insist to load updates from there.

Then you have copied the replication information while doing your extract. The --osm-file options only exists for files that carry replication information in the header. No other information is used. It is meant for people who quickly want to set up update from Geofabrik or openstreetmap.fr extracts without having to bother about finding the URL of the replication service.

mmd-osm:
I tried a couple of times and received "HTTP/1.1 429 Too Many Requests" from Geofabrik in the meantime. Unfortunately, all I saw in verbose mode was "No new diffs found." which made troubleshooting a bit difficult. Some additional information that there was an issue when downloading files would be perfect.

joto (approved):

giggls:
> The script does not check that it is already running, so be careful when using with cron.

Systemd timers already include such a lock mechanism. Thus I would prefer those over cronjobs.

PR #1413, opened 2021-02-08
cmake: add option whether to install osm2pgsql

lonvia:
This option can be useful when osm2pgsql is used as a subproject and the main project wants to offer different installation commands.

In that particular case, I'd like to have it for Nominatim which would install a private copy of the osm2pgsql it needs. I'm aware that this is not an ideal solution but we are not yet at the point where Nominatim can simply use whatever osm2pgsql that is installed in the system.


lonvia:
Superseded by #1414.

PR #1423, opened 2021-02-23
Add a post-processing option for the update script

lonvia:
Setups that process expiry data need to do that after each run of osm2pgsql. The new option provides the means to insert additional operations via a script when osm2pgsql is run multiple times.

@tomhughes would this option be sufficient for the render servers to run their expiry? When you use the `--diff-file` option, then the update script explicitly leaves the diff file around, so you should be able to run the custom expiry on it.

### DIFF
--- a/scripts/osm2pgsql-replication
+++ b/scripts/osm2pgsql-replication
@@ -184,6 +184,18 @@ def update(conn, args):
     Any additional arguments to osm2pgsql need to be given after '--'. Database
     and the prefix parameter are handed through to osm2pgsql. They do not need
     to be repeated. '--append' and '--slim' will always be added as well.
+
+    Use the '--post-processing' parameter to execute a script after osm2pgsql has
+    run successfully. If the updates consists of multiple runs because the
+    maximum size of downloaded data was reached, then the script is executed
+    each time that osm2pgsql has run. When the post-processing fails, then
+    the entire update run is considered a failure and the replication information
+    is not updated. That means that when 'update' is run the next time it will
+    recommence with downloading the diffs again and reapplying them to the
+    database. This is usually safe. The script receives two parameters:
+    the sequence ID and timestamp of the last successful run. The timestamp
+    may be missing in the rare case that the replication service stops responding
+    after the updates have been downloaded.
     """
     with conn.cursor() as cur:
         cur.execute('SELECT * FROM pg_tables where tablename = %s', (args.table, ))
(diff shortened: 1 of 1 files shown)


scripts/osm2pgsql-replication
```diff
@@ -184,6 +184,18 @@ def update(conn, args):
...
     to be repeated. '--append' and '--slim' will always be added as well.
+
+    Use the '--post-processing' parameter to execute a script after osm2pgsql has
+    run sucessfully. If the updates consists of multiple runs because the
+    maximum size of downloaded data was reached, then the script is executed
+    each time that osm2pgsql has run. When the post-processing fails, then
+    the entire update run is considered a failure and the replication information
+    is not updated. That means that when 'update' is run the next time it will
+    recommence with downloading the diffs again and reapplying them to the
+    database. This is usually save. The script receives two parameters:
```
joto, line 195: safe

scripts/osm2pgsql-replication
```diff
@@ -184,6 +184,18 @@ def update(conn, args):
     Any additional arguments to osm2pgsql need to be given after '--'. Database
     and the prefix parameter are handed through to osm2pgsql. They do not need
     to be repeated. '--append' and '--slim' will always be added as well.
+
+    Use the '--post-processing' parameter to execute a script after osm2pgsql has
+    run sucessfully. If the updates consists of multiple runs because the
```
joto, line 189: successfully

joto (approved):

tomhughes:
So you're saying that if I named the file with `--diff-file` then it would still exist when the post-processing script ran and would then get overwritten when the script returned and the next pass started?

Would that be an in place overwrite? or an unlink and create new? What about if there's no more work to do and it exits, will the file get overwritten next time it runs?

Basically what we do at the moment is to hard link the files into queue directory where they are picked up by the dirtier so I'm just wondering if it will be safe for the postprocessor to do that or if the files might get overwritten before they are processed.

It also occurs to me that the other thing we use the files for at the moment is to track the last update time for munin/prometheus monitoring - not sure what the best replacement for that would be?

lonvia:
With the `--diff-file` option, it's an unlink-if-file-exists before starting the download and then opening a completely new file. That's all it does with the file. You can overwrite or move it or do anything with the diff file in the post-processing script. The updater won't care. So I believe that the hard-linking should be fine, too.

The one thing to be careful about is failures in the post-processing script. Right now, the updater considers the entire diff application a failure when the post-processing script returns with a non-zero exit code. It will exit with non-zero itself and when started the next time, it downloads the diffs again and reapplies them to the database. I'm not entirely sure if that is a good idea or not but it ensures that the post-processing script sees the data again when it fails.

> It also occurs to me that the other thing we use the files for at the moment is to track the last update time for munin/prometheus monitoring - not sure what the best replacement for that would be?

You can simply get the date from the database: `SELECT importdate FROM planet_osm_replication_status`

tomhughes:
The simplest thing for us will be a postprocessing script that just does:

```
ln <diff-file> /var/lib/replicate/expire-queue/changes-$1.osc
```

which has minimal risk of failure, and the existing expiry processing will then pick those up and process then.

lonvia:
That looks simple enough to be useful.

PR #1427, opened 2021-03-12
add support for dockerfile build

remyleone:
This dockerfile creates an image that can run osm2pgsql. The resulting image is bit large 1.09GB on my machine. Maybe it would be interesting to take advantage of [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) to build the binary statically and then copy it to a lightweight image such as alpine. But I haven't found yet a way to build osm2pgsql statically using cmake. 

Have you tried building osm2pgsql statically in the past? It could be useful to reduce the size of the container image.


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM alpine as builder
+
+RUN apk --update-cache add cmake make g++ boost-dev expat-dev bzip2-dev zlib-dev libpq proj-dev lua5.3-dev postgresql-dev
+ADD . /opt/
+RUN mkdir /opt/build
+WORKDIR /opt/build
+RUN cmake ..
```
stephankn, line 7: I used this in my docker-build. Uses LUAJIT and builds release
 cmake -D WITH_LUAJIT=ON -D CMAKE_BUILD_TYPE=MinSizeRel ..

Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+RUN make
```
stephankn, line 8: I did
` make -j$(nproc)`

stephankn:
Not certain why you need all these layers. It doesn't matter if you throw away the temporary image. Otherwise you could combine the RUN lines.

My build was only interested in the artifact, so I had another entry to create later a versioned image only containing osm2pgsql.
For inspiration, here my file:

```
FROM osmtools/builder as builder

WORKDIR /
ARG DEBIAN_FRONTEND=noninteractive
ARG CLICOLOR_FORCE=1
ENV TZ=Europe/Berlin

ARG VERSION=master

RUN git clone --depth 1 --branch $VERSION git://github.com/openstreetmap/osm2pgsql.git \
    && mkdir osm2pgsql/build && cd osm2pgsql/build && cmake -D WITH_LUAJIT=ON -D CMAKE_BUILD_TYPE=MinSizeRel .. && make -j$(nproc)

FROM scratch
WORKDIR /
COPY --from=builder /osm2pgsql/build/osm2pgsql /build/
```

I call it like:

```
OSM2PGSQL_VERSION=1.4.1
docker build --pull -t osmtools/osm2pgsql:"$OSM2PGSQL_VERSION" --build-arg VERSION="$OSM2PGSQL_VERSION" osm2pgsql
```

remyleone:
@stephankn where is located the Dockerfile used to build osmtools/builder I don't mind using it as long as we have a way to trace how it is built :)

remyleone:
As for CMAKE_BUILD_TYPE=MinSizeRel it will reduce the size of the binary but will it statically link it? Can you use a `docker run` command with the image you created with the dockerfile you mention in your comment?

stephankn:
osmtools/builder is available on docker hub. But I do not recommend to use it. This was just for "inspiration", so you get an idea on how I build it. It is an image I use to build multiple osm tools in a similar fashion than osm2pgsql. A docker built environment for osm2pgsql should stack on top of some maintained base image. Either alpine or debian/ubuntu, like you did.

The second section of the Dockerfile shows how to use files from the earlier stage. I only extract the binary.

I don't think that MinSizeRel will trigger any static linkage, but I have not checked. I also do not run the binary from within the "FROM scratch" container. That one is very empty and just holds the versioned artifact.

hypervtechnics:
We have a similar issue here:
https://github.com/mapdiscover/import-osm-service/blob/master/Dockerfile

In our case the image size is also very large. I tried to use multi-stage builds and so on, but only could get to 700 MB by manually (using try and error) removing several dependencies. There are also no official docs which libraries are needed to run the binary release once you built it. Maybe someone else might help?

lonvia:
Just as we don't do packaging for distributions ourselves, we don't have the capacity for providing Docker builds either. We don't have the infrastructure or time to maintain those extra scripts. This discussion just shows that good packaging is not trivial. So thanks for the PR but I'm going to close it.

You are welcome to continue discussions about the best way to build a dockered osm2pgsql in the [Discussion section](https://github.com/openstreetmap/osm2pgsql/discussions/categories/general).

hypervtechnics:
#1428

Issue #1436, opened 2021-03-25
Something wrong with indexes during import


andreynovikov:
```
$ osm2pgsql -d gis --prefix osm --create --slim --output=flex --style osm2pgsql.lua --flat-nodes ./osm_nodes.bin --middle-way-node-index-id-shift 5 --cache 32000 planet-latest.o5m
2021-03-24 17:48:24  osm2pgsql version 1.4.1
2021-03-24 17:48:24  Database version: 11.10 (Debian 11.10-0+deb10u1)
2021-03-24 17:48:24  PostGIS version: 2.5
2021-03-24 17:48:24  Node-cache: cache=32000MB, maxblocks=512000*65536, allocation method=11
2021-03-25 02:27:48  Reading input files done in 31162s (8h 39m 22s).         74.7/s)
2021-03-25 02:27:48    Processed 6742003174 nodes in 2640s (44m 0s) - 2554k/s
2021-03-25 02:27:48    Processed 746437654 ways in 21669s (6h 1m 9s) - 34k/s
2021-03-25 02:27:48    Processed 8731600 relations in 6853s (1h 54m 13s) - 1k/s
2021-03-25 02:27:48  Reprocess marked ways (stage 2)...
2021-03-25 02:27:48  Creating id index on table 'osm_lines'...
2021-03-25 02:32:06  Creating id index on table 'osm_boundaries'...
2021-03-25 02:32:06  Creating id index on table 'osm_polygons'...
2021-03-25 02:39:28  Creating id indexes took 700s (11m 40s)
2021-03-25 02:39:28  There are 320513 ways to reprocess...
2021-03-25 10:20:23  Clustering table 'osm_lines' by geometry...
2021-03-25 10:20:23  Clustering table 'osm_points' by geometry...
2021-03-25 10:20:23  Clustering table 'osm_polygons' by geometry...
```

Everything is going pretty fast except stage 2: I've looked what it was doing for almost 8 hours and have found out that all that time postgresql was DELETEing 320513 ways. It does not look normal.

joto:
Can you tell us what's in `osm2pgsql.lua`?

andreynovikov:
```
local srid = 3857

local prefix = 'osm_'

local tables = {}

tables.points = osm2pgsql.define_node_table(prefix .. 'points', {
    { column = 'tags', type = 'hstore' },
    { column = 'names', type = 'hstore' },
    { column = 'geom', type = 'point', projection = srid },
})

[...]

PR #1438, opened 2021-03-27
Disable expire in create mode

joto:
I create mode we never need to create an expire list. But as a side
effect of the two-stage processing we created such a list which costs a
lot of time and RAM. This disables this which speeds up an import
considerably.

See #1436

### DIFF
--- a/src/output-flex.cpp
+++ b/src/output-flex.cpp
@@ -1294,7 +1294,7 @@ void output_flex_t::delete_from_table(table_connection_t *table_connection,
     assert(table_connection);
     auto const id = table_connection->table().map_id(type, osm_id);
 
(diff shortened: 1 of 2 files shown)


src/output-flex.cpp
```diff
@@ -1294,7 +1294,7 @@ void output_flex_t::delete_from_table(table_connection_t *table_connection,
     assert(table_connection);
     auto const id = table_connection->table().map_id(type, osm_id);
 
-    if (table_connection->table().has_geom_column()) {
+    if (!m_options.create && table_connection->table().has_geom_column()) {
```
lonvia, line 1297: You should rather skip this when expire is not enabled, which may also be the case for updates.

PR #1452, opened 2021-04-13
Fix _LARGEFILE64_SOURCE redifinition

fidomax:
_LARGEFILE64_SOURCE may be defined elsewhere (f.e. in command line)

### DIFF
--- a/src/node-persistent-cache.cpp
+++ b/src/node-persistent-cache.cpp
@@ -6,8 +6,9 @@
  * Copyright (C) 2006-2021 by the osm2pgsql developer community.
  * For a full list of authors see the git log.
  */
-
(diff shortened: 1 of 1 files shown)


joto:
It doesn't matter where this is defined, as long as it is defined. But I think there is a problem here which the PR doesn't solve. We should define this in the CMake config globally for the whole project. Otherwise me have different settings in different compilation units which might lead to problems.

But there is another problem. The mentioned man page says:
> New programs should not employ this macro; instead _FILE_OFFSET_BITS=64 should be employed.

So we should probably switch to that.

So I suggest we remove that `#define` here and add the newer one to the CMake config.

joto:
Superseeded by #1455 which is now merged.

PR #1460, opened 2021-04-24
Remove multi output

joto:
Removes the multi output as preparation for new middle implementations.


pnorman:
👍 in principle, haven't reviewed that everything is removed.

joto:
I compiled with coverage and went through the report, so I believe I found everything that can be removed.

mmd-osm:
Some additional locations that could be removed now:
* https://github.com/joto/osm2pgsql/blob/remove-multi/docs/multi.md
* https://github.com/joto/osm2pgsql/blob/remove-multi/tests/common-options.hpp#L61-L67

joto:
Thanks for your eagle eyes @mmd-osm, I have added a commit removing these.

PR #1469, opened 2021-04-29
Refactoring of WKB code

joto:
These commits refactor some of the WKB code step by step to
* make it cleaner
* pull generation of points geometries out of the `ewkb::writer_t` class to make its use faster and simpler (we'll use this in the upcoming new pgsql middle)
* fix potential buffer overflow in `ewkb::parser_t`

This leaves the WKB code a bit cleaner, there is still the `get_area()` function etc. to clean up which doesn't really belong in the `parser_t` class and some of the code should probably be moved into a `.cpp` file, but I leave that for another day.

### DIFF
--- a/src/wkb.hpp
+++ b/src/wkb.hpp
@@ -43,6 +43,44 @@ enum wkb_byte_order_type_t : uint8_t
 #endif
 };
 
+template <typename T>
+static void str_push(std::string *str, T data)
+{
+    assert(str);
+    str->append(reinterpret_cast<char const *const>(&data), sizeof(T));
+}
+
+/**
+ * Add a EWKB header to the string.
+ *
+ * \pre \code str != nullptr \endcode
+ */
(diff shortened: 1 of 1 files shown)


src/wkb.hpp
```diff
@@ -230,6 +214,22 @@ class writer_t
...
+    std::size_t m_multigeometry_size_offset = 0;
+    std::size_t m_ring_size_offset = 0;
+
+    void set_size(std::size_t offset, std::size_t size)
+    {
+        assert(m_data.size() >= offset + sizeof(uint32_t));
+        auto const s = static_cast<uint32_t>(size);
+        std::memcpy(&m_data[offset], reinterpret_cast<char const *>(&s),
+                    sizeof(uint32_t));
+    }
```
lonvia, line 232: Functions should come before members.

joto: Fixed in the original commit. I rebased to current master and force-pushed this PR.

PR #1475, opened 2021-05-01
Add rapidjson lib to contrib and use it for flex output

joto:
This adds the rapidson JSON library to contrib and uses it for supporting JSON/JSONB database columns in the flex output. We will also need the JSON support soon for the new database-backed-middle.

Note that this introduces a breaking change to the flex backend. Before this you had to use Lua to encode data for JSON/JSONB columns in Lua and use the already encoded string to the `add_row()` function. Now this is done for you behind the scenes.


joto:
Needs #1476.

PR #1478, opened 2021-05-03
Added option to skip clustering.

doskabouter:
This would partially fix https://github.com/openstreetmap/osm2pgsql/issues/27

### DIFF
--- a/docs/osm2pgsql.md
+++ b/docs/osm2pgsql.md
@@ -277,6 +277,9 @@ mandatory for short options too.
     the pgsql output (default is no schema, i.e. the `public` schema
     is used).
 
+\--skip-clustering
+:   Do not cluster tables after importing them.
+
 # EXPIRE OPTIONS
 
 -e, \--expire-tiles=[MIN_ZOOM-]MAX-ZOOM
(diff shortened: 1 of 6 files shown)


pnorman:
If clustering is skipped do we want to still create the initial tables as unlogged?

doskabouter:
No, but that is also in this pr (https://github.com/openstreetmap/osm2pgsql/pull/1478/commits/a8180eb902d9b01092eda414a46f4180605fec77#diff-b88247c729c0441b6bab3e9885aaadb052278b7e495af54f9f9f5cb0b852990aR107)

pnorman:
👍 in principle, haven't tested it yet.

lonvia:
The PR itself is okay but at this point I have strong reservations about adding any new features to the psql output at this point, even more so when the new feature adds yet another command line option. With the flex output out of experimental mode, it is the one where the future lies. I'd like to see the psql output slowly phased out or at least frozen at its current state.

Btw the flex output already has the code to not cluster tables. It just doesn't have an option yet to explicitly disable the clustering. It just disables it automatically for tables that have no geometry column. The thinking was that rather than a no-clustering option there should be a custom-clustering option where clustering can be done by any means including none at all.

joto:
The flex output now has the option to disable clustering (#1504). For the reasons mentioned by @lonvia we are not adding this to the pgsql output.

PR #1484, opened 2021-05-08
Improved geometry IsValid check

joto:
Always use trigger and don't check point geometries (in flex output).

### DIFF
--- a/tests/test-output-flex-schema.cpp
+++ b/tests/test-output-flex-schema.cpp
@@ -29,14 +29,14 @@ TEST_CASE("config with schema should work")
     REQUIRE(1 == conn.get_count("pg_namespace", "nspname = 'myschema'"));
     REQUIRE(1 == conn.get_count("pg_tables", "schemaname = 'myschema'"));
 
-    REQUIRE(1362 == conn.get_count("myschema.osm2pgsql_test_point"));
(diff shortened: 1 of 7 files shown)


tests/test-output-flex-schema.cpp
```diff
@@ -31,12 +31,12 @@ TEST_CASE("config with schema should work")
...
 
-    REQUIRE(1 ==
+    REQUIRE(0 ==
             conn.get_count("pg_proc",
                            "proname = 'osm2pgsql_test_point_osm2pgsql_valid'"));
 
-    REQUIRE(1 == conn.get_count("pg_trigger"));
-    REQUIRE(1 ==
+    REQUIRE(0 == conn.get_count("pg_trigger"));
+    REQUIRE(0 ==
```
lonvia, line 39: So you don't test at all anymore that the trigger ends up in the right schema?

lonvia (commented):
Should the trigger be dropped when --drop is given? That would correspond to the previous behaviour.

tests/test-output-flex-schema.cpp
joto: Ups. Using the linestring table now fixes this.

joto:
Function/trigger is now dropped correctly.

PR #1486, opened 2021-05-11
New thread pool implementation

joto:
Change the thread pool implementation giving us several new features:
* Threads are now numbered and when debug logging is used, the thread
  number is reported in each log line.
* Each task running in the thread pool is timed. This allows for
  better logging where our time is spent.
* When submitting a task we now return a future that can be used
  to access the running time of the thread. The code in the main
  thread can also wait for this future to make sure the background
  task is done.

This switches the minimum C++ version needed from 11 to 14 because we
need lambda capture by move.


lonvia (approved):
Needs rebasing before it can be merged.

PR #1496, opened 2021-05-14
Flex config

joto:
* Test more of the flex config files
* Use jsonb instead of hstore in most flex config files

### DIFF
--- a/flex-config/attributes.lua
+++ b/flex-config/attributes.lua
@@ -11,7 +11,7 @@ local srid = 4326
 local tables = {}
 
 tables.points = osm2pgsql.define_node_table('points', {
-    { column = 'tags', type = 'hstore' },
+    { column = 'tags', type = 'jsonb' },
     { column = 'geom', type = 'point', projection = srid },
     { column = 'version', type = 'int' },
     { column = 'changeset', type = 'int' },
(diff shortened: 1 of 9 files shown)


mboeringa:
@joto, while I understand the desire to add a jsonb as an additional type of tag storage column, wouldn't it be better to stick with hstore as default? 

hstore has been the default for quite some time, and many people may have written quite extensive SQL statements using the required syntax. Switching to jsonb might break quite a lot of existing workflows.

From a pure user perspective, there is little to gain by switching to jsonb. hstore offers most of what users likely need.

joto:
There is only one reason to stay with hstore: "we have always done it this way". Everything else points to json: better support in PostgreSQL, builtin (no "CREATE EXTENSION" needed), "industry standard", easier to understand for new users. I believe the advantages are large enough to make this the new standard. In a few years, nobody will even remember "the old ways" of doing things. But, its only in config files, so everybody can make the choice if and when to switch as they see fit. But default configs and the documentation should point to whatever is currently considered best and recommended usw.

mboeringa:
Ok, reasonable arguments. Curious when all this is going to land...

mboeringa:
@joto 

As to the switch to json, this does raise one question: since json supports nested (dictionary like) structures, I wondered how the upcoming new implementation saves the tags? 

I guess it is just a flat structure, because I don't see any use for a nested structure in the context of OSM objects, so I can use:

`tags ->> 'seasonal'`

json syntax instead of the equivalent:

`tags -> 'seasonal'`

for hstore?

Or do you plan to have a root level, and I do need slightly more complex syntax than this?

Issue #965, opened 2019-10-29
Large object id triggers segfault


mmd-osm:
While debugging some other issue, I noticed that particularly large object ids trigger a segfault in `node_ram_cache::set_dense(long, osmium::Location const&)` at node-ram-cache.cpp:141 (tested on 8d9087f1111f4a062158e8e6b10bfbceed90899b ). I'd rather expect an error message here stating that the object id is too large.

This is not really urgent in any way, as it might only affect people mixing OSM with some other data. I thought I'd raise it anyway, as segfaults are generally not so nice.

Test example: (`osm2pgsql -d osm  tesfile.osm`).

```
<?xml version='1.0' encoding='UTF-8'?>
<osm version='0.6' generator='JOSM'>
  <node id='540000000001' timestamp='2015-01-01T00:00:00Z' uid='1' user='3' version='1' changeset='10' lat='50' lon='7'/>
</osm>
```

lonvia:
Duplicate of #290. Leaving this one open though as it has a better description.

vincnetas:
Why is this "Duplicate of #290" But in #290 it says that "Closing in favour of #965."
Looks like cyclical reference.
And yes, came here because having Segfaults when importing OSM data with lua filter.

joto:
@vincnetas It is the nature of duplicates, that if one is the duplicate of the other, the other is a duplicate of the first. Doesn't matter which way you look at it. The other one is closed and this one is open. So all is in order.

jochenhebbrecht:
Hi,

I'm currently using `osm2pgsql` version `1.4.0` to load a PBF file which has OSM node ids > 10.000.000.000.000.
It fails with:
```
# osm2pgsql -H localhost -P 5432 -U postgres -W /tmp/world.osm.pbf
2021-02-22 15:54:10  osm2pgsql version 1.4.0
Password:
2021-02-22 15:54:13  Database version: 9.5.16
2021-02-22 15:54:13  Node-cache: cache=800MB, maxblocks=12800*65536, allocation method=3
2021-02-22 15:54:13  Setting up table 'planet_osm_point'
2021-02-22 15:54:13  Setting up table 'planet_osm_line'
2021-02-22 15:54:13  Setting up table 'planet_osm_polygon'
2021-02-22 15:54:13  Setting up table 'planet_osm_roads'
Segmentation fault (core dumped)
```
Is the fix described in #290 already in `1.4.0`?

Jochen

joto:
[...]

PR #1499, opened 2021-05-16
Test large ids (node location store)

joto:


mmd-osm:
Can we close #965 now?

joto:
@mmd-osm : When #1500 has landed, we can probably close #965, because then all the code that triggered that error there is gone.

PR #1506, opened 2021-05-18
Modernize the password prompt code

joto:
Note that this hasn't been tested on anything but a Linux system.

Somebody needs to test this on Windows before we can remove the draft status. To test this: Run osm2pgsql with the `-W` option, make sure it asks for a password, password should not be echoed to the screen and connecting to the db with that password should work.


mmd-osm:
Testing https://ci.appveyor.com/project/openstreetmap/osm2pgsql/builds/39212255/artifacts on Win10 works ok, password prompt is shown, password is hidden and later used to connect to the db.  Passwords with non-US ASCII characters don't appear to work though. I don't consider this as a regression, as it didn't work on a build (https://ci.appveyor.com/project/openstreetmap/osm2pgsql/builds/39218821/artifacts) either.

joto:
Thanks @mmd-osm . This sounds good.

mmd-osm:
By the way, you should see the same results when running those Windows builds using _wine_ on Linux.

joto:
Good suggestion, but the problem is I can never be sure if there isn't something different between Wine and Windows hat makes it work or not work in that case, so I don't want to rely on that. (Plus its a bother... :-)

PR #1510, opened 2021-05-26
Add sql_type column field in flex output

joto:
The "type" field in a flex output column definition really does two
things:
1. It defines the way Lua values are transformed into PostgreSQL
   values.
2. It defines which PostgreSQL type should be used in a the CREATE
   TABLE statement for a particular column.

This commit adds a new "sql_type" field which takes over the second
meaning. If it is not specified, the default is still defined by
the "type" field. The default for the type field is "text".

The result is that in most cases you use only the "type" field as before
and it does the right thing. If you need a special SQL type, use the
"sql_type" field instead or in addition to the "type" field.

Note that this is a breaking change, but still allowed because the flex
output is still marked experimental.


mboeringa:
Looks like a useful addition

PR #1521, opened 2021-06-08
Bugfix: Do not create table as unlogged if it is not clustered

joto:
Tables created with the flex output that were marked `cluster: "no"`
were created as unlogged tables. This should only be done for tables
which will be replaced by their clustered copy later on.

See #27

### DIFF
--- a/src/flex-table.cpp
+++ b/src/flex-table.cpp
@@ -173,7 +173,7 @@ void table_connection_t::start(bool append)
 
     if (!append) {
         m_db_connection->exec(table().build_sql_create_table(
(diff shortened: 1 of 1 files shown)


pnorman:
Do we have the ability to end up with a final table that is UNLOGGED, like the pgsql backend does?

joto:
@pnorman It does? I would consider this a bug then.

pnorman:
Never mind, we [removed unlogged](https://github.com/openstreetmap/osm2pgsql/issues/940#issue-477610086)

pnorman (approved):

PR #1547, opened 2021-08-04
Change normal output to go to stdout for osm2pgsql-replication

amandasaurus:
With this patch, regular default output from `osm2pgsql-replication` will go to stdout, not stderr, which is the unix standard.


lonvia:
I disagree. Logging output usually goes to stderr. osm2pgsql does exactly the same.

lonvia:
I'd prefer to keep the behaviour consistent with osm2pgsql. So, I'll take the liberty to reject this.

PR #1548, opened 2021-08-04
add a 'status' command which prints the lag (& in JSON)

amandasaurus:
This patch adds an `info` command which prints readable summary of the state of replication:

```
$ osm2pgsql-replication info -U osm -d gis
2021-08-04 14:12:28 [INFO]: Using replication service 'https://download.geofabrik.de/europe-updates/'. Current sequence 3052 (2021-08-03T20:21:39Z).
2021-08-04 14:12:28 [INFO]: Server is at sequence 3052 (2021-08-03T20:21:39Z)
2021-08-04 14:12:28 [INFO]: Server's most recent data is 17 hour(s) 50 minutes(s) old
2021-08-04 14:12:28 [INFO]: Database is up to date with server
2021-08-04 14:12:28 [INFO]: Database's most recent data is 17 hour(s) 50 minutes(s) old
```

with the `--json` option, the output is in JSON format, which is useful for writing scripts to detect if the replication has stopped, or to graph how far behind the server is:

`$ osm2pgsql-replication info -U osm -d gis --json`:
```json
{
  "server": {
    "base_url": "https://download.geofabrik.de/europe-updates/",
    "sequence": 3052,
    "timestamp": "2021-08-03T20:21:39Z",
    "age": {
      "seconds": 64920,
      "text": {
        "en": "18 hour(s) 2 minutes(s)"
      }
    }
  },
  "local": {
    "sequence": 3052,
    "timestamp": "2021-08-03T20:21:39Z",
    "age": {
      "seconds": 64920,
      "text": {
        "en": "18 hour(s) 2 minutes(s)"
      }
    }
  },
  "diff": {
    "sequence": 0,
    "timestamp": {
      "seconds": 0,
      "text": {
        "en": "0 minutes(s)"
      }
    }
  }
}
```

### DIFF
--- a/scripts/osm2pgsql-replication
+++ b/scripts/osm2pgsql-replication
@@ -39,6 +39,28 @@ from osmium import WriteHandler
 
 LOG = logging.getLogger()
 
+def pretty_format_timedelta(seconds):
+    minutes = int(seconds/60)
+    (hours, minutes) = divmod(minutes, 60)
+    (days, hours) = divmod(hours, 24)
+    (weeks, days) = divmod(days, 7)
+
+    if 0 < seconds < 60:
+        output = "<1 minute"
+    else:
+        output = []
+        # If weeks > 1 but hours == 0, we still want to show "0 hours"
+        if weeks > 0:
+            output.append("{} week(s)".format(weeks))
+        if days > 0 or weeks > 0:
+            output.append("{} day(s)".format(days))
+        if hours > 0 or days > 0 or weeks > 0:
+            output.append("{} hour(s)".format(hours))
+
+        output.append("{} minute(s)".format(minutes))
+        output = " ".join(output)
+    return output
+
 def connect(args):
     """ Create a connection from the given command line arguments.
     """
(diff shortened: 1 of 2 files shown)


amandasaurus:
Please don't merge this now. I have unintentionally added something to the man page for `--post-processing`.

lonvia:
> Please don't merge this now. I have unintentionally added something to the man page for --post-processing.

That part is okay. The man page wasn't regenerated last time.

joto:
How about naming this `status` instead of `info`? `info` is very close to `init`, especially when some kind of command line completion is involved. Also `status` is more in line with what, for instance, `systemctl` or `git` use.

amandasaurus:
I have renamed it to `status` (& force pushed the branch to update this PR)

scripts/osm2pgsql-replication
```diff
@@ -115,6 +134,84 @@ def update_replication_state(conn, table, seq, date):
 
     conn.commit()
 
+def status(conn, args):
+    """\
+    Print information about the current replication status
+    """
```
lonvia, line 140: I could swear the initial version had more documentation in the man page. Extended documentation (including a description of the json output should be added here and will then automatically appear in the man page.

Also: please finish the sentence with a full stop.

scripts/osm2pgsql-replication
```diff
@@ -340,6 +437,16 @@ def get_parser():
     cmd.add_argument('--post-processing', metavar='SCRIPT',
                      help='Post-processing script to run after each execution of osm2pgsql.')
 
+    # Arguments for status
+    cmd = subs.add_parser('status', parents=[default_args],
+                          help=status.__doc__.split('\n', 1)[0],
+                          description=dedent(status.__doc__),
+                          formatter_class=RawDescriptionHelpFormatter,
+                          add_help=False)
+    cmd.add_argument('--json', action="store_true", default=False, help="Output status as json")
```
lonvia, line 446: Full stop missing in help text.

scripts/osm2pgsql-replication
```diff
@@ -39,6 +39,25 @@ from osmium import WriteHandler
...
+    output = []
+    # If weeks > 1 but hours == 0, we still want to show "0 hours"
+    if weeks > 0:
+        output.append("{} week(s)".format(weeks))
+    if days > 0 or weeks > 0:
+        output.append("{} day(s)".format(days))
+    if hours > 0 or days > 0 or weeks > 0:
+        output.append("{} hour(s)".format(hours))
+
+    output.append("{} minutes(s)".format(minutes))
```
lonvia, line 57: minutes -> minute

scripts/osm2pgsql-replication
```diff
@@ -115,6 +134,84 @@ def update_replication_state(conn, table, seq, date):
...
+                results['server']['sequence'] = server_seq
+
+                results['server']['timestamp'] = server_ts.strftime("%Y-%m-%dT%H:%M:%SZ")
+                server_age = dt.datetime.now(dt.timezone.utc) - server_ts
+                results['server']['age'] = {}
+                results['server']['age']['seconds'] = int(server_age.total_seconds())
+                results['server']['age']['text'] = {'en': pretty_format_timedelta(server_age) }
+
+                results['local']['age'] = {}
+                local_age = dt.datetime.now(dt.timezone.utc) - db_ts
```
lonvia, line 180: `now()` should be called only once or you might end up with odd discrepancies for the age.

lonvia (changes requested):
This is okay in principle but I would like to discuss the JSON output. I find the duplicated information in the age fields rather strange. If you have a computer-readable format, then providing the computer-readable output should be enough. Pretty-printing should be done on the client side. So I would just do `age: <age in seconds>`. Same goes for the `diff` section. It's trivial for a client to subtract sequence numbers or age seconds. No need to duplicate that on the server side.

scripts/osm2pgsql-replication
amandasaurus: You can see the original version here. It had the same (low level of) detail https://github.com/openstreetmap/osm2pgsql/commit/eb4c79c4625c9fc384a3e1701c9f3a4fd07194f6

amandasaurus:
I am working on your feedback. In order to track things, I will not rebase now. (Maybe we can rebase at end)

amandasaurus:
> I would like to discuss the JSON output

How about this for an output format?

```{
  "server": {
    "base_url": "https://download.geofabrik.de/europe-updates/",
    "sequence": 3052,
    "timestamp": "2021-08-03T20:21:39Z",
    "age_sec": 64920
  },
  "local": {
    "sequence": 3052,
    "timestamp": "2021-08-03T20:21:39Z",
    "age_sec":  64920
  }
}
```

I prefer `age_secs` rather than just `age`, so that people will know the units (seconds) without having to look it up. Someone might think the `age` is the age in sequence numbers (for example).

`timestamp` is the timestamp of the server/local db, and `age_sec` is the difference between that timestamp, and now. This does mean `age_sec` is duplicating information that could be derived from the `timestamp` field. However given that parsing ISO timestamps & doing date diffs takes a little more than zero work, I think it is beneficial for the user to do the work in the `status` command.

If that output format is OK, I can implement it.

scripts/osm2pgsql-replication
```diff
@@ -115,6 +134,84 @@ def update_replication_state(conn, table, seq, date):
...
+
+                base_url, db_seq, db_ts = cur.fetchone()
+                db_ts = db_ts.astimezone(dt.timezone.utc)
+                results['server']['base_url'] = base_url
+                results['local']['sequence'] = db_seq
+                results['local']['timestamp'] = db_ts.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+                repl = ReplicationServer(base_url)
+                server_seq, server_ts = repl.get_state_info()
```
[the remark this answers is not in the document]
amandasaurus: Done. I had presumed that pyosmium would raise an exception. 😉

scripts/osm2pgsql-replication
```diff
@@ -115,6 +134,84 @@ def update_replication_state(conn, table, seq, date):
...
+        LOG.info("Server's most recent data is %s old", results['server']['age']['text']['en'])
+
+        if results['local']['sequence'] == results['server']['sequence']:
+            LOG.info("Database is up to date with server")
+        else:
+            LOG.info("Database is %d sequences behind, i.e. %s",
+                    results['diff']['sequence'],
+                    results['diff']['timestamp']['text']['en'])
+
+        LOG.info("Database's most recent data is %s old", results['local']['age']['text']['en'])
```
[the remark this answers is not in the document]
amandasaurus: For the JSON output, it does use `print(…)`. For the default "human readable format", it uses `LOG.info/fatal`

lonvia:
That json format looks good to me. One more suggestion: how about adding a status field that is also returned in the case of an error? It would make it easier to check for success. Alternatively, you could always return the error code and make it 0 on success.

amandasaurus:
I've changed the json output as requested.

I've added a `status` key to the JSON output, which is always the same as the exit code. so `0` is "OK", and then errors are `1`, `2` or `3`.

I've written help docs & updated man page

If you don't want any more changes, should I rebase again master? or squash? or just leave as it (i.e. you can merge it :) )

scripts/osm2pgsql-replication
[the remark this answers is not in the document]
lonvia: This is the last bit where we disagree. Given that the output is the direct product of the command, it should go directly to stdout and not through the logging subsystem. So my preferred handling would be to leave `LOG.fatal()` in place but replace the `LOG.info()` with `print()`.

lonvia:
Merging the other PR now causes conflicts for the man page. Can you please rebase? Stashing would not be a bad idea either but I leave that for your personal preferences.

amandasaurus:
Squashed it all into one commit off current `master`.

scripts/osm2pgsql-replication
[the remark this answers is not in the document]
amandasaurus: My apologies, I misunderstood you. I thought you were talking about the JSON. I have made the change you have requested.

lonvia (approved):

joto:
I wonder if there is value in having this command optionally not contacting the remote server. If you only want to check what the status of the local database is, not contacting the server would eliminate a potential source of problems. Might be useful when this is run in some kind of monitoring environment. Then again, maybe not. Could be an option `--local-only`, or the other way around `--remote` or something like it?

amandasaurus:
> I wonder if there is value in having this command optionally not contacting the remote server. 

That functionality is not needed for my use case, so I have not added it. It might be useful for someone. Will yous only merge this PR if I add that functionality? 🙂

joto:
Fair enough. We'll leave this as "future work" then if and when somebody needs it.

PR #1568, opened 2021-09-03
Add option for cutom OSM API server in replication script

jbronn:
Analogous to the `--server` option, this PR adds an `--osm-server` that allows overriding the default when running `osm2pgsql-replication init`.

### DIFF
--- a/scripts/osm2pgsql-replication
+++ b/scripts/osm2pgsql-replication
@@ -68,7 +68,7 @@ def connect(args):
                             host=args.host, port=args.port)
 
 
-def compute_database_date(conn, prefix):
+def compute_database_date(conn, prefix, osm_server):
     """ Determine the date of the database from the newest object in the
         database.
     """
@@ -84,7 +84,7 @@ def compute_database_date(conn, prefix):
 
     LOG.debug("Using way id %d for timestamp lookup", osmid)
     # Get the way from the API to find the timestamp when it was created.
-    url = 'https://www.openstreetmap.org/api/0.6/way/{}/1'.format(osmid)
+    url = '{}/api/0.6/way/{}/1'.format(osm_server, osmid)
     headers = {"User-Agent" : "osm2pgsql-update",
                "Accept" : "application/json"}
     with urlrequest.urlopen(urlrequest.Request(url, headers=headers)) as response:
@@ -263,7 +263,7 @@ def init(conn, args):
     this with the '--server' parameter.
     """
     if args.osm_file is None:
-        date = compute_database_date(conn, args.prefix)
+        date = compute_database_date(conn, args.prefix, args.osm_server)
         if date is None:
             return 1
 
@@ -439,6 +439,9 @@ def get_parser():
                           formatter_class=RawDescriptionHelpFormatter,
                           add_help=False)
     grp = cmd.add_argument_group('Replication source arguments')
+    grp.add_argument('--osm-server', metavar='URL',
+                     default='https://www.openstreetmap.org',
+                     help='Use OpenStreetMap server at the given URL (default: %(default)s)')
     srcgrp = grp.add_mutually_exclusive_group()
     srcgrp.add_argument('--osm-file', metavar='FILE',
                         help='Get replication information from the given file.')


lonvia:
This is one of these PRs that in theory ticks all the boxes for being mergable: I can see the use case, the changes are small and clean and the code is well maintainable. Yet, I'm still hesitant to merge this. It does add yet another command line option and the number of users for which this will be useful is very, very small. Many small PRs of this kind are how you end up with the [number of options in bold face](https://github.com/openstreetmap/osm2pgsql/blob/8ecae57db2c5340cd67992a0ca9f99534cf72ede/README.md#usage) because there are so many. So I'd lean towards recommending that you keep your own copy of this script with your private API setup.

@pnorman @joto what do you think?

joto:
I tend to agree with @lonvia: Running your own OSM server is waaay out of the normal use cases we want to support.

ldeffenb:
> I tend to agree with @lonvia: Running your own OSM server is waaay out of the normal use cases we want to support.

Huh?  "Running your own OSM server" is not a "normal use case", but yet with the OSM acceptable use policies, we have no choice.  Is this a case of "have your cake and eat it too"?

lonvia:
@ldeffenb This is about setting up your own OSM **API** server (i.e. https://github.com/openstreetmap/openstreetmap-website/), not about setting up your own tile server or other services that use official OSM data.

ldeffenb:
Thank you for the clarification.  I actually thought it was to support initializing the replication for using a different update server, but I guess that's what the --server option does.  I didn't look closely enough at the use of the new value to see that it was invoking API requests to it.  I guess this would only be really useful for someone setting up an entire toolchain for their own geo-data where their way ID to timestamp lookup would not match the base OSM data source.  I get it now.

pnorman:
Given that this only impacts setup in exceptional cases, I would be okay not merging it. There are other ways to do what is done here.

jbronn:
Yes, this supports the use case of using your own API server.  Besides the Rails API implementation, there's also zerebubuth/openstreetmap-cgimap, the API server used in production by OSM (`api.openstreetmap.org`, different from the hard-coded value here).  I thought this would be OK here because, as I said, support is already present for overriding the default planet server -- why would API server support be different, and an "exceptional use case" in that light?

jbronn:
Let me expand further and respond further to specific issues raised by @lonvia:

> It does add yet another command line option and the number of users for which this will be useful is very, very small.
> Many small PRs of this kind are how you end up with the [number of options in bold face](https://github.com/openstreetmap/osm2pgsql/blob/8ecae57db2c5340cd67992a0ca9f99534cf72ede/README.md#usage).

I'm not patching the `osm2pgsql` command but the `osm2pgsql-replication init` *subcommand* -- a script just added [April 7, 2021 in 1.4.2](https://github.com/openstreetmap/osm2pgsql/releases/tag/1.4.2) and automates a niche operations process for replicating tile databases.  No major Linux distributions package `osm2pgsql-replication` yet.¹  In fact, I have a hunch that more users run custom API instances than those with custom planet replication endpoints, a feature the `--server` option already supports.²  Thus, I suspect the existing user base for `osm2pgsql-replication` is already very, very small.

Moreover, the [OSM Wiki itself lists the API URI as `api.openstreetmap.org`](https://wiki.openstreetmap.org/wiki/API_v0.6#URL_.2B_authentication) and `masters.api.devs.openstreetmap.org` for development usage.  This clearly demonstrates that multiple public API URLs other than `www.openstreetmap.org` is reality and not a hypothetical.³

If there there are further improvements that can be resolved technically here (e.g., I noticed I neglected to update the docstring) then I'd like to continue as a new contributor.  However, I can't resolve an objection based on a belief that *this* particular feature starts the nebulous "slippery slope" of feature creep and unreasonable maintenance burden for this project.  If so, I'll close out so we can all focus our time on more productive efforts.

¹ Ubuntu's latest is [1.4.1](https://launchpad.net/ubuntu/+source/osm2pgsql) while Fedora 36 has 1.5.1 available, they haven't updated their [`osm2pgsql.spec`](https://src.fedoraproject.org/rpms/osm2pgsql/blob/main/f/osm2pgsql.spec) to include `osm2pgsql-replication` yet.
² Unlike the dedicated projects for the [Rails website](https://github.com/openstreetmap/openstreetmap-website) or [CGImap](https://github.com/openstreetmap/cgimap), the only "implementation" of a planet server I know of has to be assembled manually by grokking [Chef configuration](https://github.com/openstreetmap/chef/tree/master/cookbooks/planet) and [database tools](https://github.com/openstreetmap/osmdbt).
³ *See also* OSM's Apache [backend](https://github.com/openstreetmap/chef/blob/master/cookbooks/web/templates/default/apache.backend.erb#L8) and [frontend](https://github.com/openstreetmap/chef/blob/master/cookbooks/web/templates/default/apache.frontend.erb#L8) server configurations.

mmd-osm:
>  I have a hunch that more users run custom API instances than those with custom planet replication endpoints

The main use case here is probably to keep your extract up to date by using dedicated endpoints such as http://download.geofabrik.de/europe-updates/

Still in this case you want to validate OSM objects against the main OSM API, as that's the original source for all extracts.

As a side note, backend servers are no longer, everything (Rails + CGImap) is running on frontend servers these days. Then masters.api.devs.openstreetmap.org is probably a bad example, as none of the diff replication + planet generation code is running there.

lonvia:
This discussion has brought to light another very strong point against merging this PR: 99% of osm2pgsql users wouldn't even consider that such a thing as custom API instance exists. There is a good chance that the additional option will lead to more confusion than good.

Just to be clear: we provide the replication script simply for convenience to make the very frequently used task of updating the database easier. There is nothing special about the script. It does not rely on any osm2pgsql interna which require that it is in sync with the osm2pgsql source code. So it can pretty much stand on its own.

We simply hope that it is useful for the 90% of users with a more or less standard setup. For those, where it doesn't work: consider the script as an example on how to set up your own replication. Or you can continue to use the other existing tool chains for database update based on pyosmium-get-changes or osmosis. Nothing wrong with that either.

PR #1574, opened 2021-10-01
ci: improve the workflow

chenrui333:
- remove LUA_VERSION env for macos build, https://github.com/Homebrew/homebrew-core/pull/82233 (it builds/runs fine with lua 5.4?)
- remove `test-wrapper` input arg (use the default value as it already did)
- fix the job ref for `ubuntu20-pg13-gcc10-release`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,12 +4,7 @@ on: [ push, pull_request ]
 
 jobs:
   macos:
-
-    env:
-      LUA_VERSION: 5.2
-
     runs-on: macos-latest
-
     steps:
       - uses: actions/checkout@v2
 
@@ -30,8 +25,6 @@ jobs:
         shell: bash
 
       - uses: ./.github/actions/build-and-test
-        with:
-          test-wrapper:
         env:
           PGHOST: /tmp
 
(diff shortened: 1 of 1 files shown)


lonvia:
> * remove LUA_VERSION env for macos build, osm2pgsql 1.5.1 Homebrew/homebrew-core#82233 (it builds/runs fine with lua 5.4?)

Feel free to switch to 5.4 but I would feel more confortable if we keep the tested version pinned. The CI setup is fragile enough as it is.

> * remove test-wrapper input arg (use the default value as it already did)

It used the empty value not the default value. The point is to disable use of a test wrapper 'pg_virtualenv' which does not exist on macos.

chenrui333:
> > * remove LUA_VERSION env for macos build, osm2pgsql 1.5.1 [osm2pgsql 1.5.1 Homebrew/homebrew-core#82233](https://github.com/Homebrew/homebrew-core/pull/82233) (it builds/runs fine with lua 5.4?)
> 
> Feel free to switch to 5.4 but I would feel more confortable if we keep the tested version pinned. The CI setup is fragile enough as it is.
> 
> > * remove test-wrapper input arg (use the default value as it already did)
> 
> It used the empty value not the default value. The point is to disable use of a test wrapper 'pg_virtualenv' which does not exist on macos.

Thanks for the context, I am gonna refine my build now.

PR #1583, opened 2021-11-29
Fix logging while progress display is on

joto:
When there is an active progress display, log messages show up after the progress display instead of the next line. This adds the necessary newline character to fix that.


src/progress-display.hpp
```diff
@@ -23,6 +23,8 @@
 
 #include <osmium/handler.hpp>
 
+#include "logging.hpp"
```
lonvia, line 26: Still needed after the last change?

PR #1596, opened 2021-12-20
Modernize cmake

joto:
* Version 3.5.0 is now required, this is available even in the rather old Ubuntu 16.04 distribution, so this isn't a real restriction.
* Also sets the oldest supported PostgreSQL version to 9.6 which was already documented for a while.
* Sets minimum libosmium version to 2.17.2 which is included anyway.
* Update list of clang-tidy versions we are looking for.

The second commit adds another build to the Github actions config building in C++17 mode.


joto:
Set 9.5 as minimum version not 9.6, so we still work on Ubuntu 16.04.

.github/workflows/ci.yml
```diff
@@ -179,6 +179,25 @@ jobs:
...
 
+  ubuntu20-pg13-clang10-cpp17:
+    runs-on: ubuntu-20.04
+
+    env:
+      CC: clang-10
+      CXX: clang++-10
+      LUA_VERSION: 5.3
+      LUAJIT_OPTION: OFF
+      POSTGRESQL_VERSION: 13
```
lonvia, line 190: If you add a new one here. Please test against POSTGRESQL_VERSION 14.

PR #1608, opened 2022-01-14
Again: ignore relations with more than 32k members

### LINKED ISSUE #1607: osm2pgsql fails with large relations
Someone has recently created a relation with more than 200k members: https://www.openstreetmap.org/relation/13671879 (now deleted) This causes osm2pgsql to fail with the following message:

```
ERROR: DB copy thread failed: Ending COPY mode for 'planet_osm_rels' failed: ERROR:  value "218678" is out of range for type smallint
CONTEXT:  COPY planet_osm_rels, line 121, column way_off: "218678"
```

osm2pgsql should reject all relations that have more than 10k (50k? 100k?) members and not try to process them.

If you run into this bug: make sure to download a diff range large enough that it includes the deletion of the relation that followed about 40min later.

joto:
Reintroduce fix from 8763ce951bddee2af553fab6ce4e1f4e79eeaaa0 which got
moved to the wrong place in some refactoring.

> There is a hard restriction in the pgsql middle tables, where
> relation members are indexed using a smallint. To remain
> consistent, drop all relations which exceed this size.

Fixes #1607

### DIFF
--- a/src/osmdata.cpp
+++ b/src/osmdata.cpp
@@ -88,6 +88,13 @@ void osmdata_t::after_ways() { m_mid->after_ways(); }
 
 void osmdata_t::relation(osmium::Relation const &rel)
 {
+    if (rel.members().size() > 32767) {
+        log_warn(
+            "Relation id {} ignored, because it has more than 32767 members",
+            rel.id());
+        return;
+    }
+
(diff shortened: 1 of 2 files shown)


lonvia:
Can we have a test, so there won't be an accidental regression again?

PR #1619, opened 2022-01-20
Remove dependency on boost::system and boost::filesystem

joto:
Not the cleanest of code with the `ifdef` etc. but it removes a large dependency. If and when we switch to C++17, we can revert that and use std::filesystem.


joto:
Not worth changing this now when we are switching to C++17 soon which has std::filesystem.

PR #1620, opened 2022-01-20
Remove dep on boost property tree

joto:


joto:
The problem is due to Windows defining a macro called `GetObject` mapping to `GetObjectA`. The problem [has been discussed in rapidjson issue tracker](https://github.com/Tencent/rapidjson/issues/1448) and there is a [fix](https://github.com/Tencent/rapidjson/pull/1877), but because there have been no releases of rapidjson in ages we are stuck with it.

PR #1623, opened 2022-01-21
Install osm2pgsql-replication script/man page on make install

joto:
This also adds a new Github actions workflow which tests the
installation of osm2pgsql and osm2pgsql-replication.


pnorman:
This should wait till after the next bugfix

joto:
@pnorman Why? Is this is some way critical? It only adds functionality (that was basically there all along).

PR #1626, opened 2022-01-24
Fix manpage issues reported by lintian.

sebastic:
The lintian QA tool reported a few issues with the `osm2pgsql-replication` manpage:

 * [bad-whatis-entry](https://lintian.debian.org/tags/bad-whatis-entry)
   The `NAME` section requires the format: `<name> - <description>`
 * [groff-message](https://lintian.debian.org/tags/groff-message): warning: macro '&lt;foo&gt;'' not defined
   `'` at the start of a line is special, from the [groff(7)](https://manpages.debian.org/bullseye/groff/groff.7.en.html) documentation:
   > The single quote has two controlling tasks. At the beginning of a line and in the conditional requests it is the non-breaking control character. That means that it introduces a request like the dot, but with the additional property that this request doesn't cause a linebreak. By using the `c2` request, the non-break control character can be set to a different character. 
   > As a second task, it is the most commonly used argument separator in some functional escape sequences (but any pair of characters not part of the argument do work). In all other positions, it denotes the single quote or apostrophe character. Groff provides a printable representation with the `\(cq` escape sequence.

### DIFF
--- a/docs/osm2pgsql-replication.1
+++ b/docs/osm2pgsql-replication.1
@@ -1,6 +1,6 @@
 .TH osm2pgsql-replication "1" Manual
 .SH NAME
-osm2pgsql-replication
+osm2pgsql-replication - osm2pgsql database updater
 .SH SYNOPSIS
 .B osm2pgsql-replication
 [-h] {init,update,status} ...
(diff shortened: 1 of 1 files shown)


lonvia:
Thanks for the PR. As the man page is generated by argparse-manpage, we need to find a way to make sure these changes will be reapplied next time the page is generated. I'll look into that.

Leaving the PR open in the meantime as a reminder.

PR #1629, opened 2022-02-01
Fix reprojection when using Proj version 6 and above

joto:
The test was never run so we did not notice that the projection did not
work correctly.

Code modelled after https://proj.org/development/migration.html .

### DIFF
--- a/src/reprojection-generic-proj6.cpp
+++ b/src/reprojection-generic-proj6.cpp
@@ -15,22 +15,10 @@ class generic_reprojection_t : public reprojection
 {
 public:
     explicit generic_reprojection_t(int srs)
-    : m_target_srs(srs), m_context(proj_context_create())
-    {
-        assert(m_context);
-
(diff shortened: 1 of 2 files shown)


joto:
Something is terribly broken here. The projection to EPSG:5520 gets a different result in Ubuntu 18 than in Ubuntu 20. This is not just our code that's broken. I get a different result on [epsg.io](http://epsg.io/transform#s_srs=4326&t_srs=5520&x=10.0000000&y=53.0000000) than when I try this locally in PostgreSQL:

```
select st_astext(st_transform(st_setsrid(st_makepoint(10.0, 53.0), 4326), 5520));
┌─────────────────────────────────────────────┐
│                  st_astext                  │
├─────────────────────────────────────────────┤
│ POINT(1969579.1472745757 5896972.954884315) │
└─────────────────────────────────────────────┘

```

PR #1630, opened 2022-02-02
Use std::filesystem instead of boost::filesystem

joto:


joto:
This doesn't work correctly on Ubuntu 18 with older GCC compilers. `std::filesystem` is only available from glibc 8. But it doesn't seem to work there ending in segfaults. Testing this locally I get compile errors. So maybe we should not inflict this on Ubuntu 18 users.

PR #1635, opened 2022-02-02
Set HAVE_GENERIC_PROJ directly in CMake config

joto:
Instead of going through config.h. It is simpler and including config.h can't be forgotten.


joto:
We might want to do similar things for the other `config.h` options, too. But taking it one step at a time.

PR #1636, opened 2022-02-04
Work-in-progress: Add geometry functions in Lua

joto:
This is intended to show how geometry processing in Lua could work. Only some example functions are implemented and not for all geometry types, error detection is limited etc.
    
Geometry creation functions:
* as_point() (for nodes)
* as_linestring() (for ways)
* as_polygon() (for ways)
* as_multipolygon() (for relations)
    
Geometry functions:
* area() (for polygons)
* centroid() (for polygons)
* simplify() (for linestrings)
    
See the file flex-config/custom-geometries.lua for an example that shows how to use these.

### DIFF
--- a/flex-config/new-insert-syntax.lua
+++ b/flex-config/new-insert-syntax.lua
@@ -1,0 +1,215 @@
+-- This config example file is released into the Public Domain.
+
+-- This is Lua config showing the handling of custom geometries.
+--
+-- This is "work in progress", changes are possible.
+--
+-- Note that the insert() function is used instead of the old add_row()!
+
+-- The following magic can be used to wrap the internal "insert" function
+local orig_insert = osm2pgsql.Table.insert
+local function my_insert(table, data)
+    local json = require 'json'
+    inserted, message, column, object = orig_insert(table, data)
+    if not inserted then
+        for key, value in pairs(data) do
+            if type(value) == 'userdata' then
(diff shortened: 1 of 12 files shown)


joto:
There are some issues with this approach:
* Performance: It is unclear whether all of this processing and jumping between C++ and Lua is fast enough. This needs more testing.
* Projections: Currently you can set the projection only in the table definition and it will be done automatically when a geometry is set. This is probably not enough, for instance when calculating an area in a different projection, we need to be able to project first, then calculate the area. So something like this would probably need adding: `geom:as_polygon():transform(3857):area()`.
* Backwards compatibility and "magic" geometry processing: The old way of handling geometries was somewhat magic, for instance long lines could be split and then `add_row()` would add multiple rows with those split geometries. The "new" way doesn't do that any more. We'd have a `split_linestrings()` function in Lua and then need to write a loop in Lua to add several rows. This makes certain types of processing harder to write in the config file and (probably) slower to process. But it allows more flexibility, for instance now multiple geometry columns are easily done, because the new geometry column types behave just like any other column type. It is difficult to wrap your head around all this, especially when old and new geometry types are mixed.

wambacher:
would like to have pointonsurface(geom) too, because the centroid of an admin boundary may be outside of the area.

See https://www.openstreetmap.org/relation/408592

regards
walter

PR #1648, opened 2022-03-03
Refactor expire code

joto:
This disentangles the WKB code from the expire code. There are further refactorings to be done, but they will come in later commits to keep things separate.


src/expire-tiles.cpp
```diff
@@ -134,6 +135,46 @@ void expire_tiles::coords_to_tile(double lon, double lat, double *tilex,
...
+        if (from_bbox(box)) {
+            /* Bounding box too big - just expire tiles on the line */
+            log_debug("Large polygon ({:.0f} x {:.0f} metres, OSM ID {})"
+                      " - only expiring perimeter",
+                      box.max_x() - box.min_x(), box.max_y() - box.min_y(), osm_id);
+            if (geom.is_polygon()) {
+                from_point_list(geom.get<geom::polygon_t>().outer());
+            } else if (geom.is_multipolygon()) {
+                for (auto const &polygon : geom.get<geom::multipolygon_t>()) {
+                    from_point_list(polygon.outer());
```
lonvia, line 171: The original code expires all rings here.

joto: Fixed and force-pushed.

PR #1654, opened 2022-03-14
Use bundled includes correctly

AMDmi3:
Regardless of whether system on bundled depends are used, their includes are used with SYSTEM.
This is not correct for bundled case as it leads to incorrect include order and build problems in presence of system equivalents of affected bundled libraries.
Fix this by including external depends with SYSTEM.


lonvia:
This is not working as intended. See CI run.

joto:
No movement with this PR, so closing. @AMDmi3 please create an issue or new PR if you want to pursue this further.

PR #1656, opened 2022-03-15
Use pointers instead of references for mutable function params

joto:


src/dependency-manager.cpp
```diff
@@ -41,16 +41,15 @@ bool full_dependency_manager_t::has_pending() const noexcept
     return !m_ways_pending_tracker.empty() || !m_rels_pending_tracker.empty();
 }
 
-idlist_t full_dependency_manager_t::get_ids(osmium::index::IdSetSmall<osmid_t> &tracker)
+idlist_t
+full_dependency_manager_t::get_ids(osmium::index::IdSetSmall<osmid_t> *tracker)
 {
-    tracker.sort_unique();
```
lonvia, line 46: The line was deleted intentionally?

joto: Ups. No. I'll get you an updated PR...

PR #1670, opened 2022-04-25
Introduce behaviour-driven testing for integration tests

lonvia:
This is work-in-progress to move the regression tests and all unit test that do a simple `run_file` followed by some database queries with a behaviour-driven test framework.

There are a couple of reasons to do this:
* The `run_file` tests are currently spread out over 3 files which makes them very hard to read. The hand-written XML files are a nuisance. The BDD tests collect all in one place and have helper functions for creating point grids and writing auto-completed OPL.
* Each C++ unit test produces a binary of 80MB. The BDD tests would reduce the size of the build directory and reduce build time.
* The regression tests are extremely hard to read because tests are spread out over the entire file.

The BDD tests use [behave](https://github.com/behave/behave), a Python-port of gherkins. Chosen because I can borrow code from Nominatim's BDD framework. (There is a bit of a danger that behave is at a dead end development-wise. Current version 1.2.6 is from 2019 but it works and has all features we need.)

This is a preview to give you an idea of what the BDD feature files would look like. Currently the regression tests for import and the flex area test are ported. In theory, the same framework could later be used by osm2pgsql users to write tests for their own style files or lua flex configurations. In practice, something like this would require a lot of additional documentation.


lonvia:
This can now be merged from my side. All regression tests have been converted to BDD tests. There are still a couple of useful functions missing to convert more of the unit tests but this can be done in subsequent PRs.

tests/bdd/parameters/dsn.feature
```diff
@@ -0,0 +1,20 @@
...
+
+        When running osm2pgsql pgsql with parameters
+            | -d |
+            | <connection_parameter> |
+
+        Then table planet_osm_point has 1 row
+
+        Examples:
+            | connection_parameter    |
+            | {TEST_DB} |
```
joto, line 17: Add spaces for consistency with other tables.

tests/bdd/environment.py
```diff
@@ -0,0 +1,127 @@
...
+    'TEST_DB': 'osm2pgsql-test',
+    'HAVE_TABLESPACE': True,
+    'HAVE_LUA': True,
+    'HAVE_PROJ': True
+}
+
+use_step_matcher('re')
+
+def _connect_db(context, dbname):
+    """ Connect to the given database and return the conntection
```
joto, line 38: `conn**ection`

tests/bdd/steps/geometry_factory.py
```diff
@@ -0,0 +1,110 @@
...
+           a previously defined grid.
+        """
+        if geom.find(',') < 0:
+            out = "POINT({})".format(self.mk_wkt_point(geom))
+        elif geom.find('(') < 0:
+            out = "LINESTRING({})".format(self.mk_wkt_points(geom))
+        else:
+            out = "POLYGON(({}))".format(self.mk_wkt_points(geom.strip('() ')))
+
+        return "ST_SetSRID('{}'::geometry, 4326)".format(out)
```
joto, line 38: Use EWKT format `SRID=4326;POLYGON((0 0,0 1,1 1,1 0,0 0))` instead of the `ST_SetSRID()`?

joto (commented):
I needed `tests/bdd/steps/__pycache__` in .gitignore. The `tests/__pycache__` can be removed I think.

The `test/bdd` directory contains the tests themselves (`flex`, `parameters`, `regression`) as well as the "infrastructure" (`environment.py`, `steps`) which I find slightly confusing. Also `parameters` is very general, maybe `command-line` or so?

There are several tests with a single input node that is just there so that *some* data is imported, it isn't checked what. Does it make sense to special case this somehow?

PR #1675, opened 2022-05-22
Call ANALYZE on tables with ids when two-stage processing is used

joto:
For these tables a (temporary) index is created but an ANALYZE might
help to speed up things.

### DIFF
--- a/src/flex-table.cpp
+++ b/src/flex-table.cpp
@@ -274,7 +274,7 @@ void table_connection_t::stop(bool updateable, bool append)
     }
 
     log_info("Analyzing table '{}'...", table().name());
-    analyze_table(*m_db_connection, table().schema(), table().name());
+    analyze();
 
     teardown();
 }
@@ -285,6 +285,11 @@ void table_connection_t::prepare()
     if (table().has_id_column() && table().has_geom_column()) {
         m_db_connection->exec(table().build_sql_prepare_get_wkb());
     }
+}
+
+void table_connection_t::analyze()
+{
+    analyze_table(*m_db_connection, table().schema(), table().name());
 }
 
 void table_connection_t::create_id_index()
--- a/src/flex-table.hpp
+++ b/src/flex-table.hpp
@@ -253,6 +253,8 @@ public:
 
     void prepare();
 
+    void analyze();
+
     void create_id_index();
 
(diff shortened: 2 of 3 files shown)


joto:
@mboeringa Maybe this helps with your problem. Can you test this?

mboeringa:
@joto , reviewing the changes, I see you have put the ANALYZE call before the actual index creation. Shouldn't this be the other way around? ANALYZE needs to be able to discover the index in order for the PostgreSQL planner to have use for it.

 ```
table.analyze();
table.create_id_index();
```

**Comment by joto** [comment]
@mboeringa No. ANALYZE collects statistics. Whether there is an index or not isn't relevant for that. The planner uses those stats and the information about indexes when planning, not somehow when ANALYZE is run.

I am running ANALYZE before the CREATE INDEX on the off chance that the index building can somehow be done in a better way with better statistics, although I don't believe that that's the case.

**Comment by joto** [comment]
Here is some info about ANALYZE and indexes: https://dba.stackexchange.com/questions/241257/is-it-necessary-to-analyze-a-table-after-an-index-has-been-created

**Comment by mboeringa** [comment]
> @mboeringa Maybe this helps with your problem. Can you test this?

@joto,

I'd like to, but I am having trouble updating. I used to simply do:

```
cd osm2pgsql
git pull origin master
```

followed by make, build and install to update my local copy, but this now fails at the 'git pull origin master' with a "connection timed out" and "unable to connect to github.com" type error. I do have a working internet connection from the VM where osm2pgsql and PostgreSQL are installed.

Has something changed recently in the way you need to update to latest master?

mboeringa:
Doing:

`git config --global url."https://".insteadOf git://`

solved the issue with connection.

mboeringa:
One other possible explanation for the difference in DELETE speed might be that PostgreSQL may be able to process the JOIN statement using multiple parallel workers. However, I would need to check that because I did not monitor that aspect.

mboeringa:
With a test dataset the size of Italy and OpenStreetMap buildings, I do not see PostgreSQL using parallel workers for the DELETE stage using the JOIN statement.

That doesn't fully exclude that PostgreSQL did uses parallel workers for the Planet size processing, but I think it unlikely at this moment. I would need to test it though.

PR #1683, opened 2022-06-07
Add get_bbox() test for relations

joto:


tests/bdd/flex/bbox.feature
```diff
@@ -71,10 +71,55 @@ Feature: Test get_bbox() function
...
+            """
+            w20 v1 dV Nn10,n11
+            w21 v1 dV Nn11,n12
+            r30 v1 dV Ttype=route,route=bus Mw20@@
+            r31 v1 dV Ttype=route,route=bus Mw20@,w21@
+            """
+        And the lua style
+            """
+            local rels = osm2pgsql.define_relation_table('osm2pgsql_test_routes', {
+                { column = 'tags',  type = 'hstore' },
```
lonvia, line 95: I already noticed when porting to BDD: do we need to keep the tags column? It's not relevant to the tests.

joto: I have removed all the "tags" columns from the bbox tags.

PR #1696, opened 2022-07-23
Improved debug information about memory use of Lua

joto:
Sample memory use every 1000 function calls into Lua and and the end of the program show current and max Lua memory use.

Not sure how useful this is in actual practice, but it is worth a try.

### DIFF
--- a/src/output-flex.cpp
+++ b/src/output-flex.cpp
@@ -1357,6 +1357,14 @@ void output_flex_t::add_row(table_connection_t *table_connection,
     }
 }
 
+void output_flex_t::remember_memory_used_by_lua() noexcept
+{
+    int const mem = lua_gc(lua_state(), LUA_GCCOUNT, 0);
+    if (mem > m_lua_memory_max) {
+        m_lua_memory_max = mem;
+    }
+}
+
 void output_flex_t::call_lua_function(prepared_lua_function_t func,
                                       osmium::OSMObject const &object)
 {
@@ -1375,6 +1383,11 @@ void output_flex_t::call_lua_function(prepared_lua_function_t func,
     }
 
     m_calling_context = calling_context::main;
+
+    if (++m_lua_memory_counter > 1000) {
+        m_lua_memory_counter = 0;
+        remember_memory_used_by_lua();
+    }
 }
 
 void output_flex_t::get_mutex_and_call_lua_function(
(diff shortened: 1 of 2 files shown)


pnorman:
I'll give it a try after my current import finishes

lonvia:
I don't suppose it is possible to print the current value as part of the progress report?

joto:
We could do that. It would be a bit of work because the information is specific to the output, but the printing of the progress report happens somewhere else.

Here is an alternative. Add this Lua function to your script:
```
function create_memory_reporter(filename, frequency)
    local counter = 0
    local file = io.open(filename, 'w')
    file:write('timestamp,counter,mbyte\n')

    return function()
        if counter % frequency == 0 then
            local mem = collectgarbage('count')
            local ts = os.date('%Y-%m-%dT%H:%M:%S,')
            file:write(ts .. counter .. ',' .. math.ceil(mem / 1024) .. '\n')
            file:flush()
        end
        counter = counter + 1
    end
end
```

Then call it like this from your process functions:
```
local mr = create_memory_reporter('/tmp/osm2pgsql-lua-memlog.csv', 10000)

function osm2pgsql.process_node(object)
    mr()
    ...
end
```
You can have one memory reporter for nodes, ways, and relations together or have separate ones.

This is really something where we can use the power of Lua to have something really configurable. So maybe it doesn't make sense to add this to osm2pgsql itself. (We can add the function above to our helper library though.)

mmd-osm:
In case you're concerned about runaway Lua scripts that keep on allocating more and more memory, another way at looking at the issue could be to enforce an upper memory limit for Lua scripts by using a custom allocator: https://stackoverflow.com/questions/9671793/limiting-a-lua-scripts-memory-usage (not tested, YMMV).

joto:
@mmd-osm The issue is more to get some insight into how much memory your script is using so that you can improve the script or at least know that's where the memory is going (in contrast to somewhere else like osmium buffers or whatever).

I have added the script above with some explanation [to the manual](https://osm2pgsql.org/doc/manual.html#logging-memory-use-of-lua-scripts).

mmd-osm:
Right. I think it doesn't really matter if you run this analysis inside your Lua code, or a custom allocator, both options are feasible.

Regarding controlling the maximum memory consumed by Lua (and thereby simulating out of memory conditions), I could trigger a segfault in `push_osm_object_to_lua_stack` -->  lua_pushstring didn't get any memory, and eventually called abort().

I'm not sure how relevant this is in real life. In case you want to take a look for yourself, I've uploaded the prototype code here: https://github.com/mmd-osm/osm2pgsql/tree/prototype/customalloc

Anyway, that's probably a topic for another issue...

joto:
Most of what's in this PR can be done from Lua, so I'll retract the PR. @pnorman and everybody else: Keep testing these things from Lua and once we have a better understanding what functionality is actually useful in practice, we can see whether we want to add something to osm2pgsql.

PR #1701, opened 2022-07-25
CMake config: Modernize setting of specific C++ standard

### LINKED ISSUE #1010: Declaring C++ version in CMake config
osm2pgsql currently compiles with C++11 or newer standards. For some libraries it might be better to compile in C++14 mode (or even newer) if it is available. This is the case for `fmt` which we recently introduced in osm2pgsql, which can use compile-time checking of format strings in C++14 but not C++11. I think the policy should be:

* if the user sets a version, use that one
* if your compiler has a default version >= C++11, use that one
* if your compiler has a default version < C++11, use C++11

Is that the right policy?

Currently the CMake config hard-codes the choice (with C++11 being the default and manual change possible). How can we make the policy above work in our CMake config?

joto:
We need C++17 currently. This changes the way we ask CMake for C++17 as a minimum version. We can still set CMAKE_CXX_STANDARD to force a newer standard for testing.

This needs CMake 3.8 which has been available for something like 5 years now.

Fixes #1010.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,6 @@ jobs:
       LUAJIT_OPTION: OFF
       POSTGRESQL_VERSION: 10
       POSTGIS_VERSION: 3
-      CPP_VERSION: 17
       BUILD_TYPE: Debug
 
     steps:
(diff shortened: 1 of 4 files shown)


joto:
See #1010 and #1704 for some problems with this. Closing until we can figure this out.

PR #1703, opened 2022-07-25
Implement ""_format operator ourselves instead of using from fmt

joto:
The one in fmt is deprecated in version 8.1 and removed in version 9.
This way we can keep the syntax and keep backwards compatibility
for a while.


joto:
See #1704 for some problems with this. Closing here for the time being.

PR #1717, opened 2022-07-29
Make sure we move geometry into geometry_t if possible

joto:


joto:
I thought I had fixed this problem, but apparently I have't and I can't seem to find a solution. So closing this for the moment.

PR #1721, opened 2022-07-30
way.is_closed() should only be called when there are way nodes

joto:
When you are feeding ways without nodes to osm2pgsql, which should never
happen in normal operation, osm2pgsql will break without this check.


lonvia:
What about dropping those ways (and possibly relations, too) in osmdata.cpp already?

joto:
> What about dropping those ways (and possibly relations, too) in osmdata.cpp already?

I decided against that because it might be useful for some kind of error checking application to actually find those cases.

PR #1736, opened 2022-08-07
Add geometry functions in Lua / new insert syntax

joto:
This is the future of geometry processing in Lua. Geometries are not something happening magically behind the scenes as before, but they are (mostly) normal objects that can be manipulated in Lua. There are several functions to create those geometry objects from the OSM objects:

* as_point() (for nodes)
* as_linestring() (for ways)
* as_polygon() (for ways)
* as_multilinestring() (for relations)
* as_multipolygon() (for relations)
* as_geometrycollection() (for relations)

Geometries can be manipulated with the following functions (all based on their equivalents in PostGIS):

* area() (for polygons)
* centroid() (for polygons)
* geometries() for iterating over (multi)geometries
* geometry_n() return the nth geometry from a (multi)geometry.
* geometry_type() returns type as string ('POINT', 'LINESTRING', ...)
* is_null() returns true if this is a null geometry
* num_geometries() returns the number of geometries (0 for null geom, 1 for point/linestring/polygon, n for multi* and geometrycollection geoms) (also available through the Lua # operator)
* simplify(tolerance) (for linestrings)
* srid()
* transform(target_srid)

Some of these functions are using boost::geometry for their implementation. More functions can be implemented in the future as needed.

There is a new syntax for inserting rows into a table. Instead of "add_row" the "insert" function is used. "add_row" is still available
and it works as before. But the new geometry types only work with the "insert" command. The "insert" function also allows multiple geometry columns which wasn't possible with "add_row".

See the file flex-config/new-insert-syntax.lua for an example that shows how to use the new capabilities.

### DIFF
--- a/src/flex-lua-geom.hpp
+++ b/src/flex-lua-geom.hpp
@@ -1,0 +1,42 @@
+#ifndef OSM2PGSQL_FLEX_LUA_GEOM_HPP
+#define OSM2PGSQL_FLEX_LUA_GEOM_HPP
+
+/**
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ * This file is part of osm2pgsql (https://osm2pgsql.org/).
+ *
+ * Copyright (C) 2006-2022 by the osm2pgsql developer community.
+ * For a full list of authors see the git log.
+ */
+
+/**
+ * \file
+ *
+ * Functions implementing the Lua interface for the geometry functions.
+ */
+
+#include "geom.hpp"
+
+struct lua_State;
+
+/**
+ * Create a null geometry object on the Lua stack and return a pointer to it.
+ */
+geom::geometry_t *create_lua_geometry_object(lua_State *lua_state);
+
+/**
+ * Get a geometry object from the Lua stack and return a pointer to it.
+ *
+ * \param lua_state The lua state.
+ * \param n The position in the stack, usually 1 for the first function
+ *          parameter.
+ */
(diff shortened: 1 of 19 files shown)


tests/bdd/flex/geometry-processing.feature
```diff
@@ -0,0 +1,140 @@
...
+                    type = g:geometry_type()
+                })
+            end
+
+            function osm2pgsql.process_relation(object)
+                local g = object:as_multipolygon()
+                tables.polygons:insert({
+                    name = object.tags.name,
+                    geom = g,
+                    type = g:geometry_type()
```
lonvia, line 91: type is unused. Intentional?

lonvia (commented):
Should you update the PR maybe remove the 'work in progress' from the commit message.

src/flex-lua-geom.hpp
```diff
@@ -0,0 +1,63 @@
...
+int geom_centroid();
+int geom_geometry_n();
+int geom_geometry_type();
+int geom_is_null();
+int geom_num_geometries();
+int geom_segmentize();
+int geom_simplify();
+int geom_srid();
+int geom_tostring();
+int geom_transform();
```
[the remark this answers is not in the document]
joto: Fixed in new version.

tests/bdd/flex/geometry-processing.feature
joto: Fixed in new version.

joto:
> Should you update the PR maybe remove the 'work in progress' from the commit message.

Fixed in new version.

PR #1738, opened 2022-08-07
Fix various small code smells in geometry code

joto:
See individual commits


src/geom-functions.hpp
```diff
@@ -170,12 +170,12 @@ geometry_t centroid(geometry_t const &geom);
  *
  * Returns null geometry for anything but linestring geometries!
  *
- * \param output Pointer to output geometry.
+ * \param input Pointer to output geometry.
  * \param geom Input geometry.
```
lonvia, line 174: You changed the wrong parameter name here.

joto: damn... fixed in new version of this PR

PR #1743, opened 2022-08-10
Some more work in the geometries

joto:
See individual commits.

### DIFF
--- a/src/geom-from-osm.cpp
+++ b/src/geom-from-osm.cpp
@@ -96,27 +96,37 @@ geometry_t create_polygon(osmium::Way const &way)
 }
 
 void create_multilinestring(geometry_t *geom,
-                            osmium::memory::Buffer const &ways)
+                            osmium::memory::Buffer const &ways_buffer,
+                            bool force_multi)
 {
-    auto &mls = geom->set<multilinestring_t>();
-
-    for (auto const &way : ways.select<osmium::Way>()) {
-        linestring_t line;
+    auto ways = ways_buffer.select<osmium::Way>();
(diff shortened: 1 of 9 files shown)


src/geom-from-osm.cpp
```diff
@@ -96,27 +96,37 @@ geometry_t create_polygon(osmium::Way const &way)
...
+        }
+    } else {
+        auto &multiline = geom->set<multilinestring_t>();
+        for (auto const &way : ways) {
+            linestring_t line;
+            fill_point_list(&line, way.nodes());
+            if (line.size() >= 2U) {
+                multiline.add_geometry(std::move(line));
+            }
+        }
```
lonvia, line 118: There is a corner case here when there are multiple ways in the buffer but only one of them is valid. Then still a multilinestring with a single member is created. It's probably okay, just something we should keep in mind and potentially document.

tests/test-geom-multilinestrings.cpp
```diff
@@ -36,6 +36,25 @@ TEST_CASE("create_multilinestring with single line", "[NoDB]")
     REQUIRE(ml[0] == expected);
 }
 
+TEST_CASE("create_multilinestring with single line an no force_multi", "[NoDB]")
```
lonvia, line 39: an -> and

joto: Fixed in new version.

src/geom-from-osm.cpp
joto: I put it on my todo list.

PR #1744, opened 2022-08-10
Update/add some flex config example files using the new insert()

joto:


flex-config/geometries-using-insert.lua
```diff
@@ -0,0 +1,202 @@
...
+
+function osm2pgsql.process_way(object)
+    if clean_tags(object.tags) then
+--        return
+    end
+
+    -- A closed way that also has the right tags for an area is a polygon.
+    if object.is_closed and has_area_tags(object.tags) then
+        -- Creating the polygon geometry takes time, so we do it once here
+        -- and later store it in the table and use it to calclate the area.
```
lonvia, line 134: typo: calculate

flex-config/generic.lua
```diff
@@ -255,25 +257,25 @@ function osm2pgsql.process_relation(object)
...
         })
         return
     end
 
     if relation_type == 'boundary' or (relation_type == 'multipolygon' and object.tags.boundary) then
-        tables.boundaries:add_row({
+        tables.boundaries:insert({
             tags = object.tags,
-            geom = { create = 'line' }
+            geom = object:as_multilinestring()
```
lonvia, line 270: Shouldn't there be a line_merge to have the same result as before?

flex-config/geometries-using-insert.lua
```diff
@@ -0,0 +1,202 @@
...
+
+    -- Store boundary relations as multilinestrings
+    if relation_type == 'boundary' then
+        tables.boundaries:insert({
+            type = object:grab_tag('boundary'),
+            tags = object.tags,
+            -- For relations there is no clear definition what their geometry
+            -- is, so you have to decide on the geometry transformation you
+            -- want. In this case we want the boundary as multilinestring
+            -- and we want lines merged as much as possibl.e
```
lonvia, line 181: typo final dot

flex-config/generic.lua
joto: fixed in new version

flex-config/geometries-using-insert.lua
joto: fixed in new version

flex-config/geometries-using-insert.lua
joto: fixed in new version

PR #1753, opened 2022-09-06
Improvements to osm2pgsql-replication script

lonvia:
This PR collects some smaller improvements to the osm2pgsql-replication script to harden it against potential security issues:

* Use psycopg2's SQL module for correct quoting of PostgreSQL identifiers like table names. This includes fixing one spot where there was a SQL code injection possible via the `--prefix` parameter. Note that the threat level for the injection is low because the attacker would still need credentials for the database.
* When installing the script with `make install`, the default osm2pgsql binary will no longer be searched via PATH. Instead a hard-coded path to the installed osm2pgsql is used.
* Document that the script accepts connection parameters via libpq environment parameters and point to pgpass files for handing in passwords. Also fixes the script to work when no database name is given in the command line.

In addition there is now a more meaningful error returned when the prefix parameter is wrong or the middle tables do not exist.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -332,8 +332,11 @@ add_subdirectory(docs)
 # Install
 #############################################################
 
+include(GNUInstallDirs)
+
 if (ENABLE_INSTALL)
     install(TARGETS osm2pgsql DESTINATION bin)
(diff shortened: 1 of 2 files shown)


joto (approved):
Looks okay to me.

PR #1758, opened 2022-09-08
Check more user-supplied strings used as SQL identifier

lonvia:
We already consistently quote all identifier used in SQL strings in the code to avoid SQL injection issues. The quoting works on the assumption that there are no double quotes in the string itself. The flex output already checked all user-supplied strings for conformance.

This PR adds the same check to the prefix and schema command-line parameters which are used as SQL identifiers and for the column names and types from the pgsql style file.

### DIFF
--- a/src/pgsql.cpp
+++ b/src/pgsql.cpp
@@ -206,8 +206,7 @@ void check_identifier(std::string const &name, char const *in)
     }
 
     throw std::runtime_error{
-        "Special characters are not allowed in {} names: '{}'."_format(in,
-                                                                      name)};
+        "Special characters are not allowed in {}: '{}'."_format(in, name)};
 }
 
 std::map<std::string, std::string>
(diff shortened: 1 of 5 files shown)


src/pgsql.cpp
```diff
@@ -197,6 +197,19 @@ std::string qualified_name(std::string const &schema, std::string const &name)
...
+{
+    auto const pos = name.find_first_of("\"',.;$%&/()<>{}=?^*#");
+
+    if (pos == std::string::npos) {
+        return;
+    }
+
+    throw std::runtime_error{
+        "Special characters are not allowed in {} names: '{}'."_format(in,
+                                                                      name)};
```
joto, line 210: The message can now be confusing, for instance for the `--prefix` option it will say:
`Special characters are not allowed in prefix names: '...'.` . Maybe just remove the `names`?

joto (approved):
Gute Lösung

PR #1767, opened 2022-09-18
Fix: Create point geometry from collection with single-node way

joto:
If a geometry collection is created from a relation, any way member will normally end up as linestring relation. But if there is only a single point in the resulting linestring (because the way only contained a single node or multiple nodes with the same position), that linestring would be invalid. In this case create a point geometry instead.

### DIFF
--- a/src/geom-from-osm.cpp
+++ b/src/geom-from-osm.cpp
@@ -202,6 +202,9 @@ void create_collection(geometry_t *geom,
             if (line.size() >= 2U) {
                 collection.add_geometry(std::move(item));
             }
+            if (line.size() == 1) {
+                collection.add_geometry(geometry_t{std::move(line[0])});
+            }
         }
     }
 
--- a/tests/bdd/flex/geometry-collection.feature
+++ b/tests/bdd/flex/geometry-collection.feature
@@ -54,7 +54,7 @@ Feature: Create geometry collections from relations
             | 10 |
         And the OSM data
             """
-            w20 Nn10
+            w20 Nn11
             r30 Tname=foo Mn11@
             r31 Tname=bar Mw20@
             r32 Tname=baz Mw21@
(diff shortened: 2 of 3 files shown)


mboeringa:
@joto,

I am just wondering: is this actually desired behavior?

E.g., lets assume I have code like the one I posted here:

https://github.com/openstreetmap/osm2pgsql/discussions/1752#discussioncomment-3579004

where I use new 'flex' syntax to specifically select "w" or way role members from a relation. Now lets assume there is a bad way in there having only a single node. In that case, despite attempting to specifically select only ways, with this change, I will unexpectedly end up with a osm2pgsql generating a point geometry instead. 

Doesn't this break the:

> The GeometryCollection we spoke about contains one geometry for each node and way member, so you can iterate over both the object.members array as well as the result of object:as_geometrycollection() at the same time, ignore the members of type relation and you have a one-to-one correspondence between roles and geometry.

remark you wrote in the https://github.com/openstreetmap/osm2pgsql/discussions/1752#discussioncomment-3555849 issue remark, where the "one-to-one" correspondence no longer holds?

Wouldn't it be better to drop such a geometry as "invalid", and not return it as part of the geometry collection?

joto:
@mboeringa The problem is that if you drop that geometry, the geometry collection will contain one fewer geometry than you expected and you can't detect that. If you want to use a specific geometry of one of the members you might get the wrong one! So this is a fix for exactly that problem.

We are already thinking about a better way to access geometries of one member only. Its great that this is possible now, but it is too cumbersome. This will need some time though.

mboeringa:
> The problem is that if you drop that geometry, the geometry collection will contain one fewer geometry than you expected and you can't detect that. If you want to use a specific geometry of one of the members you might get the wrong one! So this is a fix for exactly that problem.

Thanks @joto for the response, but I am still slightly struggle to wrap my head around this.

OK, so this fixes the issue of the discrepancy that in the number of geometries in the geometry collection, and the number of members in the object.members array.

But does this fix then also ensure the 'type' returned by:
```
local geom = object:as_geometrycollection()
        local object_members = object.members

        for i = 1,geom:num_geometries(),1 do

            n_geom = geom:geometry_n(i)
            n_member_table = object_members[i]
            n_member_type = n_member_table.type
```

is auto-magically changed from 'w' to 'n' signify what happened in the **object:as_geometrycollection()** call?

Or am I simply completely misunderstanding this?

joto:
@mboeringa No. It doesn't have anything to do with the `type` in `object.members`. That's still 'w', because it is still a way that's the member.

If you want to make sure you got a LineString and not a Point, you have to check with `geometry_type()`.

lonvia:
So it turns out that this is not the only case where a geometry might be missing in the collection. A way or its geometry information might also be missing because the database was created from an extract that doesn't have the way in question. If we can't avoid missing members, then it would imo better to have the same behaviour for invalid and missing ways: simply drop the member from the collection.

The consequence would unfortunately be that @mboeringa's use-case doesn't work anymore with the `as_geometrycollection()`. Or rather: it only works when the relation is complete. The good news is that it is possible to detect the situation (mismatch between number of geometries and number of node and way members), so problematic relations can be dropped. It is just not possible to fix the error. We need a different mechanism here to collect members.

joto:
@lonvia I agree, we need to revert that. I am already working on a mechanism to get single member geometries out of relations reliably, but it will need a bit of time.

PR #1769, opened 2022-09-20
Allow multiple source in osm2pgsql-replication

pinaraf:
Instead of assuming the replication_status table can contain only one source, make it possible to add multiple sources, and iterate on the table when running update.
When replicating for instance Europe from geofabrik, if you want the whole french territories, this is a must-have : the europe.osm.pbf file does not include french territories like Réunion, Guyane or Guadeloupe. And likewise, their replication stream will not include these territories.

### DIFF
--- a/scripts/osm2pgsql-replication
+++ b/scripts/osm2pgsql-replication
@@ -212,42 +212,44 @@ def status(conn, args):
 
     if not table_exists(conn, args.table_name):
         results['status'] = 1
-        results['error'] = "Cannot find replication status table. Run 'osm2pgsql-replication init' first."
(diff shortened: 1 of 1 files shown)


pinaraf:
I've tested this successfully, so I'm marking it ready for review, but there is one pending question in the change: should I add an osm2pgsql-replication drop command, that would remove one source or all sources of replication?

joto:
If you have data from adjoining areas there will be some overlap. There can also be overlap from relation members even if the areas aren't that close. And if you get different versions of the same object from the different input files, all sorts of problems can result.

Issue #1757, opened 2022-09-08
Add `length()` Lua function for geometries


joto:
It would be great to have a `length()` function for (multi)linestrings. Implementation similar to the `area()` function for polygons using the boost::geometry function of the same name.

See #1756

pnorman:
Should this be the total length, or, when splitting is on, the length of the split linestring?

joto:
> Should this be the total length, or, when splitting is on, the length of the split linestring?

This would be a function on geometry objects which don't do magic auto-splitting. It should just do what it says on the box, i.e. calculate the length of that geometry.

victoriasiver26:
Hello, I am interested in contributing. This would be my first contribution so I might need some extra guidence, would that be alright?

joto:
@victoriasiver26 Sure. Try installing osm2pgsql from source and running it using the flex output. Try the `geometries-using-insert.lua` example from the `flex-config` directory. Look there for use of the `area()` function. Then find the places in the source where the `area()` function is implemented and do the equivalent for the `length()` function.

lonvia:
Implemented in #1777.

PR #1777, opened 2022-09-24
length for flex output

iboates:
I have implemented this.

Related issue: #1757

The CI is still running, if it passes then it should be ready to merge pending manual inspection

### DIFF
--- a/src/flex-lua-geom.cpp
+++ b/src/flex-lua-geom.cpp
@@ -69,6 +69,18 @@ static int geom_area(lua_State *lua_state)
     return 1;
 }
 
+static int geom_length(lua_State *lua_state)
+{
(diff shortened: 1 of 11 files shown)


iboates:
Some re-ordering of include statements & other changes happened because of a screw-up with clang-format

I didn't understand how it was supposed to work, and made my commits before running `git-clang-format`. I thought I could just run `clang-format -i`, but that modified every file in the repo with garbage changes. So I backed up the repo manually and did a hard reset, then copied the files that I actually meant to edit back into the repo and did `git-clang-format` before committing, like I should have in the first place..

Because of this, there were some lingering changes from `clang-format -i` that snuck into the PR. Sorry about that.

joto (approved):
Looks good. Thank you!

PR #1782, opened 2022-09-29
Add implementation of simplify function for multilinestrings

joto:
We already have an implementation for linestring, which is slightly improved in this commit. Simplification for anything other than (multi)linestring is not supported.

In the first commit there is also a refactor of the `fill_point_list()` function which makes the code more readable and consistent.

### DIFF
--- a/src/geom.hpp
+++ b/src/geom.hpp
@@ -90,7 +90,15 @@ private:
 
 }; // class point_t
 
-/// This type is used as the basis for linestrings and rings.
+/**
+ * This type is used as the basis for linestrings and rings.
+ *
+ * Point lists should not contain consecutive duplicate points. You can
+ * use the remove_duplicates() function to remove them if needed. (OGC validity
+ * only requires there to be at least two different points in a linestring, but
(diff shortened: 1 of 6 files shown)


src/geom.hpp
```diff
@@ -117,6 +120,19 @@ class linestring_t : public point_list_t
...
+     * Check validity of linestring.
+     *
+     * By the OGC definition a linestring must have at least two distinct
+     * nodes to be valid. We are a bit more strict and require that the
+     * first two nodes are also distinct. When we create linestrings from
+     * OSM data this will always be the case.
+     */
+    bool is_valid() const noexcept
+    {
+        return size() > 1 && (*this)[0] != (*this)[1];
```
lonvia, line 133: Is the second check still needed now that you are explicitly removing duplicates?

joto: Not in the case we have so far, but I would expect that this function will be useful for other cases where we didn't explicitly remove duplicates, so I'd rather keep that.

joto: I have remove the is_valid() function now and do the check without it. This should be less confusing.

PR #1792, opened 2022-10-03
Add note about use of timestamp/timestamptz to example

### LINKED ISSUE #1785: Update example config to appropriate timestamp types
While reviewing #1784 I noticed that the attributes.lua example uses `sql_type = 'timestamp'`.

`timestamp` (without time zone) [should not be used to store UTC times](https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp_.28without_time_zone.29_to_store_UTC_times). Instead, it should be `timestamptz`

joto:
Closes #1785


lonvia:
Sorry, I missed the review request. I can revert if necessary.

PR #1793, opened 2022-10-05
Fix osm2pgsql-replication installation when DESTDIR is used.

sebastic:
The Debian package build for 1.7.1 failed because it uses [`DESTDIR`](https://cmake.org/cmake/help/latest/envvar/DESTDIR.html) which was ignored when installing the `osm2pgsql-replication` template.

Ref: #1753


lonvia:
This will not do what you want because the installed script still contains the configured `CMAKE_INSTALL_PREFIX` and not the directory configured with `DESTDIR`.

lonvia:
Oh, I see. That is in fact what you want.

I need to think about how to do that.

sebastic:
Why is `INSTALL_PREFIX` set with a CMake variable in the first place, instead of just using `os.path.dirname(os.path.abspath(__file__)))`?

lonvia:
Because I wanted the script to still work like before when not installed. But that's obviously too ambitious for cmake. Let's go for the simple 'same directory' solution then.

lonvia:
Closing in favour of #1799.

PR #1799, opened 2022-10-22
always expect osm2pgsql in same path as osm2pgsql-replication

lonvia:
Changes the default search location of osm2pgsql to be in the same path as wherever osm2pgsql-replication is, as suggested by @sebastic in #1793. Anything else just confuses cmake.


sebastic:
Why `TYPE BIN` instead of `DESTINATION bin` like `osm2pgsql`?

joto:
`TYPE BIN` is not available in CMake 3.5 (currently set as minimum in the config) or 3.13 (used in Debian 10 Buster).

lonvia:
Yeah, didn't see it was a rather recent feature. Changed back.

PR #1802, opened 2022-10-29
Various refactorings of expire code

joto:
To prepare for coming changes.

### DIFF
--- a/src/output-pgsql.cpp
+++ b/src/output-pgsql.cpp
@@ -325,8 +325,21 @@ void output_pgsql_t::relation_add(osmium::Relation const &rel)
  * contain the change for that also. */
 void output_pgsql_t::node_delete(osmid_t osm_id)
 {
-    if (m_expire.from_result(m_tables[t_point]->get_wkb(osm_id)) != 0) {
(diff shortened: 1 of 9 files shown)


src/output-pgsql.cpp
```diff
@@ -325,11 +325,24 @@ void output_pgsql_t::relation_add(osmium::Relation const &rel)
...
 void output_pgsql_t::node_delete(osmid_t osm_id)
 {
-    if (m_expire.from_result(m_tables[t_point]->get_wkb(osm_id)) != 0) {
+    auto const results = m_tables[t_point]->get_wkb(osm_id);
+    if (expire_from_result(&m_expire, results) != 0) {
         m_tables[t_point]->delete_row(osm_id);
     }
 }
 
+void output_pgsql_t::expire(osmid_t id)
```
lonvia, line 334: This function also deletes rows in the tables. Maybe reflect that in the name?

joto: I have renamed the function to `delete_from_output_and_expire` and force-pushed a rebased version of those commits.

PR #1808, opened 2022-11-02
Use min/max in expire and remove superfluous check

joto:


src/expire-tiles.cpp
```diff
@@ -200,26 +200,18 @@ int expire_tiles::from_bbox(geom::box_t const &box)
     }
 
     /* Convert the box's Mercator coordinates into tile coordinates */
-    auto const tmp_min = coords_to_tile({box.min_x(), box.max_y()});
-    int min_tile_x = tmp_min.x() - tile_expiry_leeway;
-    int min_tile_y = tmp_min.y() - tile_expiry_leeway;
+    auto const tmp_min = coords_to_tile(box.min());
```
lonvia, line 203: This code is not the same. box.min() returns `min_y` while the old code has `max_y`. Is this a problem?

joto: Yes it is! Now that we have tests (from another recently added PR) we can immediately see that they fail!

PR #1809, opened 2022-11-02
Refactor expire by geometry code

joto:
Split up code into functions for each geometry type. Adds support for geometry types that have been missing so far (multipoint, geometry collection).


src/expire-tiles.cpp
```diff
@@ -78,41 +78,60 @@ void expire_tiles::from_point_list(geom::point_list_t const &list)
...
+        from_polygon_boundary(geom);
+    }
+}
+
+void expire_tiles::from_geometry(geom::multipolygon_t const &geom)
+{
+    geom::box_t box;
+    for (auto const &sgeom : geom) {
+        box.extend(sgeom.outer());
+    }
```
lonvia, line 114: Isn't that the same as `box = geom::envelope(geom)`?

lonvia (commented):
There is one more call to `from_geometry()` here: https://github.com/openstreetmap/osm2pgsql/blob/bd8d41fcdf6814404132795f0e571c7357cd22ad/src/expire-tiles.cpp#L291 That could be non-3857, too, or am I reading that wrong?

src/expire-tiles.cpp
joto: Yes, I missed that one. Changed.

joto:
I have changed the code some more based on the comment above regarding the use of `from_geometry()` vs. `from_geometry_if_3857()`. In that case the check should be done before we even reach that code, otherwise we are doing too much work (requesting the geometry from the database and then ignoring the result).

I added a check in the command line options parsing code to detect the case where expire is enabled but a projection other than Mercator used and disable the expire then. That makes checking for that condition easier later on and the user gets a nice warning.

PR #1815, opened 2022-11-04
Allow tables in flex output that only have ids columns

joto:
Might be useful for a table that works as a kind of todo list, containing only the ids of features to work on.

You still need either an id column or some other columns, table without columns are not allowed. :-)


lonvia:
I would prefer to have an explicit notation `columns={}` for these empty tables. If you allow a `nil` value, then osm2pgsql will no longer detect spelling errors in your table definition and simply create an empty table instead.

Issue #1764, opened 2022-09-16
Make simplify work with geoms other than linestrings


joto:
The `simplify()` function currently only works for linestrings.

* It should also work for multilinestrings.
* It should also work for (multi)polygons.

The problem with simplify is

a) there are many different algorithms which have different properties (I believe only Douglas-Peucker is implemented in boost::geometry which we use)
b) simplification can make geometries invalid, which is bad and we usually guarantee that we only import valid geometries. This is especially bad for polygons. 

See #1663.

joto:
See also https://github.com/kleunen/boost_geometry_simplify

PR #1830, opened 2022-11-21
WIP: Add code to generalize data using different strategies

joto:
This commit adds the code to generalize polygons using two different strategies. In both cases polygons are merged and simplified on a tile-by-tile basis.

The "vector" strategy uses the commonly used approach where polygons are buffered, their union is calculated and then a reverse buffer applied, followed by a final buffer again.

The "raster" strategy does a similar thing but does it in raster space which is much faster. First the polygons are rendered into a raster, an open/close operation is called (which basically does the same thing as the buffering in vector space) and finally the resulting raster is vectorized again.

For the raster support this adds two new library dependency: CImg and potrace.

The functionality is accessed through a new command line program called osm2pgsql-gen. Call it with -h to get some usage information. This program is for testing only, eventually the functionality should be accessible from osm2pgsql itself (using the Lua config file for configuration).

Some additional information is in README-gen.md

### DIFF
--- a/.github/actions/ubuntu-prerequisites/action.yml
+++ b/.github/actions/ubuntu-prerequisites/action.yml
@@ -16,10 +16,12 @@ runs:
     - name: Install software
       run: |
         sudo apt-get install -yq --no-install-suggests --no-install-recommends \
+          cimg-dev \
           libboost-filesystem-dev \
           libboost-system-dev \
           libbz2-dev \
           libexpat1-dev \
+          libpotrace-dev \
           libpq-dev \
           libproj-dev \
           pandoc \
--- a/.github/actions/win-install/action.yml
+++ b/.github/actions/win-install/action.yml
@@ -5,9 +5,8 @@ runs:
 
   steps:
     - name: Install packages
-      run: vcpkg install bzip2:x64-windows expat:x64-windows zlib:x64-windows proj4:x64-windows boost-geometry:x64-windows boost-system:x64-windows boost-filesystem:x64-windows boost-property-tree:x64-windows lua:x64-windows libpq:x64-windows
+      run: vcpkg install cimg:x64-windows bzip2:x64-windows expat:x64-windows zlib:x64-windows proj4:x64-windows boost-geometry:x64-windows boost-system:x64-windows boost-filesystem:x64-windows boost-property-tree:x64-windows lua:x64-windows libpq:x64-windows
       shell: bash
-
     - name: Install psycopg2 and beahve
       run: python -m pip install psycopg2 behave
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
 
       - name: Install prerequisites
         run: |
-          brew install lua boost postgis pandoc
+          brew install lua boost postgis pandoc cimg potrace
           pip3 install psycopg2 behave
           pg_ctl -D /usr/local/var/postgres init
           pg_ctl -D /usr/local/var/postgres start
@@ -45,6 +45,7 @@ jobs:
     env:
       CC: gcc-10
       CXX: g++-10
+      EXTRA_FLAGS: -Wno-unused-but-set-parameter # workaround for GCC bug
       LUA_VERSION: 5.3
       LUAJIT_OPTION: ON
       POSTGRESQL_VERSION: 9.6
@@ -79,6 +80,7 @@ jobs:
     env:
       CC: gcc-10
       CXX: g++-10
+      EXTRA_FLAGS: -Wno-unused-but-set-parameter # workaround for GCC bug
       LUA_VERSION: 5.3
       LUAJIT_OPTION: OFF
       POSTGRESQL_VERSION: 10
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -37,12 +37,14 @@ jobs:
         run: |
           sudo apt-get purge -yq postgresql*
           sudo apt-get install -yq --no-install-suggests --no-install-recommends \
+            cimg-dev \
             libboost-filesystem-dev \
             libboost-system-dev \
             libbz2-dev \
             libexpat1-dev \
             liblua${LUA_VERSION}-dev \
             libluajit-5.1-dev \
+            libpotrace-dev \
             libpq-dev \
             libproj-dev \
             lua${LUA_VERSION} \
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -205,6 +205,11 @@ include_directories(SYSTEM ${PostgreSQL_INCLUDE_DIRS})
 
 find_package(Threads)
 
+find_path(POTRACE_INCLUDE_DIR potracelib.h)
+find_library(POTRACE_LIBRARY NAMES potrace)
+
+find_path(CIMG_INCLUDE_DIR CImg.h)
+
 ############### Libraries are found now ########################
 
 set(LIBS ${Boost_LIBRARIES} ${PostgreSQL_LIBRARY} ${OSMIUM_LIBRARIES})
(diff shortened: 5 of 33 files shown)


pnorman:
Is the main goal here to verify that the algorithms produce useful results before implementing in osm2pgsql?

mboeringa:
@joto 

Interesting reads those blog posts.

A couple of remarks though:

- Unless you use **ST_Buffer**'s 'num_seg_quarter_circle' setting in the vector option, you may in fact end up with far more vertices in the output than in the input polygons, and not be generalizing the data at all... Using 'num_seg_quarter_circle' with a low value like 2 or 3, should considerably speed up the (subsequent) processing as well.
- It may be highly recommended to add an additional true generalization step using 'ST_Simplify/ST_SimplifyPreserveTopology/ST_SimplifyVW' after the buffer operation, for the same reason as listed above: truly reduce the number of vertices and processing requirements.
- There is a high chance of producing invalid output with Polygon generalization. In fact, when I ran the QGIS's 'Check Validity' option two years ago or so against the generalized land and water polygons as created by the coastline processing you refer to and as available on https://osmdata.openstreetmap.de/data/, QGIS detected multiple errors. I don't know the current situation, I should recheck it, but from my own experience developing my own algorithms, it is easy to run into trouble with invalid Polygons (and it can be really hard to fix it and get 100% valid output). In my particular case, running 'ST_MakeValid' was not enough to deal with all issues (but ArcGIS is even more finicky in this respect due to the different Polygon validity framework, although any OGC valid one is also valid in the ESRI world according to https://pro.arcgis.com/en/pro-app/latest/tool-reference/data-management/check-geometry.htm).
- Anyway, since the goal of osm2pgsql was to always produce valid output, I would certainly recommend carefully reviewing any generalization results from the developed algorithms for issues, as invalid geometries can cause a range of issues for subsequent use.

mboeringa:
A few more remarks:
- I also recommend testing any generalization algorithm at a global scale. Small regional extracts may not reveal all issues, especially related to the anti-meridian, and how some PostGIS commands deal with that.
- Mind the "fine print" with any of the 'ST_Simplify/ST_SimplifyPreserveTopology/ST_SimplifyVW' PostGIS commands as listed in the PostGIS documentation.
- Output from the current, non-generalized, osm2pgsql processing pipeline, has shown to be very robust in both QGIS and ArcGIS for me, with fully valid readable output, even with the more advanced 'flex' processing.

joto:
You are raising some valid points, @mboeringa. In fact the first version of the code in this PR would generate polygons that were all invalid! (The rings were not closed, which is fixed now.) I also added an optional  `ST_MakeValid()` step to the `raster-union` strategy which seems to work well in my experiments (use `-p make_valid` to enable) and seems to not take too much time. But overall I am not too worried about validity. The generalized data is not that accurate anyway, that's the whole point of generalization, and small rendering errors will probably not matter. Still, it is better to have things valid. We'll need some more real-world experience with this to see how important that is and whether it is worth the extra effort.

Unfortunately the potrace library will sometimes generate invalid polygons. I have seen two cases: Sometimes it snips off corners of the tiles it renders, this can be mitigated by having larger margins around tiles, but that's certainly something to look out for. The other happens when rendering polygons touching in single points (think of the figure "8") which must be modelled as two polygons, not one. There is another problem with the potrace library which I couldn't figure out yet: The left side of all polygons is slightly to the left of where it should be, the right side is okay. Maybe some kind of rounding error. You can't usually see it in the results though, because the effect is small, but still somewhat annoying.

You also mentioned the `num_seg_quarter_circle` setting and the need for simplification. I have played around with this in a different context, but didn't do that in this project. Using fewer segments for the buffer makes processing noticably faster, so that is an important optimization, but I am focusing on the new raster-based approach which is even faster and it generates far smaller polygons (with less vertexes) to begin with, so I believe there is no need for the extra simplification step. But it would be interesting to try this out. There is a huge experimental space here and I fully expect that we'll tweak code and parameters for a long time to come to better results. Anybody who wants to play around with this, please go ahead and report your findings here.

pnorman:
> The generalized data is not that accurate anyway, that's the whole point of generalization, and small rendering errors will probably not matter. Still, it is better to have things valid

Having a guarantee of validity is very important since lots of PostGIS functions require validity.

README-gen.md
```diff
@@ -0,0 +1,75 @@
...
+  supported are `raster-union` and `vector-union`.
+* `-m, --margin=MARGIN`: This sets the margin around the tile as a fraction of
+  the tile size. So a value of 0.1 sets a 10% boundary on each side of the tile,
+  so as a result the tiles will overlap by 20% of their size. When using the
+  `raster-union` strategy the margin will be rounded up to the nearest multiple
+  of 64 pixels.
+* `-w, --width=WIDTH`: Size of the imaged rendered when using the `raster-union`
+  strategy, not used in the `vector-union` strategy.
+* `-b, --buffer=BUFFER`: The amount by which the polygons will be buffered. For
+  the `raster-union` strategy this is in pixels, for the `vector-union` strategy
```
pnorman, line 56: The documentation doesn't say what resolution is used per tile

src/osm2pgsql-gen.cpp
```diff
@@ -0,0 +1,365 @@
...
+     {nullptr, 0, nullptr, 0}}};
+
+static std::array<uint32_t, 4> get_extent_from_db(pg_conn_t *connection,
+                                                  std::string const &table,
+                                                  uint32_t zoom)
+{
+    connection->query(
+        PGRES_COMMAND_OK,
+        "PREPARE get_extent AS WITH ex AS "
+        "(SELECT ST_EstimatedExtent('{}', 'geom') AS e) "
```
pnorman, line 99: This hard-codes the geometry column to `geom`

src/osm2pgsql-gen.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+        if (params.count("expire_list")) {
+            log_info("Get tiles to reprocess from table '{}'",
+                     params["expire_list"]);
+            x = 0;
+            y = 0;
+            xmax = 0;
+            ymax = 0;
+        } else {
+            if (x == 0 && y == 0 && xmax == 0 && ymax == 0) {
```
pnorman, line 285: Zero shouldn't be treated as a special value, because you might just want to generate the 0/0 tile.

README-gen.md
```diff
@@ -0,0 +1,75 @@
...
+  this is in Mercator map units.
+* `-g, --group-by-column=COL`: Set this to the column describing the type of
```
pnorman, line 58: How are multiple columns indicated?

pnorman (changes requested):
A few things I noticed while getting it running. Not a complete review.

- There's a bunch of hard-coded 3857 assumptions. If we're okay with these, we need to document them.

src/osm2pgsql-gen.cpp
joto: fixes in new version

src/osm2pgsql-gen.cpp
joto: fixed in new version

README-gen.md
joto: That's not possible. If that's really something we need, we can add that later. But I am trying to keep the config to a minimum.

joto:
New version documents that this only works for 3857. Also added a "make_valid" option to ensure results are valid polygons. Lots of bug fixes. And other stuff... I'll get around to documenting this eventually...

README-gen.md
```diff
@@ -0,0 +1,79 @@
...
+needed.
+
+## Usage
+
+Run osm2pgsql as usual. This example assumes you have a table called
+`landcover` with a polygon geometry column called `geom` and a text column
+called `type` with the type of the landcover (forest, grass, etc.).
+
+Create a table `landcover_z10` with (at least) columns `geom` and `type` as
+well as integer columns `x`, and `y`. Then call this:
```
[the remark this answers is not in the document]
joto: If you use `osm2pgsql-gen` on larger extracts or the planet you need to run it with `-D` on first import to suppress `DELETE`s, then built indexes on `x, y` and when you do updates run without the `-D`. The problem is that by default `osm2pgsql-gen` will `DELETE` all features in a tile in the output table before loading them. But when you start that for the first time, i.e. when the `PREPARE` is done, the table is empty, so even if there is an index PostgreSQL will decide not to use it. After the import you create the index and do an `ANALYZE` and then everything is fine.

This is a bit tricky, but once we put this functionality into osm2pgsql itself, we know whether we are in import or append mode and can do the right thing automatically.

pnorman:
I've been experimenting with the results now, and am having an error where the lower-left corner of each tile's polygon is being clipped away

I created the base table with
```sql
CREATE MATERIALIZED VIEW
osm_planet_trees
AS
SELECT
    osm_id,
    way,
    name,
    ref,
    way_area
FROM planet_osm_polygon
WHERE "natural" = 'wood' OR landuse = 'forest'
```
then ran
``PGDATABASE=flex time ~/osm/osm2pgsql/build/osm2pgsql-gen -t planet_osm_trees -T
trees_z10 -g way -G name -s raster-union -x0 -X1024 -y0 -Y1024 -z10 -m 0.125``

The result in qgis looks like this:

image

joto:
> I've been experimenting with the results now, and am having an error where the lower-left corner of each tile's polygon is being clipped away

I believe that's already been fixed in the current version of this PR. Did you use that?

danieldegroot2:
This PR seems to be part of #1764.

joto:
Closing in favour of #1942.

PR #1836, opened 2022-11-26
Implement and use our own JSON writer

joto:
This isn't much work and we can get rid of rapidjson for which there hasn't been a release in many years and which shows problems in clang-tidy and with UB sanitizer.


pnorman:
Is boost/json an option for serializing json?

joto:
@pnorman No, for two reasons:

a) It only appeared in Boost 1.7.5 from December 2020, so it will not be, for instance, available in Ubuntu 20.04.
b) As far as I can tell you first have to create a DOM-like tree of the data before you can serialize it. This adds a lot of overhead.

I looked at other JSON libraries, too. We don't want to use most of them because of (variants of) reason (a) or (b).

And the code in this PR is really small. It took me less than an hour to write and most of that was spent on figuring out how to serialize some special characters. At that point I asked myself whether it is really worth it to have an external dependency. I know, not invented-here-syndrome and all that...

mmd-osm:
Adding this logic to libosmium would allow us to more easily add osm json format in the future, in case there’s a need for this format in the first place.

PR #1838, opened 2022-11-27
Remove copy_thread parameter from output constructors

joto:
The output derived classes can create the copy_thread themselves. They just need the options which they have anyway.

### DIFF
--- a/src/output-gazetteer.hpp
+++ b/src/output-gazetteer.hpp
@@ -34,7 +34,7 @@ public:
                        std::shared_ptr<thread_pool_t> thread_pool,
                        options_t const &options)
     : output_t(mid, std::move(thread_pool), options),
-      m_copy(std::make_shared<db_copy_thread_t>(get_options()->conninfo)),
+      m_copy(std::make_shared<db_copy_thread_t>(options.conninfo)),
       m_proj(options.projection)
     {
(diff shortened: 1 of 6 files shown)


src/output-gazetteer.hpp
```diff
@@ -32,9 +32,9 @@ class output_gazetteer_t : public output_t
     /// Constructor for new objects
     output_gazetteer_t(std::shared_ptr<middle_query_t> const &mid,
                        std::shared_ptr<thread_pool_t> thread_pool,
-                       options_t const &options,
-                       std::shared_ptr<db_copy_thread_t> const &copy_thread)
-    : output_t(mid, std::move(thread_pool), options), m_copy(copy_thread),
+                       options_t const &options)
+    : output_t(mid, std::move(thread_pool), options),
+      m_copy(std::make_shared<db_copy_thread_t>(get_options()->conninfo)),
```
lonvia, line 37: Just use `options.conninfo` here to be consistent with m_proj initialisation below?

Dito above in output-flex.cpp.

joto: I have added a commit fixing this (and also changing the parameter name `o` to `options` in some places for consistency, and short names are bad anyway).

PR #1839, opened 2022-11-28
Check that database is using UTF-8 encoding

joto:
Otherwise osm2pgsql might not work correctly.

### DIFF
--- a/src/db-check.cpp
+++ b/src/db-check.cpp
@@ -51,6 +51,10 @@ void check_db(options_t const &options)
                     get_minimum_postgresql_server_version())};
         }
 
+        if (settings.at("server_encoding") != "UTF8") {
+            throw std::runtime_error{"Database is not using UTF8 encoding."};
+        }
+
         auto const postgis_version = get_postgis_version(db_connection);
         log_info("PostGIS version: {}.{}", postgis_version.major,
                  postgis_version.minor);


pnorman (changes requested):
This change will check the server encoding, not the database encoding.

Do we want to consider client encoding at all?

joto:
> This change will check the server encoding, not the database encoding.

Is there a difference? When I create a database with `createdb -E xxx`, the settings returned for `server_encoding` is `xxx`, so I believe this is the correct way of checking the encoding.

> Do we want to consider client encoding at all?

Do we ever had problems with that? We could simply set the `client_encoding` in the connection string.

pnorman:
> Is there a difference? When I create a database with `createdb -E xxx`, the settings returned for `server_encoding` is `xxx`, so I believe this is the correct way of checking the encoding.

Looks okay after testing. I didn't realize that pg_settings was per-DB.

> Do we ever had problems with that? We could simply set the client_encoding in the connection string.

We probably should if we're worried about non-UTF8 encodings.

lonvia:
I've run into problems on machines where the database was UTF-8 but the cluster wasn't. It might just be because we don't set the client encoding. Usually the cluster ended up not being utf-8 because the machine itself wasn't configured for it.

joto:
Addition PR for the client thing here: https://github.com/openstreetmap/osm2pgsql/pull/1840

pnorman (approved):

PR #1843, opened 2022-12-01
Consolidate database capability checks

joto:
There were several places (pgsql.hpp, pgsql-capabilities.hpp, db-check.hpp) that had code to check various capabilities, version numbers etc. of the database we are connected to. This is now consolidated in pgsql-capabilities.hpp. It needs to be initialized once at program start by calling init_database_capabilities() with a database connection object. Checking version numbers or capabilities is then much easier, because no active database connection is needed any more.

This also adds code to get the list of supported index methods from the database which we will need shortly.


.github/workflows/ci.yml
```diff
@@ -39,24 +39,6 @@ jobs:
...
-      LUAJIT_OPTION: ON
-      POSTGRESQL_VERSION: 9.5
-      POSTGIS_VERSION: 2.4
-      BUILD_TYPE: Release
-
-    steps:
-      - uses: actions/checkout@v3
-      - uses: ./.github/actions/ubuntu-prerequisites
-      - uses: ./.github/actions/build-and-test
-
```
lonvia, line 59: Are you sure you want to loose the test with gcc7?

Issue #1780, opened 2022-09-28
Making indexes more flexible


joto:
*All of the following concerns only the flex output. We are not planning to change any of the other outputs.*

One of the most asked for features is to have more functionality around indexes:

* #27
* #37
* #1311
* #1357
* #1691

## Current situation

Osm2pgsql will always create indexes on the id columns of all tables if the database is created for update (with `--slim` and without `--drop`). These indexes are needed for osm2pgsql itself, so it can do updates properly. This will not change.

In addition osm2pgsql always creates a GIST index on the geometry column of each table. If a table doesn't have a geometry column, no index is created. If it has several, only the first will get the index.

Indexes are created in parallel (up to the number set with `--number-processes`), unless `--disable-parallel-indexing` is set.

## Features we want

* Allow tables without any indexes
* Allow indexes on any column (of any column type)
* Allow indexes on several columns at once
* Allow indexes on expressions
* Allow any type of index (btree, gist, ...)
* Allow setting of WHERE conditions
* Allow setting of TABLESPACE
* Allow setting of other options like UNIQUE etc.

See also the [PostgreSQL docs for indexes](https://www.postgresql.org/docs/current/sql-createindex.html).

Also any solution must be backwards compatible, so running osm2pgsql with an unchanged config file and command line options must still do the same.

## Proposal

Add `indexes` field to the options given to the [`osm2pgsql.define_table()` command](https://osm2pgsql.org/doc/manual.html#defining-a-table).

The `indexes` fields can have one of the following values:
* `nil` (default) means to create an index on the first geometry field. This takes care of the backwards compatibilty.
* `false` means that we do not want any indexes on this table.
* a list of index definitions describing what indexes we want. The list may be empty which is the same as setting `indexes` to `false`.

## Examples

Here is an examples config showing the old way of doing things:

```Lua
local data_table = osm2pgsql.define_table{
    name = "data",
    ids = { type = 'any', id_column = 'osm_id', type_column = 'osm_type' },
    index = nil,
    columns = {
        { column = 'name', type = 'text' },
        { column = 'tags',  type = 'jsonb' },
        { column = 'geom',  type = 'geometry' }
    }
}
```

You can set this instead if you don't want any index:

```Lua
    indexes = false,
```

Or this does the same:

```Lua
    indexes = {},
```

Or you can indicate exactly what you want:

```Lua
    indexes = {{ column = 'geom', method = 'gist', fillfactor = 'auto' },
[...]

PR #1847, opened 2022-12-05
Flex: Add support for configuring table indexes in the Lua config

joto:
The table definitions have a new (optional) field called "indexes" now which takes a list of index definitions. If the field is not there, we fall back to what we did before, a GIST index on the only/first geometry column of table is created. To disable indexes, set to an empty array.

See the flex-config/indexes.lua Lua config for some usage examples.

See #1780

### DIFF
--- a/src/flex-index.hpp
+++ b/src/flex-index.hpp
@@ -1,0 +1,100 @@
+#ifndef OSM2PGSQL_FLEX_INDEX_HPP
+#define OSM2PGSQL_FLEX_INDEX_HPP
+
+/**
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ * This file is part of osm2pgsql (https://osm2pgsql.org/).
+ *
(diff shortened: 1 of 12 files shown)


src/flex-index.hpp
```diff
@@ -0,0 +1,96 @@
...
+class flex_index_t
+{
+public:
+    explicit flex_index_t(std::string method) : m_method(std::move(method)) {}
+
+    std::string const &method() const noexcept { return m_method; }
+
+    std::string columns() const;
+
+    void set_columns(std::string const &columns)
```
lonvia, line 29: That's rather 'add_column'?

joto: This is not intended for adding columns one at a time, but an overload of the `void set_columns(std::vector<std::string> const &columns)` function so that we don't have to create a vector if all we want is to add a single column.

joto: I added an assert.

Issue #1859, opened 2022-12-23
Does not build with fmt 9


AMDmi3:
As mentioned in #1653, libfmt no longer supports `_format` so osm2pgsql does not build with up to date libfmt 9.x

joto:
Duplicate of #1704.

PR #1874, opened 2023-01-05
Use fmt::format() function instead of ""_format literal

joto:
This replaces most uses of the ""_format literal with function calls to fmt::format(). The latter form is available in all versions of the fmt library, while the ""_format literal is deprecated in version 8 and removed in version 9.

This might cost us some performance and early error detection, because in at least some versions of the fmt library the _format() construct was used to compile the format patterns at run time. That's why there are some uses of the literal left in places where performance might be more critical, they need a closer look.

See #1859

### DIFF
--- a/src/flex-table.cpp
+++ b/src/flex-table.cpp
@@ -150,15 +150,16 @@ flex_table_column_t &flex_table_t::add_column(std::string const &name,
 std::string flex_table_t::build_sql_prepare_get_wkb() const
 {
     if (has_multicolumn_id_index()) {
-        return "PREPARE get_wkb(char(1), bigint) AS"
-               " SELECT \"{}\" FROM {} WHERE \"{}\" = $1 AND \"{}\" = $2"_format(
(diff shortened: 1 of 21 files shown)


src/flex-table.cpp
```diff
@@ -347,11 +351,12 @@ void table_connection_t::stop(bool updateable, bool append)
...
-                sql += "ST_GeoHash({},10)"_format(geom_column_name);
+                sql += fmt::format("ST_GeoHash({},10)",
+                                   table().geom_column().name());
             } else {
-                sql +=
-                    "ST_GeoHash(ST_Transform(ST_Envelope({}),4326),10)"_format(
-                        geom_column_name);
+                sql += fmt::format(
+                    "ST_GeoHash(ST_Transform(ST_Envelope({}),4326),10)",
+                    table().geom_column().name());
```
lonvia, line 359: Why remove the quotes?

joto: Looks like I botched the rebase. Fixed commit on the way...

PR #1897, opened 2023-01-23
Give each database connection a unique id and use it for logging

joto:
This also logs new connections with the PID of the backend used and when the connection is closed.

### DIFF
--- a/src/pgsql.hpp
+++ b/src/pgsql.hpp
@@ -23,6 +23,7 @@
 #include <libpq-fe.h>
 
 #include <array>
+#include <atomic>
 #include <cassert>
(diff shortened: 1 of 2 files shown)


src/pgsql.hpp
```diff
@@ -289,6 +290,12 @@ class pg_conn_t
     };
 
     std::unique_ptr<PGconn, pg_conn_deleter_t> m_conn;
+
+    // Used to geneate unique ids for each database connection.
```
lonvia, line 294: typo

joto: Fixed.

PR #1902, opened 2023-02-06
Various changes in preparation for flexible expiry config

joto:
These are some changes in preparation for more flexible expiry config that can be configured from Lua.

See individual commits.


joto:
I am withdrawing this in favour of #1906. The other changes will be in a new PR later.

PR #1910, opened 2023-02-10
Don't process objects without tags in pgsql/flex output in append mode

joto:
These were already not processed in create mode. In append mode we do need to remove any rows with those ids in the output tables, but we don't have to create new rows in any tables.

### DIFF
--- a/src/osmdata.cpp
+++ b/src/osmdata.cpp
@@ -61,9 +61,15 @@ void osmdata_t::node(osmium::Node const &node)
         return;
     }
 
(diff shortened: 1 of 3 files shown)


lonvia:
This looks inconsistent to me. To be consistent with *_add, it is osmdata that should handle empty tags. A *_modify on an object with no tags should translate into a *_delete.

joto:
I have fixed the PR along those lines and also added new commits simplifying the osmdata code.

PR #1917, opened 2023-02-19
extended make_clean_tags_func to make whitelisting possible

asciibeats:
I like to have full control over which tags get through, so like the title says i changed the function a bit. It does not break anything but you can now run `osm2pgsql.make_clean_tags_func(keep_keys, true)` and delete all other tags. I hope that is ok with you. Thanks for your work. I am having a lot of fun with it. Cheers!

### DIFF
--- a/src/init.lua
+++ b/src/init.lua
@@ -79,46 +79,55 @@ function osm2pgsql.make_check_values_func(list, default)
     end
 end
 
-function osm2pgsql.make_clean_tags_func(keys)
-    local keys_to_delete = {}
-    local prefixes_to_delete = {}
-    local suffixes_to_delete = {}
+function osm2pgsql.make_clean_tags_func(keys, whitelist)
+    whitelist = whitelist or false;
+
+    local keys_to_consider = {}
+    local prefixes_to_consider = {}
+    local suffixes_to_consider = {}
 
     for _, k in ipairs(keys) do
         if k:sub(-1) == '*' then
-            prefixes_to_delete[#prefixes_to_delete + 1] = k:sub(1, -2)
+            prefixes_to_consider[#prefixes_to_consider + 1] = k:sub(1, -2)
         elseif k:sub(1, 1) == '*' then
-            suffixes_to_delete[#suffixes_to_delete + 1] = k:sub(2)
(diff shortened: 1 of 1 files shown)


joto:
Sorry, I don't understand what this is for. Can you give an example?

asciibeats:
I am creating a game. The map is not the center piece but just a tool in context of the game. I do not need to surface all available tags to the user to increase the usefulness of the map. In fact I think it would be counterproductive, since it is not meant to be a tool for the real world. Every tag I utilize has a specific purpose. All other tags are overhead from my perspective. Dead weight. And since now the function is not more complicated and even a bit more efficient than before while covering both use cases - black- and whitelisting - i thought people might appreciate it.

pnorman:
If you don't need all tags, just create columns for what you need.

asciibeats:
I like the flexibility of tags. My workflow right now looks like this: I whitelist something new , make sure that the objects land in the designated table (nodes, ways or relations) and add some specific styling client-side. No need to alter or create any tables. Btw i use PostGIS’s ST_AsMVT. It’s perfect like that. And again, my function does not bloat anything but provides, for cases like mine, a sensible gain in versatility.

lonvia:
Thank you for this PR, @asciibeats, and for sharing your solution. Unfortunately, a PR is not the best place for this at the moment. We maintainers are still in discussion what is the best way to simplify writing Lua flex scripts for users. The little utility library that you added to hear is one way to go. Another way is with separate library, which could be more complex. Until the matter is decided, we won't take PRs with new features for the utility library. So I have to close this PR.

That doesn't mean that your use-case and your solution are not interesting. On the contrary, I'd like to encourage more sharing of experience with the flex output. So I have opened up a new discussion category [Flex Recipes](https://github.com/openstreetmap/osm2pgsql/discussions/categories/flex-recipes) for users to share and disucss tips and tricks around writing code for Lua flex. I hope you will repost your function there. Seeing what users do with flex will certainly help the maintainers a lot to make a decision on the utility library in the future.

asciibeats:
Thank you for your reply. As you proposed, i added an [entry to Flex Recipes](https://github.com/openstreetmap/osm2pgsql/discussions/1955).

PR #1921, opened 2023-02-21
Two more changes to expire output

joto:
See individual commits.


joto:
Closing in favour of #1922 .

PR #1925, opened 2023-02-23
Add operators to make points sortable

joto:
The order is more or less arbitrary. This is only so that there is some kind of order for comparing sets of points etc.

### DIFF
--- a/src/geom.hpp
+++ b/src/geom.hpp
@@ -84,6 +84,23 @@ public:
         return !(a == b);
     }
 
+    /// Give points an (arbitrary) order.
+    [[nodiscard]] constexpr friend bool operator<(point_t a, point_t b) noexcept
+    {
+        if (a.x() == b.x()) {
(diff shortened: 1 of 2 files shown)


src/geom.hpp
```diff
@@ -84,6 +84,23 @@ class point_t
         return !(a == b);
     }
 
+    /// Give points an (arbitrary) order.
+    [[nodiscard]] constexpr friend bool operator<(point_t a, point_t b) noexcept
+    {
+        if (a.x() == b.x()) {
+            return a.y() < b.y();
```
pnorman, line 91: This branch is not tested - the test coverage has two points with equal y, but not ones with equal x.

pnorman (changes requested):

src/geom.hpp
joto: I have amended the test.

PR #1928, opened 2023-02-23
Move Lua setup code into its own file

joto:


joto:
This will avoid code duplication with the coming generalization code.

PR #1932, opened 2023-02-25
Docs about testing

JakobMiksch:
Adapt the documentation to make it easier to understand for new people.


JakobMiksch:
@lonvia the paragraph below is unclear to me. Maybe we can refine it a bit.

https://github.com/openstreetmap/osm2pgsql/blob/master/CONTRIBUTING.md?plain=1#L123-L126

PR #1938, opened 2023-02-27
Add functionality to send data to PostgreSQL in binary

joto:
Prepared statements can now send parameters in binary as well as in the default text format. To do this, wrap the binary data in an instance of the binary_param class, which is just a wrapper of std::string_view used to signal to the exec_prepared() function that this parameter should be sent in binary.

You can mix and match any number of parameters in text or binary format.

It isn't possible any more to use exec_prepared() without any parameters, but that wasn't use (and doesn't make much sense anyway), so no loss there.

### DIFF
--- a/src/pgsql.hpp
+++ b/src/pgsql.hpp
@@ -127,6 +127,20 @@ private:
 };
 
 /**
+ * Wrapper class for query parameters that should be sent to the database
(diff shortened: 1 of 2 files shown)


joto:
Closing because it needs other changes first.

PR #1942, opened 2023-03-02
Add support for data generalization

joto:
This large commit adds a framework to generalize OSM data using different strategies and implements several strategies. The source is all in the new `src/gen` directory. It includes a new command `osm2pgsql-gen` to access this functionality. Future versions might integrate it with the `osm2pgsql` command, but because this is all experimental it is kept separate for now. Call `osm2pgsql-gen` with `-h` to get usage help.

See also https://osm2pgsql.org/generalization/ and the chapter on Generalization in the manual:
https://osm2pgsql.org/doc/manual.html#generalization

For the raster support this adds two new library dependency: CImg and potrace. Both have been around for a long time and are readily available on all systems.

The following strategies work on a tile-by-tile basis and operate on polygons:

The "vector-union" strategy buffers and unionizes polygons using vector operations.

The "raster-union" strategy does a similar thing but does it in raster space which is much faster. First the polygons are rendered into a raster, an open/close operation is called (which basically does the same thing as the buffering in vector space) and finally the resulting raster is vectorized again.

The "builtup" strategy is intended to derive a layer of builtup areas from landuse=residential/industrial etc. as well as building cover and dense road networks.

The following strategies always work on all data (not tile-based):

The "discrete-isolation" strategy rates places based on some importance metric to get a more even distribution of places on the map.

The new "rivers" strategy finds important rivers, this is still very much work in progress.

### DIFF
--- a/src/gen/canvas.cpp
+++ b/src/gen/canvas.cpp
@@ -1,0 +1,130 @@
+/**
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ * This file is part of osm2pgsql (https://osm2pgsql.org/).
+ *
+ * Copyright (C) 2006-2023 by the osm2pgsql developer community.
+ * For a full list of authors see the git log.
+ */
+
+#include "canvas.hpp"
+#include "raster.hpp"
+
+cimg_library::CImg<int> canvas_t::create_pointlist(geom::point_list_t const &pl,
+                                                   tile_t const &tile) const
+{
+    cimg_library::CImg<int> points{static_cast<unsigned int>(pl.size()), 2};
+
+    int n = 0;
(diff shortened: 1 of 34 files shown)


src/gen/canvas.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+void canvas_t::save(std::string const &filename) const
+{
+    m_rast.save(filename.c_str());
+}
+
+std::string canvas_t::to_wkb(tile_t const &tile, double margin) const
+{
+    std::string wkb;
+    wkb.reserve(61 + 2 + m_rast.size());
```
pnorman, line 90: Use variables for magic numbers

PR #1950, opened 2023-04-08
Enhance parameter checking for generalization

joto:


pnorman (commented):
Other parameters are validated but this one lives at the beginning of the function. Is it worth standardizing that?

joto:
@pnorman I don't understand your question? Can you elaborate?

pnorman:
Looking at `src/gen/gen-tile-raster.cpp` and `m_image_extent` as an example, the validation of it is right before its use.

pnorman (approved):

PR #1961, opened 2023-05-15
Various cleanups/fixes found by clang-tidy

joto:
See individual commits.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -83,4 +83,8 @@ CheckOptions:
       value: true
     - key: cppcoreguidelines-non-private-member-variables-in-classes.IgnoreClassesWithAllMemberVariablesBeingPublic
       value: true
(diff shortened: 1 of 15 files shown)


pnorman:
I'm not sure that changes like changing `1024 * 1024` to `1024UL * 1024UL` improve readability.

joto:
@pnorman I am not sure either. But the alternatives are to

a) disable the warning in general which can hide some problem
b) disable this warning specifically with a comment which is even worse
c) ignore the messages which means I'll have to check the results of th clang-tidy run every time and remember which warnings were important and which weren't

So I choose to "give in" in this case. More generally managment of the clang-tidy warnings (and similar for compiler warnings) is a careful management of selecting to act on the warning or do one of the above. There are some compromises necessary and different developers will probably make different choices on that. But overall having those checks has found so many bugs that it is absolutly useful to have them and in general act on them.

PR #1963, opened 2023-05-19
Store settings to a database table and retrieve for updates

joto:
Store command line settings in a database table called "osm2pgsql_settings" when importing data. This is mainly so that we can do updates without specifying those settings again (or checking that the settings match the ones used for import.)

The table always has the same name and will be stored in the schema used for the middle tables (set with --middle-schema).

Adds a new generic settings_t class which can store string, integer and boolean settings. Internally a map of strings is used for the settings, in the database a simple table with two text columns: option and value. For boolean values strings 'true' and 'false' are used.

Currently the following settings are stored in the database:
- "attributes" (bool): -x/--extra-attributes setting
- "flat-nodes": flat node file name
- "prefix": prefix from --prefix
- "updatable" (bool): can the database be updated, i.e. created with --slim and no --drop
- "version": osm2pgsql version


pnorman:
This seems focused on the pgsql backend. I see a few issues
1. What if someone wants to import into the same DB with multiple prefixes (assuming non-updatable, so no slim tables)
2. I view the flat nodes option similar to `-d` and `--middle-schema` in that it sets where data is going to be stored. I don't think filesystem paths belong in the DB, as they vary depending on the working directory.

joto:
> This seems focused on the pgsql backend.

No. This is mostly about the middle and about updates. The idea is that osm2pgsql can tell what parameters to use with updates. We'll have more settings in the future, the replication diff stuff can use this mechanism. And mostly this is preparation for a new middle database scheme. We can use those settings to store which scheme is in use.

> What if someone wants to import into the same DB with multiple prefixes (assuming non-updatable, so no slim tables)

Hm. That is a problem. The only way to fix this I can think of would be to not write the settings table in non-slim mode.

> I view the flat nodes option similar to -d and --middle-schema in that it sets where data is going to be stored. I don't think filesystem paths belong in the DB, as they vary depending on the working directory.

Storing the flat nodes file in the database is important, because this is a problem that comes up every so often. Users want to update a database and forget to mention the flat node file or stuff like that. Storing the file name in the database makes this easier, but the setting can be overridden with a new `--flat-nodes` setting, so you can still move the file if you want to. But there is a problem: Files should be stored as absolute file names so they don't depend on the current working directory. I had planned to do that and forget to implement that.

lonvia:
> >   What if someone wants to import into the same DB with multiple prefixes (assuming non-updatable, so no slim tables)

> Hm. That is a problem. The only way to fix this I can think of would be to not write the settings table in non-slim mode.

Is it? The second import will simply overwrite the settings table. But without updates, the information is not used anyway, so no harm done. If somebody wants to preserve the settings, they need to use different middle schemas for the imports.

joto:
Maybe not a problem, but could be rather confusing. Especially if we store more information like import timestamp and so on. But having several datasets in the same database with different prefixes is a bit strange to begin with, using different schemas makes more sense anyway. So I don't think this would be a very common use case.

joto:
I have updated the PR to use absolute filenames for the `flat-nodes` setting.

joto:
Closing. Updated PR is in #1964. (Mostly just changes "settings" to "properties" in that one plus minor fixes.)

PR #1965, opened 2023-05-23
Fix: Make sure attributes don't show up in tags in flex output

joto:


joto:
Better solution coming up...

PR #1966, opened 2023-05-24
Refactor middle: Pull special osm_* tags out of the tag list

joto:
The middle stores OSM attributes (version, timestamp, changeset, uid, user) in "osm_*" pseudo-tags. When we get data back from the middle, we used to just leave them in there. This commit changes that behaviour: We pull the data out and put it into the proper attributes and remove the special tags. This is how the flex output needs the data and this is how a new middle (we are currently working on) will provide the data.

The pgsql output puts the pseudo-tags in already, because it had to do that on import anyway.

This also fixes a problem where the "osm_*" pseudo-tags were visible in the Lua config file of the flex output.

### DIFF
--- a/src/middle-pgsql.cpp
+++ b/src/middle-pgsql.cpp
@@ -156,6 +156,54 @@ static char const *decode_to_delimiter(char const *src, std::string *dst)
 }
 
 namespace {
+
+/**
+ * Go through tags returned from a middle table, get the attributes from the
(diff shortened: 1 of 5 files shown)


joto:
I am not totally clear on what the gazetteer output does with the attributes, so there might be an issue there.

lonvia:
gazetteer only supports reading the attributes from the osm structure, so that is fine, too.

Issue #1502, opened 2021-05-16
Planning for backwards incompatible changes


joto:
At some point in the not too far future (**not** the next version as I write this) we want to make a backwards incompatible change to the pgsql middle which means everybody has to do a re-import. If we have a breaking change anyway, this might be the chance to do some other changes, too. This issue is here to collect the different things we might want do, see individual issues for discussions and details.

Planned or possible changes:

* [X] Format of the database middle tables changes. (See #1384) 
* [X] Change default value for `--cache/-C` parameter? (See #1501) -> Default is okay, still works for small OSM files.

Are there some other things "we always wanted to do" but didn't for compatibility?

joto:
Just to make this clear: There is no decision yet how to make this backwards incompatible change. It might be possible to do some kind of phased change over a longer time. But that has some cost (mainly in development time), too. So maybe a clear breaking change is better. But to figure this out it is good to collect ideas how such a change could look and what else we might want to do. That's why I created this issue.

joto:
* Removal of gazetteer. Already deprecated in 1.8.0. (See #1870).
* Removal of `--with-forward-dependencies`. Already deprecated in 1.8.0. (See #1870).
* Flex Lua command `add_row()` will at some point be deprecated/removed in favour of `insert()`, no schedule for this yet.

joto:
* Remove `area` column type in flex output. It only works with `add_row()`, not with `insert()`.
* From 1.9.0 expire configuration for the flex output will be done in the Lua config file, but the old command line options are (as I write this) still supported. We might want to change this. The command line options are: `-e, --expire-tiles`, `-o, --expire-output`, and `--expire-bbox-size`.
* Do we want to remove support for the non-bucket way-node-index in the middle?
* Do we want to rename the middle tables in the new middle?
* `pole_of_inaccessibility()` and generalization support is currently still marked as experimental.

joto:
It is unclear how the `--b, -bbox` option is supposed to work and how well it works, especially with updates. There are better solutions. i.e. creating an extract first and feeding that to osm2pgsql.

Maybe it is time to get rid of it.

Anybody using this option?

joto:
[...]

PR #1969, opened 2023-05-27
WORK IN PROGRESS: New database middle

### LINKED ISSUE #692: Store slim table tags in hstore
I did some testing, and the representation of tags in a hstore is more efficient than our current `text[]` array of alternating keys and values. For testing latter can be turned into the former with `hstore(tags)`.

Doing this on the slim table of a planet and summing the `pg_column_size` gives a size of 38GB for `text[]` representation and 30GB for `hstore`. The overall size of the current slim table is 94GB for the table and 190GB for the indexes so this is about a 5% space savings.

I considered jsonb, but it is the same size as hstore.

This change would require the hstore extension for osm2pgsql databases. I do not consider this an issue because hstore is a standard part of contrib, and many databases require it already.

In addition to being more efficient, if someone did want to use our internal slim tables to query by tag, it is much easier to do so with hstore.

### LINKED ISSUE #1170: In-band metadata in middle layer
Let's assume I've created some Overpass extract which doesn't include metadata (i.e. `out;` instead of `out meta;` ). With GDPR in mind, this could also mean that some metadata fields like user, uid and changeset may be missing in the future.

Now, when processing such a file via osm2pgsql with extra attributes enabled, a mapper (!) can control the metadata fields due to the logic in taglist_t::add_attributes. This way, random values can be written to the metadata fields. Even worse, setting some special purpose keys to invalid values can abort osm2pgsql altogether.

Example test data:

```
<?xml version="1.0" encoding="UTF-8"?>
<osm version="0.6" generator="Overpass API 0.7.56.3 eb200aeb">
 <node id="25477618" lat="49.1837670" lon="-122.9572343">
  <tag k="highway" v="traffic_signals"/>
  <tag k="osm_uid" v="x"/>
 </node>
</osm>
```

Result: Osm2pgsql failed due to ERROR: illegal user id: 'x'

./osm2pgsql ---create --slim  -O flex -S ../flex-config/simple.lua demo.osm -x

joto:
The database format we have been using for the middle has some problems:

* Node tags are not stored in the planet_osm_nodes table, because they are not needed for updates. But having access to those tags is useful in some cases, for instance when relations are processed after import.
* Attributes (version, timestamp, changeset, uid, user) of ways and relations are stored as special pseudo-tags ("osm_*") (if `--extra-attributes` is used). This has the potential of name clashes and format problems and makes the attributes difficult to access from the database. Attributes for nodes are never stored.
* The way we store tags as array of text fields with keys and values intermixed ([key1, value1, key2, value2, ...]) is cumbersome to use from the database.
* The way relation members are stored is rather arcane.
* When using `--extra-attributes/-x` the middle tables become huge (due to storage in pseudo-tags).

This commit fixes all those problems introducing a new database structure:

* Tags are stored in JSONB columns.
* The nodes table gets a new "tags" column.
* Attributes are optionally stored in normal typed database columns. The columns are only added when --extra-attributes is specified and the columns can be `NULL` if not used which makes the overhead tiny.
* Relation members are now stored as JSONB as an array of objects, for example: `[{"type": "W", "ref": 123, "role": "inner"}, ...]`. Using JSONB allows us to build the indexes needed to find all relations with certain members.
* The format for way nodes has been kept as an array of bigints.

The names of the tables PREFIX_nodes, PREFIX_ways, and PREFIX_rels (with "osm_planet" as default prefix) has been kept, but we might want to change this and get rid of the prefix, schemas are a better mechanism and they have been available for a while.

There is a new table PREFIX_users which contains a user id->name lookup table. The user name isn't stored in the other tables, just the id. This saves disk space and has the added benefit of updating the user name correctly if a user name changes.

There are two new command line options:
* `--middle-database-format=NUM` - (1 for old format, 2 for new format)
* `--middle-with-nodes` - set this to store tagged nodes in the database even if a flat-node file is used. Untagged nodes are only stored in the database if there is no flat-node file.

For the first time this new format allows you to have a database created by osm2pgsql that contains *all* the information in an OSM file, all nodes, ways, and relations with all their tags and attributes.

Note that this commit currently doesn't have a mechanism for detecting which format a database has when doing updates. This will be added later. For the time being you have to use the same command line options for updates that were used on import.

This commit doesn't properly do testing. If you want to run all the tests with the new database format, set `middle_database_format = 2` in options.hpp and recompile.

This commit adds a new dependency on a [JSON library](https://github.com/nlohmann/json). Parsing JSON isn't something we want to do ourselves. This library has been around for a while, is available everywhere and is well supported with regular releases unless the RapidJSON library we were using before.

Closes #692
Closes #1170
See #1502


mboeringa:
Sounds great, I am sure some people will find use for the new functionality, although for the majority of users, this will likely - and luckily - be more of an invisible background change to the internal processing of osm2pgsql, likely not requiring changes to the existing LUA flex style files.

> unless the RapidJSON library

I guess you meant to write "unlike"...

src/middle-pgsql.cpp
```diff
@@ -28,7 +28,10 @@
 #include <osmium/memory/buffer.hpp>
 #include <osmium/osm/types_from_string.hpp>
 
+#include <nlohmann/json.hpp>
```
pnorman, line 31: This library is not documented in the readme as being required, resulting in compilation errors.

pnorman (changes requested):
This would select a column type for osm object creation timestamps as `timestamp without time zone`, which we were not able to agree on in #1970. I recommend we seek out postgres expertise from `#postgreql@irc.libera.chat`, since I don't think further discussions with just us will result in progress.

src/middle-pgsql.cpp
joto: I have added it to the README and also to the CMake config.

lonvia:
The reason we can't agree on the timestamp data type is that none of them is suitable for our use case. No matter if you choose with or without timezone, there are use cases where one or the other type results in surprising behaviour. I've certainly wasted debugging hours a few times because Postgresql took it upon itself to convert timestamps with timezones without being asked.

Looking at the SQL necessary to fix the annoying behaviour, timestamps with timezones come out a little bit better, so we should probably go for them. The alternative is to just save unix timestamps in BIGINTs and get equal-opportunity inconvenience for everybody. Fine with me, too.

joto:
I have changed the type to "timestamp with timezone", rebased and fixed the merge failure.

pnorman (approved):

src/options.hpp
```diff
@@ -146,6 +146,16 @@ struct options_t
      */
     uint8_t way_node_index_id_shift = 5;
 
+    /// Database format (0=unknown/no database middle, 1=legacy, 2=new)
+    uint8_t middle_database_format = 1;
```
lonvia, line 150: Isn't 0 the current default (aka ram mode)?

PR #1977, opened 2023-06-20
Make the display of time intervals show seconds.

amandasaurus:
Some people would rather have non-approximate times ([e.g.](https://github.com/openstreetmap/osm2pgsql/pull/1975#pullrequestreview-1486491844)). This patch makes the times always show seconds.

I don't have a strong preference. I am OK with this patch being accepted or not.

### DIFF
--- a/scripts/osm2pgsql-replication
+++ b/scripts/osm2pgsql-replication
@@ -54,25 +54,25 @@ LOG = logging.getLogger()
 OSM2PGSQL_PATH = Path(__file__).parent.resolve() / 'osm2pgsql'
 
 def pretty_format_timedelta(seconds):
-    minutes = int(seconds/60)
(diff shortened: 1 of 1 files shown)


scripts/osm2pgsql-replication
```diff
@@ -54,25 +54,24 @@ LOG = logging.getLogger()
...
+    output = []
+    # If weeks > 1 but hours == 0, we still want to show "0 hours"
+    if weeks > 0:
+        output.append("{} week(s)".format(weeks))
+    if days > 0 or weeks > 0:
+        output.append("{} day(s)".format(days))
+    if hours > 0 or days > 0 or weeks > 0:
+        output.append("{} hour(s)".format(hours))
+
+    output.append("{} minute(s)".format(minutes))
```
lonvia, line 71: This should now be conditional as well.

PR #1980, opened 2023-06-22
osm2pgsql-replication will now use postgresql's application_name

amandasaurus:
This can make it easier to monitor PostgreSQL server activity, e.g. with [`pg_stat_activity`](https://www.postgresql.org/docs/current/monitoring-stats.html#MONITORING-PG-STAT-ACTIVITY-VIEW).


scripts/osm2pgsql-replication
```diff
@@ -80,10 +80,10 @@ def connect(args):
     """
     # If dbname looks like a conninfo string use it as such
     if args.database and any(part in args.database for part in ['=', '://']):
-        return psycopg.connect(args.database)
+        return psycopg.connect(args.database, application_name="osm2pgsql-replication")
```
pnorman, line 83: I'd rather set `fallback_application_name` so that a user can potentially override with `PGAPPNAME` if they're running multiple DBs

pnorman (changes requested):

scripts/osm2pgsql-replication
amandasaurus: Done. 🙂
I didn't know about that funtionality.
Rebased & force pushed.

PR #1996, opened 2023-07-13
Middle: Get dependencies with single queries

### LINKED ISSUE #423: Reprocessing relations multiple times
https://github.com/openstreetmap/osm2pgsql/issues/419#issuecomment-126644507 showed that with diff updates, relations can be processed twice, once in parsing and once in pending relations. The output is correct, but doing so twice is wasteful.

Two possible solutions come to mind
- Whenever a relation is processed in parsing, delete it from the trackers. If you have a relation that is then modified later in the diff, it will be re-added to the tracker and processed in pending.
- Don't have the output tables process relations that are already flagged as pending. This would never process the same relation twice, but would defer more work to the pending stage, which is believed to be slower.

Bit of a brain-dump from IRC:

> my understanding was that in processing when a relation is touched it has each backend table process that relation. depending on tag transforms, some may do nothing with it. then in pending, it does the same. now, if you have a relation modified then later on a way in that relation modified, it will have to do it in pending so maybe its better to instead skip processing anything in processing that is already marked as pending

joto:
In append mode osm2pgsql has to find all parent ways and relations of changed nodes and parent relations of changed ways. This is done by with one query per object read from the input file which is quite wasteful. This changes the code to only do a single query for all dependencies. The list of all changed ids is COPYied into a temporary table. Then parent objects ids are found and written into another temporary table and then retrieved. (We could this without the extra temporary tables, but this could allow us later to use the content of those tables somewhere else, so I chose to do it this way. Should make a big difference perfomance-wise.)

These PR also fixes #423, the longstanding problem that ways and relations that are in the input file AND parents of changed objects were process twice.

The work is split between three commits, one each for nodes, ways, and relations processing.

I did some performance measurements, in every one the new code was more efficient. For smaller (minutely) diffs the difference is small, about 30% performance improvements measured over 2000 something diffs. At the other end of the spectrum the performance gain is huge. Updating a planet file with a diff with about two weeks of data was an order of magnitude faster (from >10h down to an hour).

### DIFF
--- a/src/dependency-manager.cpp
+++ b/src/dependency-manager.cpp
@@ -15,14 +15,7 @@
 
 void full_dependency_manager_t::node_changed(osmid_t id)
 {
-    for (auto const way_id : m_object_store->get_ways_by_node(id)) {
-        way_changed(way_id);
-        m_ways_pending_tracker.set(way_id);
-    }
-
-    for (auto const rel_id : m_object_store->get_rels_by_node(id)) {
-        m_rels_pending_tracker.set(rel_id);
-    }
+    m_changed_nodes.set(id);
 }
 
 void full_dependency_manager_t::way_changed(osmid_t id)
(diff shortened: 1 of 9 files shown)


mboeringa:
> Should make a big difference perfomance-wise.

I guess you meant to write "Shouldn't make a big difference performance-wise"?

PR #2000, opened 2023-07-14
Fix build with v10 of the fmt library

tomhughes:
This fixes the build to handle v10 of the fmt library - the bundled version is currently v9 but an external version might not be.

Implicit conversion of enums to the underlying type has been removed in v10 so this makes the conversion of the libpq status enum to the underlying type explicit

### DIFF
--- a/src/pgsql.cpp
+++ b/src/pgsql.cpp
@@ -17,6 +17,7 @@
 #include <cstdlib>
 #include <stdexcept>
 #include <string>
+#include <type_traits>
 #include <utility>
 
 std::size_t pg_result_t::affected_rows() const noexcept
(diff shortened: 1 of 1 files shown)


joto:
`fmt::underlying` was only introduced in fmt 9.0.0. So this would break on Ubuntu 22.04 for instance which has fmt 8.1.1. I think we need a different solution here. Either we use `std::underlying_type` which is available since C++11, or maybe we should even output the status not as a number but some symbolic name.

tomhughes:
I almost wrote it using `std::underlying_type` then I noticed their example (of using `fmt::format_as` to define a formatter) was using `fmt::underlying` so I used that instead...

tomhughes:
I've updated it to use `std::underlying_type_t` instead - that actually needs C++14 but 17 is already the default.

joto (approved):

PR #2004, opened 2023-07-15
Update some dependency information in README

joto:


README.md
```diff
@@ -84,14 +84,15 @@ On a Debian or Ubuntu system, this can be done with:
...
+  nlohmann-json3-dev pyosmium
 ```
 
 On a Fedora system, use
 
 ```sh
 sudo dnf install cmake make gcc-c++ boost-devel expat-devel zlib-devel \
-  potrace-devel cimg-devel \
+  potrace-devel cimg-devel json-devel \
   bzip2-devel postgresql-devel proj-devel proj-epsg lua-devel pandoc
```
lonvia, line 96: At least Fedora should have pyosmium, too. Not sure about CentOS but we might consider removing that completely anyway, given the state of affairs.

joto: There are two packages in Fedora: pyosmium and python3-osmium. Not sure which one to install.

@tomhughes ?

tomhughes:
The source package is `pyosmium` but the binary package that you need to install to use it is `python3-osmium`.

README.md
tomhughes: The source package is `pyosmium` but the binary package that you need to install to use it is `python3-osmium`.

joto: I have updated the commit accordingly.

PR #2007, opened 2023-07-20
Fix: Make sure index is used for type/id indexes in flex output

joto:
Without the cast to char(1) the index is not used.

In one case I have here this cut processing time in half! Not sure why this did not show up before (or why nobody noticed).

### DIFF
--- a/src/db-copy.cpp
+++ b/src/db-copy.cpp
@@ -64,7 +64,7 @@ void db_deleter_by_type_and_id_t::delete_rows(std::string const &table,
 
         fmt::format_to(std::back_inserter(sql),
                        ") AS t (osm_type, osm_id) WHERE"
-                       " p.{} = t.osm_type AND p.{} = t.osm_id",
+                       " p.{} = t.osm_type::char(1) AND p.{} = t.osm_id",
                        type, column.c_str() + pos + 1);
     } else {
         fmt::format_to(std::back_inserter(sql),


pnorman (approved):
It looks like what's happening is the VALUES has types of `(text, int)` or `(text, bigint)`, and because the join condition could be done as `p.foo::text = t.osm_type` or `p.foo = t.osm_type::char(1)` it picks the former because the latter would involve potential data loss.

There are three ways to resolve this
1. coerce the first (or all?) the values as documented for [quoted literal constants](https://www.postgresql.org/docs/current/sql-values.html)
2. use a temporary table, which will automatically coerce to the type of the columns when VALUES is used in the insert statement. This would also avoid potential out-of-memory failures for very large numbers of rows
3. coerce the type, as done in this PR

I'm somewhat inclined towards the third one, but would prefer to also coerce `t.osm_id` to bigint at the same time. This is only to be explicit about the type conversion, as if the values contains only numbers that can be represented as a normal integer it will get coerced into a bigint anyways when comparing to the p table, which has type bigint.

So I'm okay with the PR as-is, but some food for thought and other options.

mboeringa:
Did this issue only affect --slim import, not **non-slim** ones? 

I am asking because I noticed on [some recent tests](https://wiki.openstreetmap.org/wiki/Osm2pgsql/benchmarks#Planet_(75.0GB_PBF,_2023-07-10)_on_HP_Z840_workstation,_512GB_DDR4_2133,_2x_Xeon_E5-2699v4_22C,_5x_Samsung_970_EVO_Plus_2TB_NVMe_RAID_0,_4x_Intel_OPTANE_P1600X_110GB_NVMe_RAID_0,_Ubuntu_22.04_~8h) with a full Planet import, that the _relation_ stage loading is about 2-3 faster in **non-slim** mode (5k/s, versus 2k/s with --slim), where everything is stored in RAM.

Now the fact that everything is stored in RAM may be the main cause of the speed up in relation processing, but I would in that case at least also have expected some speed up of the way processing. However, in reality, I see a consistently slightly slower (by about 10%-15%) way loading stage (+/- 83k/s to 70k/s) in **non-slim** mode, which slightly surprises me as unexpected with all data in RAM. 

Note that this is a comparison with --slim mode import using also a --flat-node file, so maybe the specific structure of --flat-node is in fact more efficient than pulling nodes from RAM for way building on high performance NVMe backed hardware??

mboeringa:
> In one case I have here this cut processing time in half!

Just for the record: I see an about 25% faster "relation" loading stage on Planet with the latest master. 
+/- 11M relations loaded in 1.5h instead of 2h with **--slim** and **--flat-nodes**.

PR #2017, opened 2023-07-27
add :area() calculation

feludwig:
the 'area' type seems to make an error; changed to only 'real' and actually calculate :area(). When way_area IS NULL, no buildings nor multipolygons are rendered


joto:
@feludwig Thanks for the PR. There is clearly something wrong with the original code. Looks like there are some leftovers from earlier when we used to have `add_row` and not `insert` which would magically fill the `way_area` column if it had the `area` type and not `real`.

But your changes only fix part of the problem. I need to look into this a bit more.

PR #2019, opened 2023-07-29
Add spherical_area() function

joto:
To calculate the "real" area of a (multi)polygon. This is still an approximation. Results are similar to what ST_Area(geography, false) calculates in PostGIS. Accessible from Lua.

### DIFF
--- a/flex-config/geometries.lua
+++ b/flex-config/geometries.lua
@@ -34,7 +34,10 @@ tables.polygons = osm2pgsql.define_area_table('polygons', {
     -- if we have multiple geometry columns and some of them can be valid
     -- and others not.
     { column = 'geom', type = 'geometry', projection = 4326 },
+    -- In this column we'll put the area calculated in Mercator coordinates
     { column = 'area', type = 'real' },
+    -- In this column we'll put the true area calculated on the spheroid
+    { column = 'spherical_area', type = 'real' },
 })
 
 tables.boundaries = osm2pgsql.define_relation_table('boundaries', {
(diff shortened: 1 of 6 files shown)


flex-config/geometries.lua
```diff
@@ -34,7 +34,10 @@ tables.polygons = osm2pgsql.define_area_table('polygons', {
     -- if we have multiple geometry columns and some of them can be valid
     -- and others not.
     { column = 'geom', type = 'geometry', projection = 4326 },
+    -- In this column we'll put the area calculated in Mercator coordinates
     { column = 'area', type = 'real' },
+    -- In this column we'll put the true area calculated on the spheroid
```
mboeringa, line 39: > on the spheroid

Shouldn't this read 'on the sphere'?

By the way, I recommend you to thoroughly test this, also with data near the date line. I ran into serious issues with crashes with Planet wide (generalized) data and PostGIS ST_Area and spherical calculations when I first attempted implementing both spherical and spheroidal area calculations, and had to switch to spheroidal (ST_Area(geography,true)) to overcome these issues. ST_Area for some reason failed with spherical calculations. 

I have never been able to track down the real cause, and simply worked around it with the (slower) spheroidal calculations, but if this solution for osm2pgsql relies on some of the same libraries as PostGIS, it may encounter similar issues if it hasn't been fixed in the mean time.

joto: We don't use the same code PostGIS does but the code from boost::geometry which might or might not be better, different or whatever. I rely on that code like I do for everything we get from a library, so yes, there are some tests, but other than that I just expect this to work.

The description of the various area() functions are a bit fuzzy about whether they use a sphere or a spheroid and there are different versions of those functions etc. So I am going with spheroid because it sounds cooler and because technically a sphere is a special case of a spheroid so I am on the safe side. :-)

Issue #2011, opened 2023-07-24
Handling of schemas


joto:
We have an issue with the way we are handling schemas that we need to solve. This has been made worse [with the changes in PostgreSQL 15 around schema permissions](https://www.crunchydata.com/blog/be-ready-public-schema-changes-in-postgres-15).

Historically osm2pgsql didn't support schemas at all, everything was simply done without mentioning any schema, which normally meant tables, indexes, etc. ended up in the `public` schema which is always there and used to be writable by every database user. Then users wanted to be able to put tables, etc. into specific schemas, so we added functionality for that. If you set a schema (on the command line or in the flex config file), that schema is used, otherwise the former behaviour is used.

Note that there is a difference between "not using a schema at all" and "using the `public` schema". In the second case the schema is hardcoded to be `public`, it just happens to be the one that would also be used by default in many cases. In the first case no schema is specified, which means PosgtreSQL will use the `search_path` and uses the first schema that exists in that search path to create objects. By default the `search_path` is `"$user", public` which means if there is a schema with the name of the user running osm2pgsql, that schema is used, otherwise `public` is used.

It used to be that nobody has a schema with the name of their user, so no schema basically meant `public`. But with the change in Pg 15 this assumption is wrong. The `public` schema isn't writable any more in databases that you don't own, which lead to more people using schemas, specifically the schema with the same name as their user, which is what a lot of docs recommend you do. This breaks the implicit assumption that "no schema == `public`" which we have in our minds and which is also in our code in some places.

Now there are probably three common setups for osm2pgsql databases:
1. Use one user as owner of the database and use the **same** user for osm2pgsql. In that case you don't need any special schemas, the owner of the database can create objects in the `public` schema.
2. Use one user as owner of the database and use a **different** user for osm2pgsql. Add a schema with the same name as the user used by osm2pgsql, don't mention any schemas on the osm2pgsql command line/config. That schema should be used by osm2pgsql if the `search_path` was not changed from the default. (This case is currently broken by some assumption in the case that "no schema == `public`".)
3. Create specific schema(s) for osm2pgsql and explicitly use them on the command line/config. No problem there.

Now there is the question how we want things to be. There are basically two options:
[...]

PR #2027, opened 2023-08-07
Set default schema to 'public' and use qualified names everywhere

### LINKED ISSUE #2010: Latest master blocks running osm2pgsql multiple times consecutively with --create
## What version of osm2pgsql are you using?

Latest master:

> osm2pgsql --version
2023-07-23 22:08:09  osm2pgsql version 1.8.1 (1.8.1-183-g5dba2405)
Build: RelWithDebInfo
Libosmium 2.19.0
Proj [API 6] 8.2.1
Lua 5.1.4 (LuaJIT 2.1.0-beta3)

## What operating system and PostgreSQL/PostGIS version are you using?

Postgres version: 15.3 (Ubuntu 15.3-1.pgdg22.04+1)
PostGIS version: POSTGIS="3.3.3 2355e8e" [EXTENSION] PGSQL="150" GEOS="3.10.2-CAPI-1.16.0" PROJ="8.2.1" LIBXML="2.9.13" LIBJSON="0.15" LIBPROTOBUF="1.3.3" WAGYU="0.5.0 (Internal)"

## What did you do exactly?

With latest master, I imported once with the **--create** option, then a second time, again with **--create**. This failed with a warning of a pre-existing 'osm2pgsql_properties' relation. Note that the second run used _modified command line options_ for testing, compared to the first run.

> osm2pgsql -O flex -d gis -U osm --create -W --number-processes 32 -P 5433 -S 'osm2pgsql/flex-config/openstreetmap-carto.lua' 'data/planet-230717.osm.pbf'
2023-07-23 20:23:52  osm2pgsql version 1.8.1 (1.8.1-183-g5dba2405)
Password:
2023-07-23 20:24:00  Database version: 15.3 (Ubuntu 15.3-1.pgdg22.04+1)
2023-07-23 20:24:00  PostGIS version: 3.3
2023-07-23 20:24:00  Storing properties to table 'osm2pgsql_properties'.
2023-07-23 20:24:00  ERROR: Database error: ERROR:  relation "osm2pgsql_properties" already exists

## What did you expect to happen?

I would expect an existing 'osm2pgsql_properties' relation to be always dropped if I run with **--create**, or at least not block running with **--create** a second time and possibly update the existing values in the table.

## What did happen instead?

osm2pgsql blocks running with the --create option a second time if there is a pre-existing 'osm2pgsql_properties' relation.

joto:
Until now the default schema was empty and the PostgreSQL search path was used to find tables etc.

We now set the default schema for the middle and the pgsql and flex outputs to 'public'.

This simplifies the code, because we don't have to handle the case without schema specially any more. More importantly it makes it easier for the user to reason about what osm2pgsql is doing, because it does not depend on the setting of the search path in the PostgreSQL session.

This fixes a problem where osm2pgsql could not find the osm2pgsql_properties table it created itself, because it was created in a different schema than expected (#2010).

The behaviour of the gazetteer output has not changed. It still uses unqualified names.

This will be a breaking change for those users who rely on the old behaviour.

Fixes #2010, #2011.

### DIFF
--- a/docs/osm2pgsql.md
+++ b/docs/osm2pgsql.md
@@ -165,7 +165,7 @@ mandatory for short options too.
 
 \--middle-schema=SCHEMA
 :   Use PostgreSQL schema SCHEMA for all tables, indexes, and functions in
-    the middle (default is no schema, i.e. the `public` schema is used).
+    the middle (default is `public`).
 
 \--middle-way-node-index-id-shift=SHIFT
 :   Set ID shift for way node bucket index in middle. Experts only. See
@@ -275,8 +275,7 @@ mandatory for short options too.
 
 \--output-pgsql-schema=SCHEMA
 :   Use PostgreSQL schema SCHEMA for all tables, indexes, and functions in
-    the pgsql output (default is no schema, i.e. the `public` schema
-    is used).
+    the pgsql output (default is `public`).
 
 # EXPIRE OPTIONS
 
--- a/src/command-line-parser.cpp
+++ b/src/command-line-parser.cpp
@@ -181,7 +181,7 @@ Middle options:\n\
        --cache-strategy=STRATEGY  Deprecated. Not used any more.\n\
     -x|--extra-attributes  Include attributes (user name, user id, changeset\n\
                     id, timestamp and version) for each object in the database.\n\
-       --middle-schema=SCHEMA  Schema to use for middle tables (default: none).\n\
+       --middle-schema=SCHEMA  Schema to use for middle tables (default: 'public').\n\
        --middle-way-node-index-id-shift=SHIFT  Set ID shift for bucket index.\n\
        --middle-database-format=FORMAT  Set middle db format (default: legacy).\n\
        --middle-with-nodes  Store tagged nodes in db (new middle db format only).\n\
@@ -213,7 +213,7 @@ Pgsql output options:\n\
     -K|--keep-coastlines  Keep coastline data rather than filtering it out.\n\
                     Default: discard objects tagged natural=coastline.\n\
        --output-pgsql-schema=SCHEMA Schema to use for pgsql output tables\n\
-                    (default: none).\n\
+                    (default: 'public').\n\
        --reproject-area  Compute area column using web mercator coordinates.\n\
 \n\
 Expiry options:\n\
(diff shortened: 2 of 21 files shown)


mboeringa:
@joto 

After this change, I am now getting a 

`permission denied for schema public`

type error in PostgreSQL 15 when attempting to run osm2pgsql.

Shouldn't the schema default to the user's schema in PG15?

mboeringa:
The problem is that while I can set 'schema=X' in a flex style file to determine the schema to store all output tables, I cannot set the schema for where to store 'osm2pgsql_properties' table on the command line (or in the style file).

If you do not want to use `search_path`, as discussed in https://github.com/openstreetmap/osm2pgsql/issues/2011, then maybe use the user name as specified with the **-U** osm2pgsql command line option, assuming as you wrote there, that best practice is to use the user name also as the schema name. Then simply error out if this fails and you cannot write there?

joto:
@mboeringa This is all too much magic. Nothing says you have a schema with your user name. You have to say which schema to use and make sure the user has the permission and then everything works. The documentation has already been updated to explain this and the release notes will have some information about this. Yes, this is a breaking change for some users, but there simply is no way to handle this in a way that doesn't involve a lot of guessing of what the user setup might be and what we can do to work around this.

mboeringa:
@joto 

> The documentation has already been updated to explain this and the release notes will have some information about this.

This is exactly the issue I currently have. What do I need to do and where? The current documentation just says:

> Version >= 1.4.0 If you are using a security scheme based on database schemas in your database you can use the --middle-schema and --output-pgsql-schema options and the schema table option in the flex output, respectively, to tell osm2pgsql to load data into specific schemas. You have to create those schemas and give them the correct rights before running osm2pgsql.

**--middle-schema** is the option to set where the middle tables are stored
**--output-pgsql-schema** suggest according to name and text that is for setting the schema for old style **'pgsql'** output
**schema=X** allows me to set the schema for **'flex'** output in the flex style

But how do I specify the schema for the 'osm2pgsql_properties' table???

mboeringa:
Ah, I now see in another section of the manual it is **--middle-schema**. Besides reading over that section, part of the confusion stems from an attempt to run in **non-slim mode**, where technically no schema is needed in create-only mode for storing 'middle' tables... except now for the new 'osm2pgsql_properties' table.

PR #2037, opened 2023-08-14
Add new command line parameter --schema that sets the default schema

joto:
This will be used as default for --middle-schema, --output-pgsql-schema, and for the different ways of setting a schema in the flex output and generalizer code.

This removes the last places where the schema was hardcoded to "public" (except as a default for this command line parameter and in some legacy gazetteer code).


lonvia:
Note to self: needs similar changes in osm2pgsql-replication.

PR #2042, opened 2023-08-16
Fix invalid escape sequence warnings

tomhughes:
These warnings are normally hidden but when a bdd test fails you see them in the output:

```
steps/steps_db.py:28: SyntaxWarning: invalid escape sequence '\d'
  @then("table (?P<table>.+) has (?P<row_num>\d+) rows?(?P<has_where> with condition)?")
steps/steps_db.py:43: SyntaxWarning: invalid escape sequence '\d'
  @then("the sum of '(?P<formula>.+)' in table (?P<table>.+) is (?P<result>\d+)(?P<has_where> with condition)?")
steps/steps_execute.py:151: SyntaxWarning: invalid escape sequence '\w'
  @when("running osm2pgsql (?P<output>\w+)(?: with parameters)?")
steps/steps_execute.py:163: SyntaxWarning: invalid escape sequence '\w'
  @then("running osm2pgsql (?P<output>\w+)(?: with parameters)? fails")
steps/steps_execute.py:182: SyntaxWarning: invalid escape sequence '\d'
  @then("running osm2pgsql-replication fails(?: with returncode (?P<expected>\d+))?")
steps/steps_execute.py:193: SyntaxWarning: invalid escape sequence '\w'
  @then("the (?P<kind>\w+) output contains")
```

### DIFF
--- a/tests/bdd/steps/steps_db.py
+++ b/tests/bdd/steps/steps_db.py
@@ -25,7 +25,7 @@ def delete_table(context, table):
         cur.execute("DROP TABLE " + table)
 
 
-@then("table (?P<table>.+) has (?P<row_num>\d+) rows?(?P<has_where> with condition)?")
+@then("table (?P<table>.+) has (?P<row_num>\\d+) rows?(?P<has_where> with condition)?")
 def db_table_row_count(context, table, row_num, has_where):
     assert table_exists(context.db, table)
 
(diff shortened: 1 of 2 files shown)


lonvia:
Thanks for finding those, though the more pythonic way to fix this is to use regex strings (r""). I've done this now in 337505d.

PR #2050, opened 2023-08-23
Github action: Test with PostgreSQL 15 instead of 14

joto:


.github/workflows/ci.yml
```diff
@@ -204,7 +204,7 @@ jobs:
       CXX: clang++-10
       LUA_VERSION: 5.3
       LUAJIT_OPTION: OFF
-      POSTGRESQL_VERSION: 14
+      POSTGRESQL_VERSION: 15
```
pnorman, line 207: The text above says pg13

pnorman (changes requested):
The job names say either pg14, or, in one case, pg13. The names should be updated to match the versions used.

PR #2052, opened 2023-08-23
docs/readme: Add link to Github for docs and example

tordans:
I was looking for the docs source code. This makes it a bit easier to find.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -6,6 +6,6 @@ the [compiled version](osm2pgsql.1) and the files necessary to create it.
 Call `make man` to re-build the man page.
 
 The documentation that used to be in this directory has been moved to
-the [osm2pgsql manual](https://osm2pgsql.org/doc/manual.html) or to the
-[examples](https://osm2pgsql.org/examples) on the website.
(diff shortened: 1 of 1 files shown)


joto:
Every page on the website has a link to the source at the bottom. And every section of the man page has a link to its source (pen icon). So I don't see the need here. In fact I think we should remove those links altogether, they were only put in there because older users of osm2pgsql expected some docs to be in that directory. But they have been moved to the web site a long time ago.

tordans:
I tend to search for content differently and I don't see the issue with linking those two github repos more closely in this context. I don't think removing the readme altogether would make it better.

However, I found the footer links to Github after submitting this PR so I know how where to look now…

joto:
The problem is maintainability. The more places where we have docs and links, the more places we have to keep updated when something changes. It happens all the time that something like this gets left behind creating more confusion than it solves in the long run.

I totally makes sense to look in the `docs` directory and expect to find the docs there or at least some kind of link. So I understand the reason for your change. But I think the root cause behind the problem is that the `docs` directory has the wrong name. You should have never looked in there in the first place, because it is only marginally about docs, it just contains the man pages (which are, in a way, docs, okay, ...). So I propose to change the directory name to `man` and remove that README. See #2055.

PR #2061, opened 2023-09-03
Improve performance of some low level code

joto:
Code is only used in the generalization.

Faster code for hex decoder and bit-fiddling in tracer code.


pnorman:
Windows test failures seem related to the changes, and there's warnings too.

joto:
> Windows test failures seem related to the changes, and there's warnings too.

Yeah. I aciddentally pushed the wrong version of that code. Should work now.

PR #2067, opened 2023-09-04
Fail on 32bit architectures.

sebastic:
From https://github.com/openstreetmap/osm2pgsql/issues/2065#issuecomment-1705401937:
> Officially we are not supporting 32bit architectures and have for a while now. Its not that you can do much with a 32bit system with OSM data. So maybe it is time to take these architectures out?

Let's make this explicit then, and fail early at configure time on 32bit architectures.


sebastic:
Duplicate of #2066

PR #2084, opened 2023-09-23
Generalization: Switch from using CImg to OpenCV library.

joto:
Processing is an order of magnitude faster.


pnorman:
For reference, this was a 10x speed up in my limited tests.

PR #2099, opened 2023-10-28
Use tags = NULL in middle tables if object doesn't have any tags

joto:
This doesn't make much of a difference for the ways and rels table, but if we store all nodes in the database, it does make a huge difference, because most nodes don't have any tags. For a current planet, disk usage for the nodes table goes from 476 GB down to 409 GB saving 67 GB or nearly 15%.

Additionally it makes use of that table simpler. If you want to do any queries on tags, you need an index on the tags column on the nodes/ways/rels tables like this:

`CREATE INDEX ON planet_osm_ways USING gin (tags);`

But that is wasteful, because of the empty tags. We probably want to generate them as

`CREATE INDEX ON planet_osm_ways USING gin (tags) WHERE tags != '{}'::jsonb;`

But now all queries on those tables have to include that extra condition so that the query planner will use the index.

`SELECT * FROM planet_osm_ways WHERE tags ? 'highway' AND tags != '{}'::jsonb;`

If we use NULLs, the index can be created as:

`CREATE INDEX ON planet_osm_ways USING gin (tags) WHERE tags IS NOT NULL;`

And now the query becomes simpler, because the NOT NULL is automatically taken into account by the query planner:

`SELECT * FROM planet_osm_ways WHERE tags ? 'highway';`

Note that this is an incompatible change to the new format middle tables, but they are still marked as experimental, so we can do this.

This PR also contains a second commit for future proofing the members list of the rels middle table in case we want to do a similar change for that column in the future.

### DIFF
--- a/src/middle-pgsql.cpp
+++ b/src/middle-pgsql.cpp
@@ -323,6 +323,10 @@ template <typename T>
 void pgsql_parse_json_tags(char const *string, osmium::memory::Buffer *buffer,
                            T *obuilder)
 {
+    if (*string == '\0') { // NULL
+        return;
+    }
+
     auto const tags = nlohmann::json::parse(string);
     if (!tags.is_object()) {
         throw std::runtime_error{"Database format for tags invalid."};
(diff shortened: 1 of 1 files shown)


joto:
The failing test is not related to this PR. I re-run an older [run on  master](https://github.com/osm2pgsql-dev/osm2pgsql/actions/runs/6390514138/job/18157343396) and it now fails also. It looks like a bug in the fmt lib or std lib to me. And it only appears when running clang in C++20 mode. We run this test so that if and when we eventually switch to C++20 we can be sure it will work, but that is way off, so I think we can ignore this for now.

pnorman:
Note: The documentation needs updating to reflect this, as it currently states they are NOT NULL.

PR #2108, opened 2023-11-07
We only need some parts of the huge OpenCV library

joto:


joto:
Doesn't work, because CMake doesn't notice that opencv is installed.

PR #2123, opened 2024-01-26
Remove info that new middle db format is experimental from man page

joto:
This should have been removed before version 1.10.0.


pnorman (approved):
The wording feels awkward, but it's no worse than before and I can't think of anything better.

PR #2131, opened 2024-02-02
Disable synchronous commit on all connections

joto:
This simplifies the code, because we don't have to decide whether to set this or not on each connection. And it doesn't matter much when setting this for read-only connections which don't need this.

Setting synchronous_commit = off speeds up writing to the database, because commis are not delayed until the WAL is written to disk. It can not lead to corruption of the database.

https://www.postgresql.org/docs/current/runtime-config-wal.html#GUC-SYNCHRONOUS-COMMIT

### DIFF
--- a/src/db-copy.cpp
+++ b/src/db-copy.cpp
@@ -130,9 +130,6 @@ void db_copy_thread_t::thread_t::operator()()
     try {
         m_conn = std::make_unique<pg_conn_t>(m_connection_params, "copy");
 
-        // Let commits happen faster by delaying when they actually occur.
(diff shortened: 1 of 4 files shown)


pnorman:
I was thinking turning it off could result in data loss if the DB rolls back but the state files don't, but I guess with storing state in-db that's no longer a concern.

joto:
@pnorman Yes, I should have mentioned that. The osm2pgsql-replication script runs outside osm2pgsql and will store the state in the db after everything else is done. (And it uses sychronous commit.)

PR #2135, opened 2024-02-08
Switch middle database default format to "new"

joto:


src/command-line-parser.cpp
```diff
@@ -763,11 +759,22 @@ options_t parse_command_line(int argc, char *argv[])
...
+        }
 
-    if (!options.slim) {
-        options.middle_database_format = 0;
+        options.middle_database_format = middle_database_format;
+    } else if (options.slim) {
+        // For append mode we keep old default for backwards compatibility,
+        // for create mode the new format is the default.
+        // Will be overwritten by properties setting if it exists.
+        options.middle_database_format = options.append ? 1 : 2;
```
lonvia, line 774: Shouldn't the append case remain '0' for 'please determine from the database'?

PR #2157, opened 2024-04-02
Remove gazetteer output and --with-forward-dependencies option

joto:


pnorman:
Is this targeted at 2.0?

joto:
@pnorman Yes

PR #2169, opened 2024-04-09
Throw errors for some command line options values/combinations

joto:
Instead of just printing warnings.


lonvia:
Can we delay this specific change for a version or two? Nominatim currently still supports gazetteer output (with older versions of osm2pgsql) and flex and simply uses the same options for both trusting that the ones it doesn't need for flex are ignored.

joto:
Okay, closing here for now.

PR #2172, opened 2024-04-11
Remove support for legacy middle database format

joto:


src/osm2pgsql.cpp
```diff
@@ -231,17 +231,7 @@ static void check_db_format(properties_t const &properties, options_t *options)
...
 
-    if (format == -1) {
-        // db_format not set, so this is a legacy import
-        return;
-    }
-
-    if (format == 0) {
-        throw std::runtime_error{
-            "This database is not updatable (db_format=0)."};
-    }
```
lonvia, line 242: This part still looks useful?

joto: This case was done earlier (in the `check_updatable()` function). I found some more improvements/simplifications I could do to that checking code and added another commit with that.

PR #2181, opened 2024-05-09
Add tile-based generalizer calling SQL commands

joto:
It is named "tile-sql". Parameters are "name", "src_table", "dest_table", "zoom", and "sql", the command to be called. Use {ZOOM}, {X}, and {Y} in the command to set the zoom level and tile coordinates.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -269,6 +269,7 @@ else()
                     src/gen/gen-rivers.cpp
                     src/gen/gen-tile-builtup.cpp
                     src/gen/gen-tile-raster.cpp
+                    src/gen/gen-tile-sql.cpp
                     src/gen/gen-tile-vector.cpp
                     src/gen/gen-tile.cpp
                     src/gen/params.cpp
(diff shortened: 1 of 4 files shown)


pnorman:
What is this strategy used for that the others do not do? Is it possible to make `vector-union` a  special case of `tile-sql`?

joto:
> What is this strategy used for that the others do not do?

It allows all strategies that can be implemented in SQL. It can also be used as an experimenting ground for new strategies that get later implemented in C++ for performance.

> Is it possible to make vector-union a special case of tile-sql?

Yes, that should be possible. I am undeciced on whether we want to have specialized code for stuff like vector-union to make usage simpler than with a generic mechanism implemented in this commit. We'll see how it goes. The generalization stuff is still marked as experimental, so I want to allow more experimentation at the moment.

pnorman:
> I am undeciced on whether we want to have specialized code for stuff like vector-union to make usage simpler than with a generic mechanism implemented in this commit.

Could we have a separate mode for vector-union that internally calls tile-sql with the right setup? I'm also fine with removing vector-union since no should use that over raster-union.

PR #2186, opened 2024-05-20
Flush and sync COPYs after nodes, ways, and relations in flex output

joto:
Most output tables are only filled by either nodes or ways or relations. So it makes sense to close COPY operations after the last node/way/relation was found. Before this change active COPY operations could sometimes be open for a long time.


joto:
I pushed a new version of this commit, that still ends the COPY command, but doesn't wait for it to finish any more so as not to block the main thread.

PR #2188, opened 2024-05-26
Remove special case for old PostGIS versions when clustering

joto:
Ordering by geometry was done differently before PostGIS 2.4. This commit removed the old code and adds a check that PostGIS 3 is available.

For this specific case a check for PostGIS 2.4 or above would be sufficient, but PostGIS 3 is available for a long time now (it is in Debian Bullseye ("oldstable", released 2021) and Ubuntu 20.04).

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,8 +67,8 @@ with other versions of those libraries (set the `EXTERNAL_*libname*` option to
 * [protozero](https://github.com/mapbox/protozero) (>= 1.6.3)
 
 It also requires access to a database server running
-[PostgreSQL](https://www.postgresql.org/) 9.6+ and
-[PostGIS](https://www.postgis.net/) 2.2+.
+[PostgreSQL](https://www.postgresql.org/) (version 9.6+ works, 13+ strongly
+recommended) and [PostGIS](https://www.postgis.net/) (version 2.5+).
 
 Make sure you have installed the development packages for the libraries
 mentioned in the requirements section and a C++ compiler which supports C++17.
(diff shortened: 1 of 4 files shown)


src/osm2pgsql.cpp
```diff
@@ -81,6 +81,10 @@ static void check_db(options_t const &options)
 
     init_database_capabilities(db_connection);
 
+    if (get_postgis_version().major < 3) {
+        throw std::runtime_error{"Need at least PostGIS version 3"};
+    }
```
pnorman, line 86: We don't need to check for PostGIS >= 3 as the SQL statement we've got will work on older versions. The order the old versions produce is not as good as GeoHash but it still works.

pnorman (changes requested):
I'm okay with bumping our requirements to >= 3 but I don't think we should enforce that, since there's nothing stopping it from running on 2.4

lonvia:
The minimum version of PostGIS we test for in the CI is 2.5. So this should also be the minimum supported version. I'd keep the PR as is and leave in the test but test for >2.5 instead.

joto:
I have updated the commit to check for 2.5 instead of 3.

README.md
```diff
@@ -67,8 +67,8 @@ with other versions of those libraries (set the `EXTERNAL_*libname*` option to
 * [protozero](https://github.com/mapbox/protozero) (>= 1.6.3)
 
 It also requires access to a database server running
-[PostgreSQL](https://www.postgresql.org/) 9.6+ and
```
lonvia, line 70: This is, what we test, so I'd leave it in. May be do something like (9.6+ works, 13+ strongly recommended)?

src/osm2pgsql.cpp
lonvia: Shouldn't that be `pv.minor < 5`?

README.md
joto: fixed

PR #2200, opened 2024-06-15
Allow empty config file in flex output

joto:
No flex config file, or one that doesn't define any tables is usually not very useful. But it can be used in slim mode to have the middle tables only. Or it can be used to do some other processing or instead if the null output for testing.

We still output a warning, but no error any more if the style file is missing from the command line or if there are no tables defined.


pnorman (approved):

lonvia:
Empty tables are fine but I would prefer that you have to explicitly state `-S empty.lua` to achieve the goal. It's an easy mistake to forget the -S parameter and then searching for hours why the output tables do not appear.

joto:
Changed as requested.

PR #2203, opened 2024-06-16
Use anonymous namespaces instead of declaring static functions

joto:
This is recommended in C++ Core Guidelines SF.22
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-unnamed2

Sorry for the long PR. This basically sticks a `namespace { ... } // anonymous namespace` around all sections that had static functions and removes the `static` keyword. In some cases I moved some functions around to consolidate several anon namespace sections into one. This is not consistently done though, in some case I thought it to be better to leave the anon section in the context of the code around it.

### DIFF
--- a/src/flex-lua-table.cpp
+++ b/src/flex-lua-table.cpp
@@ -18,7 +18,9 @@
 
 #include <lua.hpp>
 
-static void check_tablespace(std::string const &tablespace)
+namespace {
+
+void check_tablespace(std::string const &tablespace)
 {
     if (!has_tablespace(tablespace)) {
         throw fmt_error(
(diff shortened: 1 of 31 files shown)


src/flex-lua-table.cpp
```diff
@@ -202,10 +203,12 @@ static std::size_t idx_from_userdata(lua_State *lua_state, int idx,
     return eo;
 }
 
-static void parse_and_set_expire_options(lua_State *lua_state,
-                                         flex_table_column_t *column,
-                                         std::size_t expire_outputs_size,
-                                         bool append_mode)
+} // anonymous namespace
```
lonvia, line 206: Too early? The functions below were all static.

joto: I have fixed this and rebased to current master.

PR #2221, opened 2024-08-12
Make --flat-nodes also work in non-slim mode

joto:
This is useful if memory is tight.


lonvia:
Can we add a test for that combination?

joto:
* Added test
* Changed behaviour so that flat node file is not deleted by default, only when --drop is used (same as in slim mode) 
* Updated man page
* Rebase to master

PR #2261, opened 2024-10-02
Properly set schema for osm2pgsql_find_changed_ways()

falko17:
A temporary function `osm2pgsql_find_changed_ways` may be created during the update process, which has been placed under the `public` schema, regardless of any chosen custom schema. This is a problem if one implements security based on schemas, e.g., by assigning a custom schema for osm2pgsql and restricting it from accessing `public` elements (as mentioned at the end of [section 3.3](https://osm2pgsql.org/doc/manual.html#security-considerations) of the manual).

This PR fixes that by using the configured schema. Additionally, tests are added in which a situation like the above is simulated (e.g., a user that can only access a custom schema and not `public`). Please tell me if there's anything I should improve/change before this can be merged.

### DIFF
--- a/tests/test-output-flex-update.cpp
+++ b/tests/test-output-flex-update.cpp
@@ -11,6 +11,7 @@
 
 #include "common-import.hpp"
 #include "common-options.hpp"
+#include "common-pg.hpp"
 
 namespace {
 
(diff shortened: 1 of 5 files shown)


tests/test-output-flex-update.cpp
```diff
@@ -38,8 +39,51 @@ struct options_slim_expire
...
+        return testing::opt_t()
+            .slim()
+            .flex(conf_file)
+            .schema("myschema")
+            .user("limited");
+    }
+};
+
+// Return a string with the schema name prepended to the table name.
+std::string with_schema(char const *name, options_t const *options)
```
joto, line 76: We are not changing the options, so we should probably use a `const &` instead of `const *` as is done elsewhere in the osm2pgsql code. This will also make calling the function simpler.

joto, line 76: Can we name the parameter `table_name`? So it is clear that that's the table parameter and the schema comes from the options.

joto (changes requested):
Thanks, the PR looks pretty good except  for the small changes I mention.

joto:
@falko17  Most CI runs failed with this:

```
Connecting to database failed (context=middle.main): connection to server at
  "localhost" (::1), port 5432 failed: FATAL:  password authentication failed
  for user "limited"
```

Please check that.

falko17:
Thank you for the quick review! I've addressed both comments now.

> Most CI runs failed [...]

On most systems, password authentication seems to be enforced for Postgres—this wasn't the case on my system, hence I didn't catch it. Should be fixed now.

joto:
Looks much better. The failing builds are probably transitory, we'll try to run them again.

Can you please fix this report from clang-tidy:
```
/home/runner/work/osm2pgsql/osm2pgsql/tests/test-output-flex-update.cpp:79:16: warning: avoid repeating the return type from the declaration; use a braced initializer list instead [modernize-return-braced-init-list]
   79 |         return std::string(name);
      |                ^
```
 i.e. use `return {name};`

Thanks!

falko17:
Yes, I've changed it to the braced initializer list now.

falko17:
@joto The PR should be fully fixed now, but two of the CI runs are still failing—given the errors shown in the logs, I don't think this is the fault of the changes. Can we try simply re-running those?

joto:
Sorry for the delay. It is always annoying when the CI breaks for unknown reasons, it really has nothing to do witth your code, but it means we have to be extra careful. Thanks for the nice PR with tests and everything!

PR #2329, opened 2025-05-13
Add optional --end-at-sequence parameter to limit updates to a specific sequence number

panglose:
## Summary

This pull request introduces an **optional `--end-at-sequence` parameter** to `osm2pgsql-replication update`, allowing users to **limit the replication process to a specific sequence number**.

If provided, the replication process stops **before applying** any sequence **strictly greater** than the specified value.

---

## Motivation

When managing large datasets or performing staged updates, it is sometimes necessary to:

- **Apply updates incrementally**, stopping at a known sequence checkpoint.
- **Avoid consuming all available diffs** in a single run.
- **Synchronize updates across multiple systems** or environments by controlling the applied range.

This feature makes such workflows **safer** and **more predictable**, without impacting the default behavior when the option is not provided.

---

## Example usage

```bash
osm2pgsql-replication update --database=mydb --end-at-sequence=12345
```

- Applies all diffs **up to and including 12345**.
- **Stops before applying 12346 or higher**.

If `--end-at-sequence` is not specified, the replication process behaves **exactly as before**.

---

## Implementation Details

- The option is parsed and passed as `args.end_at_sequence` (integer, optional).
- The update loop **checks `endseq` before applying diffs**.
- **No changes** to the replication state if the stop condition is met.
- Includes **informative log messages** when stopping due to this limit.

---

## Compatibility

- **Backward-compatible**: default behavior is unchanged if the option is not used.
- **No impact** on existing workflows.

---

## Additional Notes

- This feature was designed to support **controlled or audited replication workflows**.
- It may be useful in **data pipelines**, **batch updates**, or **multi-stage import processes**.

---

### DIFF
--- a/scripts/osm2pgsql-replication
+++ b/scripts/osm2pgsql-replication
@@ -551,6 +551,8 @@ def update(props, args):
     LOG.debug("Calling osm2pgsql with: %s", ' '.join(osm2pgsql))
 
     while seq < current.sequence:
+
+        # --------- new CODE --end-at-sequence
         LOG.debug("Importing from sequence %d", seq)
         if outfile.exists():
             outfile.unlink()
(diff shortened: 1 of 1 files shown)


panglose:
Hope it helps somebody else than me.

PR #2333, opened 2025-05-29
New feature: Check location of OSM objects against list of regions

joto:
This PR introduces a new feature: Locators. A locator is initialized with one or more regions, each region has a name and a polygon or bounding box. A geometry of an OSM object can then be checked against this region list to figure out in which region(s) it is located. This check is much faster than it would be to do this inside the database after import.

Locators can be used for all sorts of interesting features:
* Read larger OSM file but import only data inside some area.
* Annotate each OSM object with the country (or other region) it is in. This can then, for instance, be used to show special highway shields for each country.
* Use the information which region the data is in for further processing, for instance setting of default values for the speed limit or using special language transliterations rules based on country.

Locators are created in Lua with `define_locator()`. Bounding boxes can be added with `add_bbox()`. Polygons can be added from the database by calling `add_from_db()` and specifiying an SQL query which can return any number of rows each defining a region with the name and the (multi)polygon as columns.

A locator can then be queried using `all_intersecting()` returning a list of names of all regions that intersect the specified OSM object geometry. Or the `first_intersecting()` function can be used which only returns a single region for those cases where there can be no overlapping data or where the details of objects straddling region boundaries don't matter.

Several example config files are provided in the flex-config/locator directory.


src/locator.cpp
```diff
@@ -0,0 +1,128 @@
...
+    log_debug("Building index for locator '{}'", name());
+    std::vector<idx_value_t> m_data;
+    m_data.reserve(m_regions.size());
+
+    std::size_t n = 0;
+    for (auto const &region : m_regions) {
+        m_data.emplace_back(region.box(), n++);
+    }
+
+    m_rtree.insert(m_data.cbegin(), m_data.cend());
```
lonvia, line 85: I don't think m_rtree gets cleared ever. So this function may produce duplicates in the index if someone calls locator.add() some time during the import in one of the callbacks.

tests/bdd/flex/locator.feature
```diff
@@ -0,0 +1,233 @@
...
+            local regions = osm2pgsql.define_locator({})
+            print('NAME[' .. regions:name() .. ']')
+            """
+        When running osm2pgsql flex
+        Then the standard output contains
+            """
+            NAME[]
+            """
+
+    Scenario: Define a locator without name is okay
```
lonvia, line 48: Title not adapted.

tests/bdd/steps/steps_execute.py
```diff
@@ -128,8 +128,10 @@ def setup_lua_tagtransform(context):
 def setup_inline_lua_style(context):
     outfile = context.workdir / 'inline_style.lua'
     outfile.write_text(context.text)
-    context.osm2pgsql_params.extend(('-S', str(outfile)))
-
+    if '-S' in context.osm2pgsql_params:
+        context.osm2pgsql_params[context.osm2pgsql_params.index('-S') + 1] = str(outfile)
+    else:
+        context.osm2pgsql_params.extend(('-S', str(outfile)))
```
lonvia, line 134: Please 'fix' the function below (setup_style_file()) in the same way.

joto:
> Just curious: what makes this processing by osm2pgsql so much faster than doing this in the database after the import?

I think this is mostly because everything happens in memory. Once the Locator is set up, the checks run completely in memory. The OSM data is loaded anyway as part of the osm2pgsql processing, so it is available for the check. Doing this in the database means getting all the data from disk, doing all the checks and then doing lots of database updates involving more IO including all the clever things the database has to do to keep consistent and all that. There is just so much more work involved.

Memory usage is negligable.

PR #2336, opened 2025-05-31
Enum refactorings

joto:
See individual commits


lonvia:
I like going away from the awkward enum constructs. Should we go back to using upper case for constants with constexpr? It's not exactly C++ coding style but being able to make out constants seems useful.

PR #2365, opened 2025-07-29
Simplify and fix code getting relation members

joto:
This changes the way we are getting node members of relations. We now always get the locations of those nodes directly in the middle code instead of only getting the nodes themselves and adding the locations through an extra call from the output.

This also fixes a problem where node locations for untagged nodes were not retrieved from the RAM middle. The node_get() code does that correctly, while the old code didn't.


lonvia:
This will make the code quite a bit slower when "store nodes" is enabled together with flatnode stiore. Before the change it would just go to the flatnode store to load the locations, now it will send a separate DB query for every single node.

joto:
Closing in favour of #2367

PR #2372, opened 2025-08-04
Fix: Remove extra colon that messes up the test

joto:


lonvia:
Lo and behold! behave has released a new version after 7 years.

PR #2376, opened 2025-08-11
Process untagged ways/relations in stage 1b/c correctly

joto:
When we introduced the "untagged" callbacks in bc87ea2c8b4ea0bcac09975627e78765fda8927e, we did not call the process_untagged_way/relation callbacks for untagged ways/relations  in stage 1b and 1c, but called the normal process_way/relation callbacks. That's not correct. So this is changed in this commit.

For stage2 processing the answer which callback to call is a bit more complicated: On first glance we should also call the "untagged" function for untagged objects. But if we call the "untagged" function in that case, we would have to implement that function and it would be called for the millions of untagged objects we are not interested in. So it is arguably better to call the normal processing functions here. After all we explicitly requested the processing function to be called with the select_relation_members() call. So we are keeping this behaviour.

### DIFF
--- a/src/output-flex.cpp
+++ b/src/output-flex.cpp
@@ -835,7 +835,7 @@ void output_flex_t::get_mutex_and_call_lua_function(
 
 void output_flex_t::pending_way(osmid_t id)
 {
-    if (!m_process_way) {
+    if (!m_process_way && !m_process_untagged_way) {
         return;
     }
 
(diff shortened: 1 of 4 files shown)


src/output-flex.cpp
```diff
@@ -844,8 +844,13 @@ void output_flex_t::pending_way(osmid_t id)
...
 
     way_delete(id);
+    auto const &func = m_way_cache.get().tags().empty() ? m_process_untagged_way
+                                                        : m_process_way;
+    if (!func) {
+        return;
+    }
 
-    get_mutex_and_call_lua_function(m_process_way, m_way_cache.get());
+    get_mutex_and_call_lua_function(func, m_way_cache.get());
```
lonvia, line 853: Can we make the code simpler (also below):

```
    if (func) {
        get_mutex_and_call_lua_function(func, m_way_cache.get());
    }
```

lonvia (commented):
We really should have tests for that.

PR #2436, opened 2025-12-02
Add script for style testing

lonvia:
This adds the script osm2pgsql-test-style which allows blackbox testing of osm2pgsql lua styles. The script uses the same BDD test style as already introduced with osm2pgsql's integration tests. For details and documentation see: https://github.com/osm2pgsql-dev/osm2pgsql-website/pull/78

The PR also changes the internal tests to use the style tester. The semantics of the steps have slightly changed to make them more consistent and understandable. So there were some changes to the tests necessary, most notably:

* Failure/success of osm2pgsql execution is now tested in a separate step.
* Geometry comparisons no longer need the geometry to be wrapped in `ST_AsText()` but use the `!geo` matcher instead.
* The node grid now has it's origin in the lower left corner as one would ordinarily expect. That required turning the grid tables upside down for some tests that relied on nodes having certain coordinate values.

Note: the tester originally also changed the behaviour with respect to environment variables and ignored the ones set outside. However, it turns out that `pg_virtualenv` needs some `PG*` variables set to have osm2pgsql use the right cluster. And the Windows CI just refused to do such basic things like name resolution when some of its variables are unset. I've not further investigated but simply reinstated the previous behaviour of forwarding all environment variables.


scripts/osm2pgsql-test-style
```diff
@@ -0,0 +1,833 @@
...
+        data_stdin = None
+
+    proc = Popen(cmdline, cwd=str(context.workdir),
+                 stdin=PIPE, stdout=PIPE, stderr=PIPE)
+
+    outdata = proc.communicate(input=data_stdin)
+    context.osm2pgsql_cmdline = ' '.join(cmdline)
+    context.osm2pgsql_outdata = [d.decode('utf-8').replace('\\n', '\n') for d in outdata]
+    context.osm2pgsql_returncode = proc.returncode
+    context.osm2pgsql_params = {'-d': context.user_args.test_db}
```
joto: Why is this set here *after* building the command line in line 503 and 507?

joto (changes requested):

PR #2438, opened 2025-12-05
Add note on FreeBSD support to README.md

louwers:


README.md
```diff
@@ -29,6 +29,7 @@ to install and run osm2pgsql.
   the latest version via backports.
 * __macOS__: osm2pgsql is available through [Homebrew](https://brew.sh/)
 * __Windows__: use our [binary builds](https://osm2pgsql.org/download/windows/)
+* __FreeBSD__: osm2pgsql is available in the [FreeBSD ports tree](https://cgit.freebsd.org/ports/tree/converters/osm2pgsql)
```
pnorman: Is there a better page to drop users on? Even the [ports search page](https://ports.freebsd.org/cgi/ports.cgi?query=osm2pgsql) has a more user-friendly display because it contains information, rather than links to files with information

PR #2460, opened 2026-04-07
Introduce new "id cache" for node tables

joto:
It is sometimes useful to know whether a way contains nodes of a certain type, i.e. with certain tags. This commit adds a new functionality called "id caches" that allows to find out.

When defining a node table in Lua with `define_table` the `ids` section can now contain a field `cache = true`. If this is set the ids of all nodes written to that table are stored in memory. Later, when processing ways, that cache can be queried with `tablename:in_id_cache()`. Usually this would be used to query all the member nodes of a way: `...:in_id_cache(object.nodes)`. The `in_id_cache()` function returns the indexes into the `object.nodes` array where the id matches. From there it is possible to get the id of the matching node (with `object.nodes[idx]`) or the location of that node (with `object:as_point(idx)`. This information can be stored in a way table which will be correctly updated in append mode. (In append mode the in-memory id cache is populated with all the ids of the table for which the cache was defined.)

The whole thing only works for node tables and querying only works in the process_way() function. It can possibly be extended later to work in other contexts.

### DIFF
--- a/flex-config/turning-circles.lua
+++ b/flex-config/turning-circles.lua
@@ -1,0 +1,69 @@
+-- This config example file is released into the Public Domain.
+
+-- Create a table with turning circles that can be styled in sync with the
+-- highway they are on.
+
+local turning_circles = osm2pgsql.define_table({
+    name = 'turning_circles',
+    ids = { type = 'node', id_column = 'node_id', cache = true },
+    columns = {
+        { column = 'geom', type = 'point', not_null = true },
+    }
+})
+
+local highways = osm2pgsql.define_table({
+    name = 'highways',
+    ids = { type = 'way', id_column = 'way_id' },
+    columns = {
+        { column = 'htype', type = 'text', not_null = true },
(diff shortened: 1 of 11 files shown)


flex-config/turning-circles.lua
```diff
@@ -0,0 +1,69 @@
...
+})
+
+-- This table will contain entries for all node/way combinations where the way
+-- is tagged as "highway" and the node is tagged as "highway=turning_circle".
+-- The "htype" column contains the highway type, the "geom" the geometry of
+-- the node. This can be used, for instance, to draw the point in a style that
+-- fits with the style of the highway.
+--
+-- Note that you might have multiple entries for the same node in this table
+-- if it is in several ways. I that case you might have to decide at rendering
```
lonvia: I -> In

joto: Fixed and force pushed.

joto:
New version with somewhat cleaner code and changed various error checking, added some tests.

PR #2471, opened 2026-05-16
Add timestamp(tz) column types

joto:
With the special behaviour that a number is interpreted as seconds since the epoch which is the format used by object.timestamp. This avoids having to format the timestamp as string in the Lua script.


lonvia (commented):
The tests don't cover timestamp-as-string, don't they?

joto:
The tests in tests/bdd/flex/timestamp.feature take timestamps from tags and write them as string.

PR #2473, opened 2026-05-16
Allow detecting duplicate nodes in ways when creating geometry

joto:
When linestrings or polygons are created from ways, duplicate nodes are removed automatically. Generally this is what you want. But sometimes it makes sense to be able to detect this situation, for instance when creating some debug output.

The information is available on the C++ side anyway, this change exposes this information in the Lua code as a second return parameter of the as_linestring and as_polygon functions which can be ignored during normal use.

### DIFF
--- a/src/geom-from-osm.cpp
+++ b/src/geom-from-osm.cpp
@@ -85,13 +85,15 @@ void fill_polygon(polygon_t *polygon, osmium::Area const &area,
 
 } // anonymous namespace
 
-void create_linestring(geometry_t *geom, osmium::Way const &way)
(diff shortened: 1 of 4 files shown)


lonvia:
I'm curious what use-case you have in mind here. I could see that it would be useful to have the exact mapping node id -> linestring supporting point in case there is no 1:1 matching. Having just a boolean seems not a lot of information to go on.

joto:
This just allows detecting those cases in a cheap way, supplying the information we have to the user. To fix these cases you have to go into an editor anyway and look at the detailed situation. Sure it would be better to have the location of the duplicated node, but the low-level function doesn't provide it, so that's better than nothing.

PR #2474, opened 2026-05-17
Add spherical_length() function to geometries in Lua

joto:
For linestrings and multilinestrings this returns the same value as the PostGIS ST_Length() function on a geography. For other geometry types this always returns 0.0.

### DIFF
--- a/src/flex-lua-geom.cpp
+++ b/src/flex-lua-geom.cpp
@@ -100,6 +100,25 @@ int geom_length(lua_State *lua_state)
     return 1;
 }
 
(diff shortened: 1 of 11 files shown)


joto:
Looks like this is failing on PostgreSQL 11 due to results having less precision. On all newer PostgreSQL versions it is fine. And Pg 11 is not supported upstream any more, I am not sure why we are even testing against it.

lonvia:
Because PG11 is the lowest supported version, see README.md.

PR #2480, opened 2026-05-18
Set minimum supported PostgreSQL version to 12

joto:
There is a change somewhere either between pg 11 and 12 or with the associated PostGIS version used where float resuls have more precision in the new version which makes our tests fail on 11.

Version 11 is out of support upstream anyways.


lonvia:
CMakeLists.txt has also references to minimal Postgres.

joto:
> CMakeLists.txt has also references to minimal Postgres.

Fixed

PR #2484, opened 2026-06-02
Throw errors for some command line options values/combinations

joto:
Instead of just printing warnings.

Warnings are in since at least 2.0.0 released in September 2024.


lonvia:
Expect outcries.

PR #2488, opened 2026-06-05
Support type="double" for columns

amandasaurus:
osm2pgsql already supports `real`, which is [single precision floating point in PostgreSQL](https://www.postgresql.org/docs/current/datatype-numeric.html#:~:text=real). This patch adds `type = "double"` for PostgreSQL [`double precision` columns](https://www.postgresql.org/docs/current/datatype-numeric.html#:~:text=double%20precision).

Users were still able to add this type of column with `sql_type = "double precision"}` (e.g. [postpass lua](https://github.com/woodpeck/postpass-ops/blob/0d526cf414eb4e52efb0fbbca169409f70fe8ed0/postpass.lua#L74)). But this makes it easier.

Lua uses [double precision floating point types](https://www.lua.org/manual/5.5/manual.html#2.1:~:text=Standard%20Lua%20uses%2064%2Dbit%20integers%20and%20double%2Dprecision%20%2864%2Dbit%29%20floats%2C) by default.

This PR is in draft, because I was unable to run the tests. I have successfully tested this code locally. But I am not confident enough with how the interacts with lua. I would appreciate someone confirming it's OK.

Documentation for this is on my [`osm2pgsql-website` fork](https://github.com/osm2pgsql-dev/osm2pgsql-website/compare/master...amandasaurus:osm2pgsql-website:pg-num-types?expand=1)


src/flex-table-column.cpp
```diff
@@ -45,6 +45,7 @@ std::vector<column_type_lookup> const COLUMN_TYPES = {
      {"int8", table_column_type::int8},
      {"bigint", table_column_type::int8},
      {"real", table_column_type::real},
+     {"double", table_column_type::double_},
```
joto: To avoid the awkward C++ name `double_`, maybe name it `double_precision`? `double` is okay for the Lua side, I am only talking about the C++ side here.

src/flex-write.cpp
```diff
@@ -359,6 +359,16 @@ void flex_write_column(lua_State *lua_state, geometry_cache_t *geom_cache,
             throw fmt_error("Invalid type '{}' for real column.",
                             lua_typename(lua_state, ltype));
         }
+    } else if (column.type() == table_column_type::double_) {
```
joto: Because this is just a copy of the code for the `real` type (except the one place in the error message), I think we can do both in one if-condition with "type=real or type=double" and change the error message to "...for real/double column".

joto (changes requested):
We need tests for this, I suggest in `tests/test-output-flex-types.cpp` . If you can't run the tests locally, you'll see the results of the test run in your branch on Github.

joto:
Thank you!

PR #2491, opened 2026-06-28
Allow table inserts without context object

joto:
So far an `insert()` into a table was only allowed in the node, way and relation processing functions, i.e. those function with a "context object". That makes sense in so far as the id column(s) can only been filled magically if there is a context object. But osm2pgsql has allowed tables without id column(s) for a long time. These tables can now be filled from any processing function, for instance in the `after_*` callbacks.

It is still not possible to call `insert()` from the main Lua code outside any of the callbacks. That's because the database connections have not been set up yet.

### DIFF
--- a/src/output-flex.cpp
+++ b/src/output-flex.cpp
@@ -735,43 +735,38 @@ bool output_flex_t::relation_cache_t::add_members(middle_query_t const &middle)
     return true;
 }
 
-osmium::OSMObject const &
+osmium::OSMObject const *
 output_flex_t::check_and_get_context_object(flex_table_t const &table)
 {
     if (m_calling_context == calling_context::process_node) {
         if (!table.matches_type(osmium::item_type::node)) {
             throw fmt_error("Trying to add node to table '{}'.", table.name());
         }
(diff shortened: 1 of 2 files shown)


src/output-flex.cpp
```diff
@@ -806,8 +807,16 @@ int output_flex_t::table_insert()
...
             }
             if (column.type() == table_column_type::id_type) {
-                copy_mgr->add_column(type_to_char(object.type()));
+                if (objtype == osmium::item_type::undefined) {
+                    throw fmt_error("No context object for column '{}'",
+                                    column.name());
+                }
+                copy_mgr->add_column(type_to_char(objtype));
             } else if (column.type() == table_column_type::id_num) {
+                if (id == 0) {
```
lonvia: Isn't it safer to check for `objtype == osmium::item_type::undefined` here as well? Just in case somebody works with zero IDs? (Or, in fact, just check for `(!object)`.)

joto: Yes, its better to just check for `!object` consistently. I have changed that now.

PR #2498, opened 2026-07-29
Transform all estimated extents to EPSG:3857

Nakaner:
If the database table contains geometries in EPGS:4326, ST_EstimateExtent would return a wrong extent for tile-based processing.

My use case: I upgrade our OSM Carto fork for WMS at Geofabrik. I port our existing simplification/generalisation toolchain to osmpgsql-gen. All database tables (except those polygon layers which use the raster-union strategy) are in EPSG:4326.

When osm2pgsql-gen retrieves the extent from the database, it is projection-agnostic. But tiling isn't. Tiling happens in EPSG:3857.

I don't know if the patch to the raster extents works at all. Tests pass but they passed even with a syntax error in the SQL query for vector layer extents (i.e. coverage of this part of the code needs to be improved by someone at some point in time :smile:).


joto:
Haven't look at it in detail yet, but `ST_SRID(way)` is definitely wrong, because it hardcodes the geometry column. Also: What happens if table is empty?

Nakaner:
> Haven't look at it in detail yet, but ST_SRID(way) is definitely wrong, because it hardcodes the geometry column.

Oh, sorry, that `way` should not have made it into the pull request. I pushed a new commit.

> What happens if table is empty?

If the table is empty, `NULL` are returned by both the old and the new query. However, if geometries have no SRID set, the `ST_Transform` throws an error: "ST_Transform: Input geometry has unknown (0) SRID". I pushed a new commit fixing this.

Old implementation (for empty tables, it does not matter if the geometry column has an SRID set or not):

```sql
CREATE TABLE test (osm_id BIGINT NOT NULL, geom geometry(Point, 3857));
ANALYZE test;
SELECT ST_XMin(e), ST_YMin(e), ST_XMax(e), ST_YMax(e) FROM ST_EstimatedExtent('public', 'test', 'geom') AS e; -- NULLs returned
SELECT ST_XMin(e), ST_YMin(e), ST_XMax(e), ST_YMax(e)
  FROM ST_Transform(
    ST_SetSRID(
      ST_EstimatedExtent('public', 'test', 'geom'), 
      (SELECT ST_SRID(geom) FROM public.test LIMIT 1)
    ),
  3857) e; -- NULLs returned
```

If the SQL query returns `NULL`s, an invalid extent will be returned (and logged as "Source table empty, nothing to do.").

By looking on the code a second time, I found that an unset SRID will cause an error. I fixed this in a new commit.

src/gen/osm2pgsql-gen.cpp
```diff
@@ -114,13 +114,32 @@ tile_extent get_extent_from_db(pg_conn_t const &db_connection,
...
+            " FROM ("
+            "  SELECT ST_Transform(extent, 3857) AS extent"
+            "   FROM raster_columns WHERE r_table_schema='{}'"
+            "   AND r_table_name='{}' AND r_raster_column = '{}'"
+            " ) a",
             schema, table, column);
     } else {
         result = db_connection.exec(
+            "WITH srid AS ("
+            "  SELECT ST_SRID({}) AS srid FROM {} LIMIT 1"
```
joto, line 126: The column and table name should be quoted.

`ST_SRID("{}") AS srid FROM "{}" LIMIT 1`

joto (commented):
I have two problems with this:

a) Everything working with tiles in osm2pgsql assumes 3857 and we don't guarantee that it works with anything else. Presumably you have tested this in your case, but I think we still need to figure out what the guarantees of osm2pgsql are here and we should have tests to make sure we keep those guarantees.
b) It is not obvious that the implementation here is correct for all cases. We could have several geometries with different SRIDs for instance (although I have never seen that). Or no `extent` is set for the raster columns (which is currently the case in a database I have on hand here). It all hinges on obscure settings in the `raster_columns` and `geometry_columns` views and the behaviour of the `ST_EstimatedExtent()` function.

I am not saying this is wrong, I am just not sure we are understanding all the ramifications well enough. And maybe we can improve the implementation. We can run more than one query for instance, first getting the SRID then doing something else or so. That might make this more readable than putting everything into the one query. And it would give us the chance of generating better warning/error messages.

joto:
Oh, I should have said something else: The original code isn't all that great and robust either. So the questions I am raising are not necessarily all due to the new code but more of a "lets do this right while we are at it anyway" thing.

